PR #2, opened 2023-06-05
Zaza utils

esunar:
The zaza utilities copied from [zaza](https://github.com/openstack-charmers/zaza) and [zaza-openstack-tests](https://github.com/openstack-charmers/zaza-openstack-tests). Their corresponding unit tests are also copied and keep in a seperate module in case we need to update those.

Adding zaza/zaza-openstack-tests would have been an option. However they have tremendous dependencies.


pyproject.toml
```diff
@@ -70,7 +70,7 @@ exclude = [
 [tool.coverage.run]
 relative_files = true
 source = ["."]
-omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py"]
+omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py", "cou/zaza_utils/**"]
```
rgildein, line 73: Why do not use lib directory for zaza?

rgildein (approved):
I think that zaza could be in lib directory, but other than that LGTM. 
Maybe you could also provide commit of hash which was used to get these files.

gabrielcocenza (approved):
LGTM

pyproject.toml
rgildein: That's why I think we should attached hash of zaza version, so we can easily see which version was used. Using `lib` directory is just more intuitive for me instead of hiding it in source code, but yes these is not sub-module so no strong opinion.

PR #4, opened 2023-06-07
Add charm upgrade and channel refresh utilities

agileshaw:
This PR adds `charm_operation.py` which make use of zaza's upgrade_charm function to upgrade revisions and switch channels.

### DIFF
--- a/cou/steps/charm_operation.py
+++ b/cou/steps/charm_operation.py
@@ -15,24 +15,17 @@
 
 """Charm operation utilities."""
 import logging
-from cou.zaza_utils.model import upgrade_charm, block_until_all_units_idle
+
+from cou.zaza_utils.model import upgrade_charm
 
 
 def charm_upgrade(application_name: str) -> None:
     """Upgrade a charm to the latest revision in the current channel."""
-    logging.info(
-        "Upgrading %s to the latest revision in the current channel",
-        application_name
-    )
+    logging.info("Upgrading %s to the latest revision in the current channel", application_name)
     upgrade_charm(application_name)
-    block_until_all_units_idle()
 
 
 def charm_channel_refresh(application_name: str, target_channel: str) -> None:
     """Refresh a charm to track a target channel."""
-    logging.info(
-        "Refresh %s to the %s channel",
-        application_name, target_channel
-    )
-    upgrade_charm(application_name, target_channel)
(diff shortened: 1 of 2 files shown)


cou/steps/charm_operation.py
```diff
@@ -0,0 +1,32 @@
...
+"""Charm operation utilities."""
+import logging
+from cou.zaza_utils.model import block_until_all_units_idle, upgrade_charm
+
+
+def charm_upgrade(application_name: str) -> None:
+    """Upgrade a charm to the latest revision in the current channel."""
+    logging.info("Upgrading %s to the latest revision in the current channel", application_name)
+    upgrade_charm(application_name)
+    block_until_all_units_idle()
```
esunar, line 25: This will block all execution. I think we should remove this and call after all units are refreshed

cou/steps/charm_operation.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def charm_channel_refresh(application_name: str, target_channel: str) -> None:
+    """Refresh a charm to track a target channel."""
+    logging.info("Refresh %s to the %s channel", application_name, target_channel)
+    upgrade_charm(application_name, target_channel)
+    block_until_all_units_idle()
```
esunar, line 32: This will block all execution. I think we should remove this and call after all units are refreshed

esunar (changes requested):
Missing unit tests

agileshaw:
@esunar Thanks, I have addressed your comments

cou/steps/charm_operation.py
```diff
@@ -0,0 +1,31 @@
...
+from cou.zaza_utils.model import upgrade_charm
+
+
+def charm_upgrade(application_name: str) -> None:
+    """Upgrade a charm to the latest revision in the current channel."""
+    logging.info("Upgrading %s to the latest revision in the current channel", application_name)
+    upgrade_charm(application_name)
+
+
+def charm_channel_refresh(application_name: str, target_channel: str) -> None:
```
rgildein, line 28: not-blocking:
```suggestion
def charm_channel_refresh(application_name: str, channel: str) -> None:
```

rgildein (approved):
LGTM

esunar (approved):

PR #6, opened 2023-06-07
Usability improvements for plan.

esunar:
- Added colorful input for not getting lost in logs.
- Catched invalid input values and re-request.

### DIFF
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -25,6 +25,8 @@ from termcolor import colored
 from cou.steps import UpgradeStep
 from cou.steps.backup import backup
 
+AVAILABLE_OPTIONS = "cas"
+
 
 def generate_plan(args: Namespace) -> UpgradeStep:
     """Generate plan for upgrade."""
@@ -59,7 +61,7 @@ def prompt(parameter: str) -> str:
 def apply_plan(upgrade_plan: Any) -> None:
     """Apply the plan for upgrade."""
     result = "X"
-    while result.casefold() not in "cas".casefold():
+    while result.casefold() not in AVAILABLE_OPTIONS:
         result = input(prompt(upgrade_plan.description)).casefold()
         match result:
             case "c":
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -20,6 +20,8 @@ import sys
 from argparse import Namespace
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


agileshaw (approved):
LGTM

cou/steps/plan.py
```diff
@@ -34,18 +36,42 @@ def generate_plan(args: Namespace) -> UpgradeStep:
...
-        sys.exit(1)
-    else:
-        logging.info("Skipped")
-
-    for sub_step in upgrade_plan.sub_steps:
-        apply_plan(sub_step)
+    result = "X"
+    while result not in "CcAsSs":
+        result = input(prompt(upgrade_plan.description))
+        match result:
```
rgildein, line 64: We will not support Python 3.8 (focal)?

cou/steps/plan.py
```diff
@@ -34,18 +36,42 @@ def generate_plan(args: Namespace) -> UpgradeStep:
...
-        upgrade_plan.run()
-    elif result.casefold() == "a".casefold():
-        sys.exit(1)
-    else:
-        logging.info("Skipped")
-
-    for sub_step in upgrade_plan.sub_steps:
-        apply_plan(sub_step)
+    result = "X"
+    while result not in "CcAsSs":
```
rgildein, line 62: I suggest to make this as constant + you made mistake it should be `"CcAaSs"`.
However we can do better with:
```python
while result := input(prompt(upgrade_plan.description)).lower() not in "cas":
    match result:
        case "c":
            upgrade_plan.run()
                for sub_step in upgrade_plan.sub_steps:
                    apply_plan(sub_step)
        case "a":
            logging.info("Aborning plan")
            sys.exit(1)
        case "s":
            logging.info("Skipped")
        case _:
            logging.info("No valid input provided!")
```

rgildein (changes requested):

cou/steps/plan.py
esunar: baseline will be 3.10

cou/steps/plan.py
esunar: If we enter c,a or s it never runs

esunar: fixed `CcAaSs` issue

esunar: done

PR #7, opened 2023-06-07
analyze packages and openstack-release to determine openstack version

gabrielcocenza:
- check if openstack version is the same across different application units.

- check if openstack version is the same across applications.

- check if "openstack-origin" is set right.

- check if charm is configured at the right channel.

- check if it's necessary to migrate from cs to ch.


esunar (changes requested):
We should not touch the zaza_utils package if it is absolutely necessary. Because we copied them from zaza and zaza-openstack-test repos. If there is any update it would be very hard to update those files. Please keep your code separate from those files.

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+        for os_release in os_sequence[:-1]:
+            next_release = determine_next_openstack_release(os_release)[1]
+            upgrade_charms[next_release].update(versions[os_release])
+            logging.warning(
+                "upgrade charm: %s from: %s to %s", versions[os_release], os_release, next_release
+            )
+
+    else:
+        actual_release = list(versions)[0]
+        next_release = determine_next_openstack_release(actual_release)[1]
```
rgildein, line 132: Can we make function to return actual and next release? This function is too long and I do not think that using `[0]` and `[1]` like this is easy to understand.

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+    for app in os_versions.keys():
+        app_channel, charm_origin = extract_app_channel_and_origin(app)
+        actual_release = list(os_versions[app].keys())
+        if len(actual_release) > 1:
+            logging.warning(
+                "Skip check of channels. App %s has units with different openstack version",
+                app,
+            )
+        else:
+            actual_release = actual_release[0]
```
rgildein, line 157:
```suggestion
            actual_release = actual_releases[0]
```

rgildein, line 157: To avoid nesting too deeply.
```suggestion
        if len(actual_release) > 1:
            logging.warning(
                "Skip check of channels. App %s has units with different openstack version",
                app,
            )
            continue

        actual_release = actual_release[0]
```

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+                    )
+                    change_openstack_release["distro"].add(app)
+            else:
+                if expected_os_origin not in os_charm_config:
+                    change_openstack_release[expected_os_origin].add(app)
+                    logging.warning(
+                        "App: %s need to set openstack-origin or source to %s",
+                        app,
+                        expected_os_origin,
+                    )
```
rgildein, line 199: What is purpose to have one if condition in if condition?
```suggestion
            if actual_release == "ussuri" and os_charm_config != "distro":
                logging.warning("App: %s need to set openstack-origin or source to 'distro'", app)
                change_openstack_release["distro"].add(app)
            elif expected_os_origin not in os_charm_config:
                change_openstack_release[expected_os_origin].add(app)
                 logging.warning(
                    "App: %s need to set openstack-origin or source to %s",
                    app,
                    expected_os_origin,
                )
```

cou/zaza_utils/generic.py
```diff
@@ -66,25 +66,26 @@ def get_network_config(net_topology, ignore_env_vars=False, net_topology_file="n
     return net_info
 
 
-def get_pkg_version(application, pkg):
+def get_pkg_version(application_charm, pkg, model_name=None):
```
rgildein, line 69: As Erhan mention we should not touch these function, since they are copied from zaza.

cou/zaza_utils/openstack.py
```diff
@@ -0,0 +1,155 @@
+import logging
```
rgildein, line 1: If we are coping something from zaza, I think that best practice would be do in in separate commit and use commit hash from which it was copied.

rgildein (changes requested):
I understand that these code need to use lot of nesting, but we should avoid using too many for, if, etc in one tree.

cou/zaza_utils/generic.py
aieri: why are we copying files from zaza? can the module not be imported and used?

cou/zaza_utils/openstack.py
aieri: or use a submodule if you can't simply import the package

aieri (changes requested):
I have not completed reviewing the whole PR, but I think I can start sending a few requests for changes

cou/steps/analyze.py
```diff
@@ -0,0 +1,256 @@
...
+                    change_openstack_release[expected_os_origin].add(self.name)
+                    logging.warning(
+                        "App: %s need to set openstack-origin or source to %s",
+                        self.name,
+                        expected_os_origin,
+                    )
+        return change_openstack_release
+
+
+def get_openstack_release(unit, model_name):
```
aieri, line 176: the more I think of this, the more centralizing version checks seems wrong to me: charms are inherently decentralized, why are we ssh-ing into the units? determining the version of an app should be determined by the app itself; the upgrader should simply ask the charm

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def extract_app_channel_and_origin(app: str) -> Tuple:
+    """Extract application channel and origin (cs or ch) by the juju status."""
+    app_status = get_application_status(app)
+    return app_status.get("charm-channel"), app_status.get("charm").split(":")[0]
+
+
+def extract_os_charm_config(app: str) -> str:
+    """Extract application configuration for openstack-origin."""
```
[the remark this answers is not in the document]
gabrielcocenza: I've changed the name of the function to `get_os_origin`

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+    app_config = model.get_application_config(app)
+    for origin in ("openstack-origin", "source"):
+        if app_config.get(origin):
+            return app_config.get(origin).get("value")
+
+    logging.warning("Failed to get origin for %s, no origin config found", app)
+    return ""
```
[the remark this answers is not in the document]
gabrielcocenza: I tried to apply your suggestion, but then started several issues with pylint. I think it's ok return an empty string.

cou/steps/analyze.py
```diff
@@ -0,0 +1,200 @@
...
+    upgrade_charms = defaultdict(set)
+
+    for app, os_release_units in os_versions.items():
+        os_version_units = set(os_release_units.keys())
+        for os_version_unit in os_version_units:
+            versions[os_version_unit].add(app)
+        if len(os_version_units) > 1:
+            logging.warning("Units from %s are not in the same openstack version", app)
+            os_sequence = sorted(
+                list(os_version_units),
```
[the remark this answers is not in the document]
gabrielcocenza: It's using set to have unique openstack releases and then  it's transforming to a list to have it sorted.

cou/steps/analyze.py
gabrielcocenza: `determine_next_openstack_release` comes from zaza, so I think it's better not change it.

cou/steps/analyze.py
```diff
@@ -0,0 +1,452 @@
...
+        )
+        for app, app_status in juju_status.applications.items()
+    ]
+    # NOTE(gabrielcocenza) Not all openstack-charms are mapped in the zaza lookup.
+    supported_os_charms = CHARM_TYPES.keys()
+    openstack_apps = [app for app in apps if app.charm in supported_os_charms]
+    return openstack_apps
+
+
+def analyze() -> None:
```
esunar, line 282: This should return Analyze as a result

cou/steps/analyze.py
gabrielcocenza: I left a comment in the code saying that Ideally, the application should provide the openstack version and packages versions by a charm action and this might be possible with Sumbeam.

esunar (approved):
LGTM for this iteration

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+    not_supported_apps_names = sorted([app.name for app in not_supported_apps])
+    logging.warning(
+        "App(s): %s are not supported in the analyze process", ", ".join(not_supported_apps_names)
+    )
+    return openstack_apps
+
+
+def analyze() -> Analyze:
+    """Analyze the deployment before planning."""
+    logging.info("Analyzing the openstack release in the deployment...")
```
aieri, line 315:
```suggestion
    logging.info("Analyzing the OpenStack deployment...")
```

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+def analyze() -> Analyze:
+    """Analyze the deployment before planning."""
+    logging.info("Analyzing the OpenStack deployment...")
+    apps = generate_model()
+    # E.g: {"ussuri": {"keystone"}, "victoria": {"cinder"}}
+    os_versions: defaultdict[str, set] = defaultdict(set)
+
+    # E.g {"victoria": {"keystone/0", "keystone/1"}} this means that units
+    # keystone/0 and keystone/1 should be upgraded to victoria
+    upgrade_units: defaultdict[str, set] = defaultdict(set)
```
aieri, line 322: this set of dictionaries feels misplaced to me. If this is the analyze step, we should not be making decisions about what should happen; that should rather be in the planning phase. I think a better logical separation would be for the analyze step to only assess the status of the cloud, and for the plan step to make decisions around what should change in order to reach the goal.

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+                    }
+                    for unit, details in self.units.items()
+                },
+            }
+        }
+
+    def extract_charm_name(self) -> str:
+        """Extract the charm name using regex."""
+        match = re.match(
+            r"^(?:\w+:)?(?:~[\w\.-]+/)?(?:\w+/)*([a-zA-Z0-9-]+?)(?:-\d+)?$", self.status.charm
```
aieri, line 139: has this been reverse engineered? do we have a formal and trusted specification of how a charm name is composed?

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+                    unit: {
+                        "pkg_version": details.get("pkg_version", ""),
+                        "os_version": details.get("os_version", ""),
+                    }
+                    for unit, details in self.units.items()
+                },
+            }
+        }
+
+    def extract_charm_name(self) -> str:
```
aieri, line 136: in my mind this would be clearer as a static method

aieri (changes requested):

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+    """Analyze result."""
+
+    upgrade_units: defaultdict[str, set]
+    upgrade_charms: defaultdict[str, set]
+    change_channel: defaultdict[str, set]
+    charmhub_migration: defaultdict[str, set]
+    change_openstack_release: defaultdict[str, set]
+
+
+@dataclass
```
rgildein, line 70: This class will be not be changed, if I'm correct, so I'll suggest to use `frozen=True` to prevent any change. (you will need to move __post__init__ to different function, which will be served as constructor)

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+        """Check openstack versions in an application."""
+        if len(self.os_release_units.keys()) <= 1:
+            return upgrade_units
+
+        logging.warning("Units from %s are not in the same openstack version", self.name)
+        os_sequence = sorted(
+            list(self.os_release_units.keys()),
+            key=CompareOpenStack,
+        )
+        for os_release in os_sequence[:-1]:
```
rgildein, line 193: Why we are skipping lest item?

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+    ) -> Tuple:
+        """Check openstack channel and if it's necessary a charmhub migration."""
+        if len(self.os_release_units.keys()) > 1:
+            logging.warning(
+                "Skip check of channels. App %s has units with different openstack version",
+                self.name,
+            )
+            return change_channel, charmhub_migration
+
+        actual_release = list(self.os_release_units.keys())[0]
```
rgildein, line 215: What if there is no units?

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+                self.name,
+                self.channel,
+                expected_channel,
+            )
+        if self.charm_origin == "cs":
+            charmhub_migration[expected_channel].add(self.name)
+            logging.warning(
+                "App: %s is from charmstore. It's expected to be from charmhub", self.name
+            )
+        return change_channel, charmhub_migration
```
rgildein, line 230: again we are changing input arguments directly

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+            name=app,
+            status=app_status,
+            config=model.get_application_config(app),
+            model_name=model_name,
+        )
+        for app, app_status in juju_status.applications.items()
+    }
+    # NOTE(gabrielcocenza) Not all openstack-charms are mapped in the zaza lookup.
+    supported_os_charms = CHARM_TYPES.keys()
+    openstack_apps = {app for app in apps if app.charm in supported_os_charms}
```
rgildein, line 304:
```suggestion
    openstack_apps = {app for app in apps if app.charm in CHARM_TYPES}
```

cou/steps/plan.py
```diff
@@ -17,18 +17,17 @@
...
+from colorama import Fore, Style
 
 from cou.steps import UpgradeStep
 from cou.steps.backup import backup
 
 AVAILABLE_OPTIONS = "cas"
 
 
-def generate_plan(args: Namespace) -> UpgradeStep:
+def generate_plan(args: Any) -> UpgradeStep:
```
rgildein, line 30: Why this is any and not Analyze?

rgildein (changes requested):

cou/steps/analyze.py
gabrielcocenza: If we have an application without units something is wrong with the model, right? I think this is a very rare scenario. Do you think it worth adding a try except IndexError?

cou/steps/analyze.py
gabrielcocenza: It seems that comes from [here](https://github.com/juju/charm/blob/master/url.go#L51). What python libjuju shows as the field `charm` seems to be this url.

cou/steps/analyze.py
gabrielcocenza: I found that we already have a function doing this extraction on [utils](https://github.com/canonical/charmed-openstack-upgrader/pull/10/files#diff-555158e0b6270c493abf1b1f89d7577a60a1cab02a61afd96137a6e7557bc9e4L25) so it really makes sense your suggestion.

Thanks

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+) -> Dict[str, Dict]:
+    """Prepare the log for change openstack release."""
+    for next_os_origin, app_names in change_openstack_release.items():
+        for app_name in app_names:
+            current_os_origin = outputs[app_name]["os_origin"]
+            outputs[app_name]["os_origin"] = add_color_fields(current_os_origin, next_os_origin)
+    return outputs
+
+
+def add_color_fields(old: str, new: str) -> str:
```
jneo8, line 483: **question:** 

What if we want to disable the color output in the future?

cou/zaza_utils/generic.py
```diff
@@ -66,21 +66,23 @@ def get_network_config(net_topology, ignore_env_vars=False, net_topology_file="n
     return net_info
 
 
-def get_pkg_version(application, pkg):
+def get_pkg_version(application, pkg, model_name=None):
```
jneo8, line 69: **nitpick:** 

I would like to know the type of the model_name from type annotation.

cou/zaza_utils/openstack.py
```diff
@@ -0,0 +1,91 @@
+import re
```
jneo8, line 1: **chore (blocking):** 

I would like to provide unit tests for the functions in this module.
It's will be nice if those string handling functions has tests.

jneo8 (changes requested):
Basic unit tests required.

Also because this PR to so huge.
Is that possible that we separate it to different small PRs so it will be more friendly for reviewers?

cou/steps/analyze.py
```diff
@@ -0,0 +1,488 @@
...
+            self.units[unit]["os_version"] = os_version
+            self.os_release_units[os_version].add(unit)
+
+    def __hash__(self) -> int:
+        """Hash magic method for Application."""
+        return hash(self.name)
+
+    def __eq__(self, other: Any) -> Any:
+        """Equal magic method for Application."""
+        return other.name == self.name
```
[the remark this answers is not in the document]
gabrielcocenza: I've changed to use charm and name. I think this will be enough.

gabrielcocenza:
These changes are now included at #12 with the removal of zaza_utils and turn the interaction with juju async.

PR #8, opened 2023-06-09
Added functional tests

esunar:


cou/steps/backup.py
```diff
@@ -25,7 +25,7 @@
 )
 
 
-def backup() -> None:
+def backup() -> AnyStr:
```
rgildein, line 28: `os.path.abspath` return string, no bytes so `str` is enough.

tests/functional/tests/backup.py
```diff
@@ -0,0 +1,35 @@
...
+        """Run class setup for noop tests."""
+        super(BackupTest, cls).setUpClass()
+
+    def tearDown(self):
+        """Close thread."""
+        logger.info("Closing libjuju thread.")
+        clean_up_libjuju_thread()
+        logger.info("Removing test file")
+        try:
+            os.remove(self.backup_file)
```
rgildein, line 26: Why we need to use exception here? I think removing it if exist would be more than enough.

tests/functional/tests/base.py
```diff
@@ -0,0 +1,18 @@
+"""Generic setup for functional tests."""
+import logging
+import unittest
+
+logger = logging.getLogger(__name__)
+
+
+class NoopTest(unittest.TestCase):
```
rgildein, line 8: Can you explain why this is needed?

tests/functional/tests/tests.yaml
```diff
@@ -0,0 +1,14 @@
+tests:
+  - openstack:
+    - tests.functional.tests.base.NoopTest
+    - tests.functional.tests.backup.BackupTest
+gate_bundles:
+  - openstack: openstack
+
+configure:
+  - openstack:
+#    - zaza.openstack.charm_tests.glance.setup.add_lts_image
```
rgildein, line 10: If any configuration is not needed, I would prefer to remove them instead of commenting them.

tox.ini
```diff
@@ -10,6 +10,7 @@ skip_missing_interpreters = True
 [testenv]
 basepython = python3
 setenv = PYTHONPATH={toxinidir}
+allowlist_externals = pytest
```
rgildein, line 13: Why this is needed? We are adding functional tests, which are using zaza and not pytest.

rgildein (changes requested):

tests/functional/tests/base.py
esunar: removed

tests/functional/tests/backup.py
esunar: easier

cou/steps/backup.py
esunar: AnyStr in posixpath.pyi

tests/functional/tests/tests.yaml
esunar: removed

tox.ini
esunar: Because zaza uses pytest and if we do not define in tox.ini gives error

cou/steps/backup.py
rgildein: I do not understand? I have no idea what is `posixpath.pyi`

tests/functional/tests/backup.py
rgildein: Yes, but like this you are ignoring other error exception also, you should use `FileNotFoundError`.

tox.ini
rgildein: Where, we never used this with any func tests and we used pytest a lot a never define it in `allowlist_externals`. So I'm confused here.

esunar: sorry it is a tox4 issue. Since we are using tox3 I removed it

.github/workflows/pr.yaml
```diff
@@ -13,12 +13,31 @@ on:
...
-      snapcraft: true
-      commands: "['make functional']"
+      python-version: "['3.10']"
+      tox-version: ""
+      working-directory: "."
+
+  #functional tests
+  func:
+    name: Functional tests
+    runs-on: self-hosted
```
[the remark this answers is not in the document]
esunar: Yes we need to add it.

Makefile
```diff
@@ -53,7 +53,7 @@ dev-environment:
 	@echo "Creating virtualenv and installing pre-commit"
 	@tox -r -e dev-environment
 
-functional: build
+functional:
```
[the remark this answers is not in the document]
esunar: Good catch, testing code :blush:

tox.ini
```diff
@@ -51,8 +51,10 @@ commands = pytest {toxinidir}/tests/unit \
...
 
 [testenv:func]
+changedir = {toxinidir}/tests/functional
 deps = .[functests]
 passenv =
     TEST_*
+    OS_*
 commands =
-    pytest {toxinidir}/tests/functional {posargs:-v}
+    functest-run-suite {posargs}
```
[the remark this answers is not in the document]
esunar: done

tests/functional/tests/backup.py
```diff
@@ -0,0 +1,32 @@
...
+        clean_up_libjuju_thread()
+        logger.info("Removing test file")
+        os.remove(self.backup_file)
+
+    def test_backup(self):
+        """Backup Test."""
+        logger.info("Running backup test....")
+        self.backup_file = backup()
+        logger.info("Backup file: %s", self.backup_file)
+        assert os.path.getsize(self.backup_file) > 0
```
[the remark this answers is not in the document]
esunar: done

rgildein (approved):

tests/functional/tests/bundles/openstack.yaml
gabrielcocenza, line 1: non-blocker: change the name of the bundle to `focal-ussuri.yaml`

gabrielcocenza (approved):
LGTM.

Left one comment about changing the bundle name

tests/functional/tests/bundles/openstack.yaml
esunar: leaving it as openstack.yaml until we add new one

PR #9, opened 2023-06-20
added log file with debug level

gabrielcocenza:
- stream log gets level from the cli input.
- new log file with DEBUG level is created every execution at $HOME/.local/share/cou/log.
- plugin to have write access at $HOME/.local/share/cou/.

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -20,12 +20,12 @@ import logging.handlers
 import os
 import pathlib
 import sys
+from datetime import datetime
 from typing import Any
 
 from cou.steps.plan import apply_plan, dump_plan, generate_plan
 
-HOME = os.getenv("HOME")
-COU_DIR_LOG = os.getenv("COU_DIR_LOG", f"{HOME}/.local/share/cou/log")
+COU_DIR_LOG = pathlib.Path(os.getenv("HOME", ""), ".local/share/cou/log")
 
 
 def parse_args(args: Any) -> argparse.Namespace:
@@ -68,10 +68,10 @@ def setup_logging(log_level: str = "INFO") -> None:
     root_logger.setLevel("DEBUG")
 
     # handler for the log file. Log level is DEBUG
+    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
+    file_name = f"{COU_DIR_LOG}/cou-{time_stamp}.log"
     pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
-    log_file_handler = logging.handlers.TimedRotatingFileHandler(
-        f"{COU_DIR_LOG}/cou.log", when="D", interval=1
-    )
+    log_file_handler = logging.FileHandler(file_name)
     log_file_handler.setFormatter(log_formatter)
 
     # handler for the console. Log level comes from the CLI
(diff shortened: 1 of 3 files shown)


agileshaw (approved):
LGTM

snap/snapcraft.yaml
```diff
@@ -41,3 +44,7 @@ plugs:
     interface: personal-files
     write:
       - $HOME/.local/share/juju
+  logs:
+    interface: personal-files
+    write:
+      - $HOME/.local/share/cou/log
```
esunar, line 50: I think we should have write permission to `$HOME/.local/share/cou` not its subfolder

cou/cli.py
```diff
@@ -16,11 +16,17 @@
...
+import logging.handlers
+import os
+import pathlib
 import sys
 from typing import Any
 
 from cou.steps.plan import apply_plan, dump_plan, generate_plan
 
+HOME = os.getenv("HOME")
+COU_DIR_LOG = os.getenv("COU_DIR_LOG", f"{HOME}/.local/share/cou/log")
```
rgildein, line 28: not-blocking: maybe we can make Path directly here, e.g. `pathlib.Path(os.getenv(...))`

cou/cli.py
```diff
@@ -59,11 +65,22 @@ def setup_logging(log_level: str = "INFO") -> None:
...
-    root_logger.setLevel(log_level)
-    if not root_logger.hasHandlers():
-        console_handler = logging.StreamHandler()
-        console_handler.setFormatter(log_formatter)
-        root_logger.addHandler(console_handler)
+    root_logger.setLevel("DEBUG")
+
+    # handler for the log file. Log level is DEBUG
+    pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
+    log_file_handler = logging.handlers.TimedRotatingFileHandler(
```
esunar, line 72: We can create separate file for for each run like in snapcraft with cou-yyyyMMddhhmmss-somerandomsuffix.log and show this while starting the application.

cou/cli.py
esunar, line 28: Since snap is confined, I think we can drop dir env support.

esunar (changes requested):

cou/cli.py
```diff
@@ -59,11 +65,22 @@ def setup_logging(log_level: str = "INFO") -> None:
...
+        f"{COU_DIR_LOG}/cou.log", when="D", interval=1
+    )
+    log_file_handler.setFormatter(log_formatter)
```
rgildein, line 75: not-blocking: Would it make any difference to use `log_file_handler.setLevel("DEBUG")` instead of `root_logger.setLevel("DEBUG")`?

rgildein (approved):
LGTM

cou/cli.py
gabrielcocenza: The [logger](https://docs.python.org/3/howto/logging.html#logging-flow) acts as first filter before passing the event to the handlers. So if the logger level is at higher severity (like INFO) it will filter the event and the handler won't receive it, even if the event was matching the handler's level (DEBUG).

esunar (approved):

PR #11, opened 2023-06-22
added traceback for better logs on executions with errors.

gabrielcocenza:
currently, just logging the exception error doesn't give enough context.

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -98,5 +98,5 @@ def entrypoint() -> int:
 
         return 0
     except Exception as exc:  # pylint: disable=broad-exception-caught
-        logging.error(exc)
+        logging.exception(exc)
         return 1


cou/cli.py
```diff
@@ -99,4 +100,5 @@ def entrypoint() -> int:
         return 0
     except Exception as exc:  # pylint: disable=broad-exception-caught
         logging.error(exc)
+        logging.error(traceback.format_exc())
```
rgildein, line 103:
```suggestion
        logging.exception(error)
```

rgildein (commented):
I think that you can use `logging.exception` to get same result.

cou/cli.py
gabrielcocenza: I didn't know about `logging.exception`. Thanks it's very helpful and cleaner.

rgildein (approved):

PR #19, opened 2023-07-21
Fix logging to not use root logger

### LINKED ISSUE #18: Logs are showing debug logs of python libjuju
running the `cou` started showing a lot of unsuspected logs as it's possible to see at this [pastebin](https://pastebin.canonical.com/p/V4B6jHw65d/).
This is happening because the project is using the root logger and as the [documentation](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) says:
> It is strongly advised that you do not log to the root logger in your library. Instead, use a logger with a unique and easily identifiable name, such as the __name__ for your library’s top-level package or module. Logging to the root logger will make it difficult or impossible for the application developer to configure the logging verbosity or handlers of your library as they wish.

gabrielcocenza:
- configure the top-level module for log
- every module references to the same logger object
- avoid duplicated logs by getting the logger on every sub module and not log via root logger
- added %(name)s to identify where the logs comes from
- filter console logs to have just logs from cou
- file logs have all logs on debug level

[0] https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
[1] https://docs.python.org/3/howto/logging-cookbook.html#using-logging-in-multiple-modules

Closes #18

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -80,8 +80,11 @@ def setup_logging(log_level: str = "INFO") -> None:
     :returns: Nothing: This function is executed for its side effect
     :rtype: None
     """
-    log_formatter = logging.Formatter(
+    log_formatter_file = logging.Formatter(
         fmt="%(asctime)s [%(name)s] [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
+    )
+    log_formatter_console = logging.Formatter(
+        fmt="%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
     )
     root_logger = logging.getLogger()
     root_logger.setLevel("DEBUG")
@@ -91,12 +94,12 @@ def setup_logging(log_level: str = "INFO") -> None:
     file_name = f"{COU_DIR_LOG}/cou-{time_stamp}.log"
     pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
     log_file_handler = logging.FileHandler(file_name)
-    log_file_handler.setFormatter(log_formatter)
+    log_file_handler.setFormatter(log_formatter_file)
 
     # handler for the console. Log level comes from the CLI
     console_handler = logging.StreamHandler()
     console_handler.setLevel(log_level)
-    console_handler.setFormatter(log_formatter)
+    console_handler.setFormatter(log_formatter_console)
     # just cou logs on console
     console_handler.addFilter(logging.Filter(__package__))
 
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -33,8 +33,7 @@ from cou.utils import juju_utils as utils
 COU_DIR_LOG = pathlib.Path(os.getenv("HOME", ""), ".local/share/cou/log")
 AVAILABLE_OPTIONS = "cas"
 
-# NOTE(gabrielcocenza) top-level module to reference the same logger object across modules.
-logger = logging.getLogger(__package__)
+logger = logging.getLogger(__name__)
 
 
 def parse_args(args: Any) -> argparse.Namespace:
@@ -98,6 +97,7 @@ def setup_logging(log_level: str = "INFO") -> None:
     console_handler = logging.StreamHandler()
     console_handler.setLevel(log_level)
     console_handler.setFormatter(log_formatter)
+    # just cou logs on console
     console_handler.addFilter(logging.Filter(__package__))
 
     root_logger.addHandler(log_file_handler)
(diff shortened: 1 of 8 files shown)


esunar (approved):
The code is totally ok. I've only one question, should we externalize the logging config and instead of giving log level should we give some logging profiles. Such as cou-only, everything etc.

cou/cli.py
```diff
@@ -33,6 +33,9 @@
 COU_DIR_LOG = pathlib.Path(os.getenv("HOME", ""), ".local/share/cou/log")
 AVAILABLE_OPTIONS = "cas"
 
+# NOTE(gabrielcocenza) top-level module to reference the same logger object across modules.
```
aieri, line 36: is this accurate? the various submodules use `logger = logging.getLogger(__name__)` which should yield submodule-specific names (ftr, I think either a single logger name or submodule-specific logger names are fine)

aieri (changes requested):
The code is overall fine, I just have a couple of nits.
Additionally, please add to the commit message why you made this change. Your intent is partially recorded in the linked bug, but I think it would be convenient for it to also be in the git history.

cou/cli.py
```diff
@@ -33,6 +33,9 @@
...
+logger = logging.getLogger("cou")
```
rgildein, line 37: Why we could not use root logger to set "debug" log level?
Can we be more smart here and get module name or logger from module itself? Using `"cou"` sting it not look to good to me.






























`

gabrielcocenza: The problem that I faced here is that we are setting the log on a submodule of the project. If we use `__name__`here this would result in `cou.cli` and this is not the `top` module that will reference to the same logger object. If we were doing this on the `__init__.py` of `cou` than it wouldn't be a problem.

Using the root logger it's not advisable from the [docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

> Logging to the root logger will make it difficult or impossible for the application developer to configure the logging verbosity or handlers of your library as they wish

cou/cli.py
gabrielcocenza: See the comment bellow why I used `cou` here. Otherwise the other modules won't take this logger object as reference

cou/cli.py
rgildein: In this file we should use `logger = logging.getLogger(__name__)` to do any logging. So we need to update line `logging.info("Logs of this execution can be found at %s", file_name)` to `logger.info("Logs of this execution can be found at %s", file_name)`.
You are right it's not advisable to use root logger for any logging, but we need to use root logger for any configuration. Without usage of root logger, we will lost all logs from different libraries such are `libjuju`, `websockets`, ...
I know that from `libjuju` and `websockets` there are to many logs, so I'll recommend to use filter for console_handler to log only `cou` logs.

cou/cli.py
```diff
@@ -34,7 +34,7 @@
 AVAILABLE_OPTIONS = "cas"
 
 # NOTE(gabrielcocenza) top-level module to reference the same logger object across modules.
-logger = logging.getLogger("cou")
+logger = logging.getLogger(__package__)
```
rgildein, line 37: There is no reason to not use `__name__` here, because it will make much more sense to see that logs in this file are coming from this file.

cou/cli.py
```diff
@@ -82,25 +82,27 @@ def setup_logging(log_level: str = "INFO") -> None:
     :rtype: None
     """
     log_formatter = logging.Formatter(
-        fmt="%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
+        fmt="%(asctime)s [%(name)s] [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
```
rgildein, line 85: I think that we should use use logger name in our logs, because it will be more simpler to see from where those logs are coming. However we are using filter for console handler, which will block all other logs, and I do not see how we will benefit from seeing logger name there.
We should use simple logging format for console handler and more complicated (with logger name or other attributes) for file handler.

rgildein (changes requested):

cou/cli.py
gabrielcocenza: True. I've fixed it.

cou/cli.py
gabrielcocenza: I've removed this part and now the logger is using `logging.getLogger(__name__)`

rgildein (approved):
LGTM

aieri (approved):
Looks good; just please ensure you squash and have a meaningful commit message (summary + details)

PR #21, opened 2023-07-25
Plan integration

gabrielcocenza:
- AppFactory to register and create Applications
- Application able to generate upgrade plan considering three big groups: pre-upgrade, upgrade and post_upgrade
- added set application config on juju_utils
- non-openstack charms or subordinate charms are not instantiated


cou/steps/plan.py
```diff
@@ -16,25 +16,94 @@
...
+    :type analysis_result: Analysis
+    :return: Tuple containing the current and next OpenStack release and a  list
+        of applications to be upgraded.
+    :rtype: Tuple[str, str, List[Application]]
+    """
+    # E.g: {"ussuri": {"keystone"}, "victoria": {"cinder"}}
+    os_versions: defaultdict[str, set] = defaultdict(set)
+
+    for app in analysis_result.apps:
+        # don't consider charms that can have multiple OpenStack releases to determine the oldest
```
aieri, line 70: is this necessary? You're taking the oldest available version in your list so applications that are compatible with multiple o7k codenames should not skew the result

gabrielcocenza: You are right. We can remove this logic because it won't affect if we always get the latest compatible version.

Thanks

gabrielcocenza:
It's considered out of scope of this PR:

- Plan generation for special charms like rabbitmq-server, ceph charms and etc.
- Disable unattended upgrades, apt-update/upgrade the machines keeping old configuration on pre-upgrade steps
- re-enable unattended upgrades on post-upgrade
- paused-single-unit strategy for data plane charms

aieri (changes requested):
Sorry I realize my old comments did not get published... they still seem relevant though

cou/steps/plan.py
```diff
@@ -16,25 +16,94 @@
...
+    UPGRADE_ORDER,
+    CompareOpenStack,
+    determine_next_openstack_release,
+)
 
 logger = logging.getLogger(__name__)
 
 
-async def generate_plan(args: Analysis) -> UpgradeStep:
+async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
```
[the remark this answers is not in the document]
esunar: We can do like that however plan makes generics things easier. Such as:
- Asking for confirmation on each step.
- Marking steps as serial or parallel without changing the executors
- For example there is a request to see how long each step took. 
- Dumping the overall upgrade plan
- ...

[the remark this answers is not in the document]
aieri: Yes, having the applications only return libjuju commands back to the execute step does make tweaking the execution easier, but I don't think it'd be impossible to do it within the application classes. For example:
* confirmation can be given from the execute step before running any (potentially asynchronous batch of) `Application.upgrade()` calls
* I'd be in favor of a single predetermined way to upgrade a given group of apps; too much complexity otherwise
* timing execution could be done within the Application class
* printing out the upgrade plan could be done by the Application classes logging their behavior + a dry-run parameter in the `upgrade()` method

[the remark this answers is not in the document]
esunar: But those are not the responsibility of Application itself. It just need to know how to upgrade application or unit.

cou/apps/app.py
```diff
@@ -38,7 +39,7 @@
 class AppFactory:
     """Factory class for Application objects."""
 
-    apps_type: Dict[str, Type[Application]] = {}
+    apps_type: dict[str, type[Application]] = {}
```
rgildein, line 42: Why we are changing it from `typing.Dict` to `dict` and `typing.Type` to `type`?

gabrielcocenza: I've been reading about `typing` and on [PEP 585](https://peps.python.org/pep-0585/) Dict, List, Type, Set and etc will be [deprecated](https://docs.python.org/3/library/typing.html#deprecated-aliases) from python 3.14. I think we should use the built-in types from now on.

rgildein: Thanks, that make sense.

rgildein (approved):
LGTM

esunar (approved):

aieri (changes requested):
I've done a high-level pass and before I go deeper I would like to comment on the overall design of the Application class.
Prior to this PR, the Application class only reflected the status of an application in the model.
In order to implement the step-generating logic, you are augmenting the class with properties that represent an implicit intent (`expected_current_channel`, `next_channel`, `next_os_release`, etc.). I think this muddles up status and intent. I would find it more cleanly separated if `Application` attributes/properties only captured the current reality of that application, and the plan-generating methods took a target argument: I believe it should be up to them to determine what the correct starting status should be, comparing it to the actual status, and outputting the steps needed to reach the target status (or raise an exception).
Also consider that Antelope is a SLURP release, so the axiom that start and target releases are only 1 release apart will soon no longer be true.

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+    def disable_action_managed(self, plan: UpgradeStep, parallel: bool = False) -> None:
+        """Disable action-managed-upgrade to upgrade as "all-in-one" strategy.
+
+        :param plan: Plan to disable action managed upgrade as sub step.
+        :type plan: UpgradeStep
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool
+        """
+        if self.config.get("action-managed-upgrade"):
+            if self.config["action-managed-upgrade"].get("value", False):
```
aieri, line 453: You could skip one if with
```suggestion
            if self.config.get("action-managed-upgrade", {}).get("value", False):
```

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
         """
         for origin in ("openstack-origin", "source"):
             if self.config.get(origin):
+                self.origin_setting = origin
                 return self.config[origin].get("value", "")
 
         logger.warning("Failed to get origin for %s, no origin config found", self.name)
         return ""
+
+    async def check_upgrade(self) -> None:
```
esunar, line 285: this should be private

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+                parallel=parallel,
+                function=async_upgrade_charm,
+                application_name=self.name,
+                channel=self.expected_current_channel,
+                model_name=self.model_name,
+                switch=switch,
+            )
+        )
+
+    def refresh_next_channel(self, plan: UpgradeStep, parallel: bool = False) -> None:
```
esunar, line 424: this should be private

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+                            "'action-managed-upgrade' to False."
+                        ),
+                        parallel=parallel,
+                        function=async_set_application_config,
+                        application_name=self.name,
+                        configuration={"action-managed-upgrade": False},
+                    )
+                )
+
+    def workload_upgrade(self, plan: UpgradeStep, parallel: bool = False) -> None:
```
esunar, line 467: This should be private

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+            )
+        else:
+            logger.warning(
+                "App: %s already have %s set to %s",
+                self.name,
+                self.origin_setting,
+                self.new_origin,
+            )
+
+    def reached_expected_target(self, plan: UpgradeStep, parallel: bool = False) -> None:
```
esunar, line 496: This should be private

esunar (changes requested):
The application should abstract us from upgrade details, such as refresh_current_channel or workload_upgrade. Users or inheritors of this class should only detail with pre/post/upgrade step generation and generate_upgrade_step. Other functions are implementation details for charms. I recommend to make other functions private.

Also prefer to return the UpgradeStep instead of modifying it in parameter.

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+
+    def can_generate_upgrade_plan(self) -> bool:
+        """Check if application can generate an upgrade plan.
+
+        Applications where it's not possible to identify the current OpenStack release,
+        should not generate upgrade plan.
+        :return: True if can generate a plan, else False.
+        :rtype: bool
+        """
+        if not self.current_os_release:
```
agileshaw, line 311: We can't generate upgrade plan if `self.next_os_release` is None, so I think we should:
```suggestion
        if not self.next_os_release:
```

cou/apps/app.py
```diff
@@ -173,6 +245,28 @@ def current_os_release(self) -> Optional[OpenStackRelease]:
...
+        return None
+
+    @property
+    def new_origin(self) -> Optional[str]:
+        """Return the new openstack-origin or source configuration.
+
+        :return: Repository from which to install.
+        :rtype: Optional[str]
+        """
+        if not self.current_os_release:
```
agileshaw, line 266: Again I think we should check for `self.next_os_release` instead

agileshaw (changes requested):

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+
+        if self.charm_origin == "cs":
+            description = f"Migration of '{self.name}' from charmstore to charmhub"
+            switch = f"ch:{self.charm}"
+
+        elif self.channel not in (self.expected_current_channel, self.next_channel):
+            description = (
+                f"Changing '{self.name}' channel from: '{self.channel}' "
+                f"to: '{self.expected_current_channel}'"
+            )
```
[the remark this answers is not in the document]
aieri: I've thought about this some more and came around it making sense for this code to be here. BootStack also confirmed they would appreciate automating cs → ch migrations.
I would however not call the function `_get_refresh_current_channel_plan` since it may yield a channel switch. How about `_get_refresh_current_release_charm_plan` or even simply `_get_refresh_charm_plan`? (and then `_get_refresh_next_channel_plan` becomes `_get_upgrade_charm_plan`)

Finally, we should think about what should happen when you encounter an application running on channel `ussuri/edge`. Should we automatically switch to `ussuri/stable`?

agileshaw (commented):
A few more things spotted now that I'm working on the class myself.

cou/steps/plan.py
```diff
@@ -23,17 +23,25 @@
...
 
     :param args: Analysis result.
     :type args: Analysis
     :return: Plan with all upgrade steps necessary based on the Analysis.
     :rtype: UpgradeStep
     """
-    logger.info(args)  # for placeholder
+    target = (
+        analysis_result.current_cloud_os_release.next_release
+        if analysis_result.current_cloud_os_release
```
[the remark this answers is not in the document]
gabrielcocenza: Now I'm raising an exception when target is None. It's possible a scenario where there is a current_cloud_os_release and not `current_cloud_os_release.next_release`. Ex. if the cloud is on caracal

cou/steps/plan.py
```diff
@@ -23,17 +23,25 @@
...
+        else None
+    )
     plan = UpgradeStep(description="Top level plan", parallel=False, function=None)
     plan.add_step(
         UpgradeStep(description="backup mysql databases", parallel=False, function=backup)
     )
+    for app in analysis_result.apps:
+        app_upgrade_plan = app.generate_upgrade_plan(target)
+        if app_upgrade_plan:
+            plan.add_step(app_upgrade_plan)
```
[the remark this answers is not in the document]
gabrielcocenza: Right now we are not differentiating what is control plane from what is data plane. In a serial upgrade if we respect the `UPGRADE_ORDER` and every application know what it should do, then we are ok. However, this will be necessary on parallel upgrades.

In any case I've added a upgrade plan step and the structure is almost the same.

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+        if not self.can_generate_upgrade_plan():
+            logger.warning("Aborting upgrade plan for '%s'", self.name)
+            return None
+        if self.current_os_release >= target_version:
+            logger.warning(
+                "Application: '%s' already on a newer version than %s. Aborting upgrade.",
+                self.name,
+                target,
+            )
+            return None
```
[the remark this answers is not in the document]
gabrielcocenza: In my mind it doesn't make sense to start generating the pre-upgrade plan to then halt on upgrade plan with a simple check to know if target is lesser than the current OpenStack release of an Application. Maybe this should be at `can_generate_upgrade_plan`?

cou/apps/app.py
```diff
@@ -25,12 +26,58 @@
...
+        def decorator(application: type[Application]) -> type[Application]:
+            for app_type in app_types:
+                cls.apps_type[app_type] = application
+            return application
+
+        return decorator
+
+
 @dataclass
 class Application:
```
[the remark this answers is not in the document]
gabrielcocenza: I thought about this and tried to decouple what is OpenStack from what is not and I faced some issues:

- hacluster and vault for example, is on the lookup table and doesn't have `origin` or `openstack-origin`. 
- neutron-openvswitch is an OpenStack charm and doesn't have `origin` or `openstack-origin.

I think it's easier for us ignoring the applications that we are not able to identify the `current_os_release` and/or `next_os_release` then having a separate class for then.

cou/apps/app.py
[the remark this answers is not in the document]
gabrielcocenza: I've changed the naming:

-  `_get_refresh_current_channel_plan` to `_get_refresh_charm_plan`
- `_get_refresh_next_channel_plan` to `_get_upgrade_charm_plan`

I liked this naming idea. Thanks

> Finally, we should think about what should happen when you encounter an application running on channel ussuri/edge. Should we automatically switch to ussuri/stable?

From the [docs](https://docs.openstack.org/charm-guide/latest/admin/upgrades/openstack.html#manual-intervention):
`By design, the latest stable charms will support the software changes related to the OpenStack services being upgraded.`

So, in theory we should use `stable`, but it might have the case of a bug-fix that not landed yet. We are showing that we are going to change from `x` to `y`. I think this is a choice from the operator to continue or not.

cou/apps/app.py

cou/utils/juju_utils.py
```diff
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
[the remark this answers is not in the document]
gabrielcocenza: Yes, thanks

cou/apps/app.py
```diff
@@ -135,6 +181,32 @@ def __str__(self) -> str:
...
+
+    @property
+    def next_channel(self) -> Optional[str]:
+        """Return the next channel based on the next OpenStack release.
+
+        If it's not possible to determine the next OpenStack release, the value is None.
+        :return: The next channel for the application.
+        :rtype: Optional[str]
+        """
+        if self.current_os_release:
```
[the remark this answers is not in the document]
gabrielcocenza: > I think you meant to check self.next_os_release here?
Yes, that is better.

I think the new properties are already covered on [unit tests](https://github.com/gabrielcocenza/charmed-openstack-upgrader-1/blob/plan/tests/unit/apps/test_app.py#L58-L62)

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+            if self.next_os_release:
```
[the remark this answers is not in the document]

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+            parallel=False,
+            function=None,
+        )
+        self.pre_upgrade_plan(upgrade_plan)
+        self.upgrade_plan(upgrade_plan)
+        self.post_upgrade_plan(upgrade_plan)
+        return upgrade_plan
+
+    def refresh_current_channel(self, plan: UpgradeStep, parallel: bool = False) -> None:
+        """Add Plan for refresh current channel.
```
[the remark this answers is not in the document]

cou/apps/app.py
gabrielcocenza: True, and this also indirectly cover the situation when `current_os_release` is None.

Done

cou/apps/app.py
```diff
@@ -182,7 +267,270 @@ def _get_os_origin(self) -> str:
...
+        if not self.next_os_release:
+            logger.warning("Not possible to generate upgrade plan for '%s'.", self.name)
+            logger.warning("'%s' might need manual intervention to upgrade.", self.name)
+            return False
+        return True
+
+    def pre_upgrade_plan(self) -> list[Optional[UpgradeStep]]:
+        """Pre Upgrade planning.
+
+        :param plan: Plan that will add pre upgrade as sub steps.
```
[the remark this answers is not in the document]

cou/apps/app.py
```diff
@@ -182,7 +267,270 @@ def _get_os_origin(self) -> str:
...
+        :type plan: UpgradeStep
+        """
+        return [self._get_refresh_current_channel_plan()]
+
+    def upgrade_plan(self) -> list[Optional[UpgradeStep]]:
+        """Upgrade planning.
+
+        :param plan: Plan that will add upgrade as sub steps.
+        :type plan: UpgradeStep
```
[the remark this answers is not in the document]

aieri (changes requested):
Submitting a partial review; I'll continue tomorrow

esunar (dismissed):

cou/apps/app.py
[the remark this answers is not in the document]
gabrielcocenza: I've removed can_generate_upgrade_plan and moved this logic to `upgrade_plan`.

Now we are generating a new exception named `HaltUpgradePlanGeneration` that halt if we identify something to skip the app

cou/apps/app.py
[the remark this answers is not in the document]
gabrielcocenza: After thinking I agree with you that using `current_os_release` to skip can be dangerous. I've changed the code and now non-openstack charms and subordinates (we need to take care of this in a later PR) are not instantiated and if there is no `current_os_release` an exception will happen `ApplicationError` or `MismatchedOpenStackVersions`

cou/apps/app.py
```diff
@@ -182,7 +276,279 @@ def _get_os_origin(self) -> str:
...
+            ),
+            parallel=False,
+            function=None,
+        )
+        self.pre_upgrade_plan(upgrade_plan)
+        self.upgrade_plan(upgrade_plan)
+        self.post_upgrade_plan(upgrade_plan)
+        return upgrade_plan
+
+    def refresh_current_channel(self, plan: UpgradeStep, parallel: bool = False) -> None:
```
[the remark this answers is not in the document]

cou/steps/plan.py
```diff
@@ -16,24 +16,47 @@
...
+async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
     """Generate plan for upgrade.
 
     :param args: Analysis result.
     :type args: Analysis
     :return: Plan with all upgrade steps necessary based on the Analysis.
     :rtype: UpgradeStep
     """
-    logger.info(args)  # for placeholder
+    target = (
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/steps/plan.py
```diff
@@ -16,24 +16,47 @@
...
+
+    upgrade_plan = UpgradeStep(
+        description="Application(s) upgrade plan", parallel=False, function=None
+    )
+    for app in analysis_result.apps:
+        try:
+            app_upgrade_plan = app.generate_upgrade_plan(target)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            logger.error("It was not possible to generate upgrade plan for '%s'.", app.name)
+            raise PlanError() from exc
```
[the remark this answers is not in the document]
gabrielcocenza: it wouldn't hide in the traceback, but in any case I don't think it was adding value, so I've removed

cou/apps/app.py
```diff
@@ -135,6 +180,28 @@ def __str__(self) -> str:
             yaml.dump(summary, stream)
             return stream.getvalue()
 
+    @property
+    def expected_current_channel(self) -> Optional[str]:
+        """Return the expected current channel based on the current OpenStack release.
+
+        Note that this is not necessary equal to the "channel" property if the application
```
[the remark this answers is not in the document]

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+            switch = f"ch:{self.charm}"
+        elif self.channel == self.expected_current_channel:
+            description = (
+                f"Refresh '{self.name}' to the latest revision of "
+                f"'{self.expected_current_channel}'"
+            )
+        elif os_track_release_channel >= target:
+            logger.warning(
+                (
+                    "App: %s already has the channel set for a bigger or equal OpenStack "
```
aieri, line 442:
```suggestion
                    "Skipping charm refresh for %s, its channel is already set to %s."
```

aieri: and then format with `self.channel` instead of `target`

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+        if self.charm_origin == "cs":
+            description = f"Migration of '{self.name}' from charmstore to charmhub"
+            switch = f"ch:{self.charm}"
+        elif self.channel == self.expected_current_channel:
+            description = (
+                f"Refresh '{self.name}' to the latest revision of "
+                f"'{self.expected_current_channel}'"
+            )
+        elif os_track_release_channel >= target:
+            logger.warning(
```
aieri, line 440: this may not be a warning, especially if it's the result of rerunning cou after a partial upgrade. How about `info`?

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+        """Upgrade planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation
+        :return: Plan that will add upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        if self.current_os_release >= target:
+            logger.warning(
```
aieri, line 356: it may be more useful to say which version the app is actually running. Also, `info` may be more appropriate in case we end up here after resuming an upgrade.

cou/apps/app.py
```diff
@@ -135,6 +231,27 @@ def __str__(self) -> str:
...
+        """Return the expected current channel based on the current OpenStack release.
+
+        Note that this is not necessarily equal to the "channel" property since it is
+        determined based on the workload version.
+        :return: The expected current channel for the application. E.g: ussuri/stable
+        :rtype: str
+        """
+        return f"{self.current_os_release}/stable"
+
+    def next_channel(self, target: OpenStackRelease) -> str:
```
aieri, line 245: nit, optional: `next_channel` implies an N+1 channel. `target_channel` would be more generic

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+                    f"Change charm config of '{self.name}' "
+                    f"'{self.origin_setting}' to '{self.new_origin(target)}'"
+                ),
+                parallel=parallel,
+                function=async_set_application_config,
+                application_name=self.name,
+                configuration={self.origin_setting: self.new_origin(target)},
+            )
+        logger.warning(
+            "App: %s already have %s set to %s",
```
aieri, line 531: WARN here is correct, since if the workload upgrade has already happened we should have hit that HaltUpgradePlanGeneration exception. I'd make the message more prominent:

```suggestion
            "Not triggering the workload upgrade of app %s: %s already set to %s",
```

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+    def generate_upgrade_plan(self, target: str) -> UpgradeStep:
+        """Generate full upgrade plan for an Application.
+
+        :param target: OpenStack codename to upgrade.
+        :type target: str
+        :return: Full upgrade plan if the Application is able to generate it.
+        :rtype: UpgradeStep
+        """
+        target_version = OpenStackRelease(target)
+        upgrade_plan = UpgradeStep(
```
aieri, line 387: nit: you already have an `upgrade_plan` function. Renaming this variable to a more unique name (e.g. `upgrade_step`) would make reading and searching through the code easier

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
...
+        app_status = status.applications.get(self.name)
+        units_not_upgraded = []
+        for unit in app_status.units.keys():
+            workload_version = app_status.units[unit].workload_version
+            compatible_os_versions = OpenStackCodenameLookup.lookup(self.charm, workload_version)
+            if target not in compatible_os_versions:
+                units_not_upgraded.append(unit)
+        if units_not_upgraded:
+            logger.error(
+                "App: '%s' has units: '%s' didn't upgrade to %s",
```
aieri, line 330: Unit names already contain the app name so this message could be more compact
```suggestion
                "Units '%s' failed to upgrade to %s",
```

cou/apps/app.py
```diff
@@ -182,7 +305,251 @@ def _get_os_origin(self) -> str:
         """
         for origin in ("openstack-origin", "source"):
             if self.config.get(origin):
+                self.origin_setting = origin
                 return self.config[origin].get("value", "")
 
         logger.warning("Failed to get origin for %s, no origin config found", self.name)
         return ""
+
+    async def _check_upgrade(self, target: OpenStackRelease) -> None:
```
aieri, line 314: failing to complete the upgrade is pretty serious; the overall cloud upgrade should probably halt. Though this will be the responsibility of the execute phase, should we think of throwing an exception here?

cou/utils/openstack.py
```diff
@@ -330,3 +330,18 @@ def lookup(cls, component: str, version: str) -> List[OpenStackRelease]:
             if version in version_range:
                 compatible_os_releases.append(OpenStackRelease(openstack_release))
         return compatible_os_releases
+
+    @classmethod
+    def charm_supported(cls, component: str) -> bool:
```
aieri, line 335: this function is so similar to `lookup`... could we get by without it? `AppFactory.create` could simply interpret an empty list returned by lookup as 'not supported'

cou/apps/app.py
```diff
@@ -17,25 +17,114 @@
...
+                        "supported for upgrading"
+                    ),
+                    name,
+                )
+                return None
+            return app_class(
+                name=name, status=status, config=config, model_name=model_name, charm=charm
+            )
+        logger.warning(
+            "'%s' it's not supported as an OpenStack related application and will be ignored.",
```
aieri, line 91:
```suggestion
            "'%s' is not a supported OpenStack related application and will be ignored.",
```

cou/apps/app.py
```diff
@@ -17,25 +17,114 @@
...
+                        "'%s' is a subordinate application and it's not currently "
+                        "supported for upgrading"
+                    ),
+                    name,
+                )
+                return None
+            return app_class(
+                name=name, status=status, config=config, model_name=model_name, charm=charm
+            )
+        logger.warning(
```
aieri, line 90: warning may be too much. There may be several spurious apps in an openstack model.

cou/apps/app.py
```diff
@@ -17,25 +17,114 @@
...
+    def register_application(
+        cls, app_types: list[str]
+    ) -> Callable[[type[OpenStackApplication]], type[OpenStackApplication]]:
+        """Register Application subclasses.
+
+        Use this method as decorator to register Applications that
+        have special needs.
+
+        Example:
+        @AppFactory.register_application(CHARM_TYPES["ceph"])
```
aieri, line 106: for an example, it's not too obvious that `CHARM_TYPES["ceph"]` returns a list. I'd make it more explicit (even if when we get to the Ceph class we will want to leverage the `CHARM_TYPES` dictionary.
Possibly:
```
ceph_types = ["ceph-mon", "ceph-fs", "ceph-radosgw", "ceph-osd"]

@AppFactory.register_application(ceph_types)
class Ceph(OpenStackApplication):
    pass
```

cou/steps/analyze.py
```diff
@@ -50,29 +55,33 @@ async def create(cls) -> Analysis:
...
                 status=app_status,
                 config=await async_get_application_config(app),
                 model_name=model_name,
+                charm=await extract_charm_name(app),
             )
             for app, app_status in juju_status.applications.items()
         }
-        upgradeable_apps = {app for app in apps if app.charm in UPGRADE_ORDER}
+        # remove non-supported charms that return None on AppFactory.create
+        apps.remove(None)
```
aieri, line 77:
```suggestion
        apps.discard(None)
```
You may not have non-OpenStack applications in your model (hence no None members of your set)

cou/steps/analyze.py
```diff
@@ -50,29 +55,33 @@ async def create(cls) -> Analysis:
...
+        upgradeable_apps = {app for app in apps if app and app.charm in UPGRADE_ORDER}
         unknown_apps = apps - upgradeable_apps
         upgradeable_apps_sorted = sorted(
             upgradeable_apps, key=lambda app: UPGRADE_ORDER.index(app.charm)
         )
-        return upgradeable_apps_sorted + list(unknown_apps)
+        # mypy complains that unknow_apps can have None, but we already removed None from apps
```
aieri, line 83:
```suggestion
        # mypy complains that unknown_apps can have None, but we already removed None from apps
```

cou/steps/plan.py
```diff
@@ -16,24 +16,49 @@
...
     )
+
+    upgrade_plan = UpgradeStep(
+        description="Application(s) upgrade plan", parallel=False, function=None
+    )
+    for app in analysis_result.apps:
+        try:
+            app_upgrade_plan = app.generate_upgrade_plan(target)
+        except HaltUpgradePlanGeneration:
+            # we do not care if application halt the upgrade plan generation for some known reason.
```
aieri, line 52:
```suggestion
            # we do not care if applications halt the upgrade plan generation for some known reason.
```

cou/steps/analyze.py
```diff
@@ -50,29 +55,33 @@ async def create(cls) -> Analysis:
...
+        return upgradeable_apps_sorted + list(unknown_apps)  # type: ignore
```
aieri, line 84: `plan.py` will try to generate plans for unknown_apps (where unknown = apps I can lookup in my app/codename table, but are not in the `UPGRADE_ORDER` - maybe we can find a better name). Is this ok?

aieri (changes requested):
This is looking almost ready for merge now! I have proposed several rewordings and asked a couple of questions. Aside from those I don't have major comments.

FTR I haven't had a chance to look at the tests yet.

cou/steps/analyze.py
aieri: and sorry for the double double negation :laughing:

cou/apps/app.py
gabrielcocenza: Agree. Changed to debug

cou/apps/app.py
gabrielcocenza: Changed to info

cou/apps/app.py
gabrielcocenza: It makes sense. Changed to `target_channel`

cou/apps/app.py

cou/apps/app.py
gabrielcocenza: I thought about that too. I've changed to raise `ApplicationError`

cou/apps/app.py

cou/steps/analyze.py
gabrielcocenza: I've changed to:

- `apps_to_upgrade_in_order` (apps that we know that have a order to upgrade)
- `sorted_apps_to_upgrade_in_order` (apps that we know that have a order to upgrade sorted in the right sequence)
- `disordered_upgrade_apps` (apps that can upgrade, but we don't care about the order)
- `disordered_upgrade_apps_sorted_by_name` (apps that can upgrade, but we don't care about the order and are sorted by charm name)

agileshaw (changes requested):
Overall LGTM. Just some small suggestions inline

cou/utils/openstack.py
aieri: That's right, you could still get an empty list from lookup even if the component name is in the table, I had not considered that. I also had not considered that the lookup function would not have worked unless you also pass a version. So... in retrospect my advice was not particularly good.

Now, your new solution works, but relies on side-effects of the lookup function, which now does double duty as both a get_compatible_versions and is_openstack_component... and makes me realize that the name `lookup` wasn't all that accurate in the first place.

Here's a different proposal that still remains DRY:
* keep the lookup function as an actual (possibly private) lookup function: it'd simply set up `cls._OPENSTACK_LOOKUP`, and return whatever `cls._OPENSTACK_LOOKUP.get(component)` finds (possibly an exception)
* write a "find compatible versions" function that uses the lookup (i.e. the old `lookup()`)
* write a separate "is this charm supported?" function that uses the lookup and returns a boolean

cou/steps/analyze.py
```diff
@@ -74,14 +74,20 @@ async def _populate(cls) -> list[OpenStackApplication]:
...
         }
         # remove non-supported charms that return None on AppFactory.create
-        apps.remove(None)
-        upgradeable_apps = {app for app in apps if app and app.charm in UPGRADE_ORDER}
-        unknown_apps = apps - upgradeable_apps
-        upgradeable_apps_sorted = sorted(
-            upgradeable_apps, key=lambda app: UPGRADE_ORDER.index(app.charm)
+        apps.discard(None)
+        apps_to_upgrade_in_order = {app for app in apps if app and app.charm in UPGRADE_ORDER}
+        disordered_upgrade_apps = apps - apps_to_upgrade_in_order
```
aieri, line 79: 'disordered' is not conveying what you're intending to (apps that can be upgraded in any order). Perhaps a very simple `other_o7k_apps` would be enough.

I also suggest explaining the upgrade order in the docstring. For example:
```
Applications that must be upgraded in a specific order will be returned first, followed by applications that can be upgraded in any order.
Applications that are not supported will be ignored.
```

cou/steps/analyze.py
```diff
@@ -74,14 +74,20 @@ async def _populate(cls) -> list[OpenStackApplication]:
...
+        sorted_apps_to_upgrade_in_order = sorted(
+            apps_to_upgrade_in_order, key=lambda app: UPGRADE_ORDER.index(app.charm)
+        )
+        disordered_upgrade_apps_sorted_by_name = sorted(
```
aieri, line 83: I think you're sorting them by name so that the plan doesn't randomly change over multiple executions, right? I suggest adding a quick comment explaining why the sorting is relevant.

aieri (changes requested):

tests/unit/utils/test_openstack.py
```diff
@@ -65,6 +66,13 @@ def test_get_compatible_openstack_codenames(charm, workload_versions, results):
         assert result == actual
 
 
+def test_lookup_raises_keyerror_unknown_component():
+    with pytest.raises(KeyError):
+        OpenStackCodenameLookup.lookup("my_charm", "13.1.2")
+    with pytest.raises(KeyError):
+        OpenStackCodenameLookup.lookup("my_charm")
```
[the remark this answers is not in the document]
gabrielcocenza: I ended up removing this unit tests with the changes on OpenStackCodenameLookup

cou/apps/app.py
```diff
@@ -242,8 +247,8 @@ def expected_current_channel(self) -> str:
         """
         return f"{self.current_os_release}/stable"
 
-    def next_channel(self, target: OpenStackRelease) -> str:
-        """Return the next channel based on the target passed.
+    def target_channel(self, target: OpenStackRelease) -> str:
```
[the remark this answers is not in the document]
gabrielcocenza: Based on the stand up conversation, we decided to keep `target_channel`

cou/utils/openstack.py
gabrielcocenza: I ended up changing the method `lookup` to ` find_compatible_versions` and creating a new one named `is_charm_supported` that is used in the `find_compatible_versions` and populates `_OPENSTACK_LOOKUP` if necessary.

cou/apps/app.py
```diff
@@ -353,9 +358,13 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
         :rtype: list[Optional[UpgradeStep]]
         """
         if self.current_os_release >= target:
-            logger.warning(
-                "Application: '%s' already on a newer version than %s. Ignoring.",
+            logger.info(
+                (
+                    "Application: '%s' already running %s that is equal or bigger "
+                    "version than %s. Ignoring."
```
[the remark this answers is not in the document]
aieri: @gabrielcocenza you've missed the suggestion on line 369, was that intentional?

aieri (approved):

agileshaw (approved):
LGTM

esunar (approved):

PR #22, opened 2023-07-27
Move Application class to its own module

gabrielcocenza:
- separate the Application class from the analyze
- makes easier the review on next patch containing changes on this class because github doesn't recognize when a file is moved.


aieri (approved):
LGTM. `diff cou/steps/application/app.py <(git show main:cou/steps/analyze.py)` confirms it's a direct class move.
That said, is `cou.steps.application.app` the best path? The Application class is not a step.

esunar (changes requested):
`__init__.py` file missing : blocker
cou/apps may be better location assuming there will be several apps: not blocker

esunar (approved):

PR #23, opened 2023-07-27
Set interactive prompt and dry run as default.

gabrielcocenza:
- currently, the cli is running as non-interactive as default and this is not good for user experience that can start upgrading the cloud.

- changed the flag "--dry-run" to "--run". This way, if the user run the cli without flag, the plan will be printed.

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -53,6 +53,7 @@ def parse_args(args: Any) -> argparse.Namespace:
         "--run",
         help="Use this flag to run the upgrade, otherwise just print out the upgrade steps.",
         action="store_true",
+        default=False,
     )
     parser.add_argument(
         "--log-level",
@@ -69,7 +70,7 @@ def parse_args(args: Any) -> argparse.Namespace:
         help="Set the model to operate on.",
     )
     parser.add_argument(
-        "--non-interactive", help="Run upgrade without prompt.", action="store_true"
+        "--non-interactive", help="Run upgrade without prompt.", action="store_true", default=False
     )
 
     return parser.parse_args(args)
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -141,14 +141,9 @@ async def apply_plan(upgrade_plan: UpgradeStep, non_interactive: bool) -> None:
     :param upgrade_plan: Plan to be executed on steps.
     :type upgrade_plan: UpgradeStep
     """
-    if non_interactive:
-        await upgrade_plan.run()
-        for sub_step in upgrade_plan.sub_steps:
-            await sub_step.run()
-        return
     result = "X"
     while result.casefold() not in AVAILABLE_OPTIONS:
-        result = input(prompt(upgrade_plan.description)).casefold()
+        result = "c" if non_interactive else input(prompt(upgrade_plan.description)).casefold()
         match result:
             case "c":
                 await upgrade_plan.run()
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -50,9 +50,8 @@ def parse_args(args: Any) -> argparse.Namespace:
         exit_on_error=False,
     )
     parser.add_argument(
-        "--dry-run",
-        default=False,
-        help="Do not run the upgrade just print out the steps.",
+        "--run",
+        help="Use this flag to run the upgrade, otherwise just print out the upgrade steps.",
         action="store_true",
     )
     parser.add_argument(
@@ -69,7 +68,9 @@ def parse_args(args: Any) -> argparse.Namespace:
         dest="model_name",
         help="Set the model to operate on.",
     )
-    parser.add_argument("--interactive", default=True, help="Sets the interactive prompts")
+    parser.add_argument(
+        "--non-interactive", help="Run upgrade without prompt.", action="store_true"
+    )
 
     return parser.parse_args(args)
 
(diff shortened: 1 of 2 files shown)


esunar (approved):
LGTM, this approach looks better

cou/cli.py
```diff
@@ -134,20 +135,25 @@ def normal(text: str) -> str:
...
     :type upgrade_plan: UpgradeStep
     """
+    if non_interactive:
+        await upgrade_plan.run()
+        for sub_step in upgrade_plan.sub_steps:
+            await sub_step.run()
+        return
     result = "X"
     while result.casefold() not in AVAILABLE_OPTIONS:
         result = input(prompt(upgrade_plan.description)).casefold()
```
rgildein, line 151: not-blocking: What about
```suggestion
        result = "c" if non_interactive else input(prompt(upgrade_plan.description)).casefold()
```
This will be more readable to me and we will not define logic of run and iteration in two places.

cou/cli.py
```diff
@@ -134,20 +135,25 @@ def normal(text: str) -> str:
...
+async def apply_plan(upgrade_plan: UpgradeStep, non_interactive: bool) -> None:
     """Apply the plan for upgrade.
 
     :param upgrade_plan: Plan to be executed on steps.
     :type upgrade_plan: UpgradeStep
     """
+    if non_interactive:
+        await upgrade_plan.run()
+        for sub_step in upgrade_plan.sub_steps:
+            await sub_step.run()
```
rgildein, line 147: Is this ok? We are not running any `sub_sub_step`. What I mean is that we are not doing
```python
for sub_sub_step in upgrade_plan.sub_step.sub_steps:
            await sub_sub_step.run()
```

esunar: Yes we should make this also recursive.

tests/unit/test_cli.py
```diff
@@ -7,15 +7,26 @@
...
+    expected_loglevel = expected_results[1]
+    expected_non_interactive = expected_results[2]
+
+    assert parsed_args.run == expected_run
+    assert parsed_args.loglevel == expected_loglevel
+    assert parsed_args.non_interactive == expected_non_interactive
 
     with pytest.raises(ArgumentError):
-        args = parse_args(["--dry-run", "--log-level=DDD"])
+        args = parse_args(["--log-level=DDD"])
```
rgildein, line 29: This should be in different test. it's not good practice test different behavior in one test and even worse here we are testing this in each parametrize run (3 times).

tests/unit/test_cli.py
```diff
@@ -7,15 +7,26 @@
...
+    ],
+)
+def test_parse_args(args, expected_results):
     parsed_args = parse_args(args)
-    assert parsed_args.dry_run
-    assert parsed_args.loglevel == "DEBUG"
-    assert parsed_args.interactive
+    expected_run = expected_results[0]
+    expected_loglevel = expected_results[1]
+    expected_non_interactive = expected_results[2]
```
rgildein, line 22: not-blocking: What is purpose to use single `expected_results` and than use it only in single place to unpack it to multiple variable. Consider to use ` expected_run`, `expected_loglevel` and `expected_non_interactive` as inputs for this tests.

tests/unit/test_cli.py
```diff
@@ -53,14 +64,17 @@ async def test_entrypoint_dry_run():
         "cou.cli.generate_plan"
     ) as mock_generate_plan, patch("cou.cli.apply_plan"), patch("cou.cli.Analysis.create"), patch(
         "builtins.print"
-    ) as mock_print:
+    ) as mock_print, patch(
+        "cou.cli.apply_plan"
+    ) as mock_apply_plan:
```
rgildein, line 69: not-blocking: Can we move all this mocking as decorators? It will be much more readable like this.

tests/unit/test_cli.py
```diff
@@ -71,14 +85,29 @@ async def test_entrypoint_real_run():
...
-        mock_parse_args.return_value.dry_run = False
+        mock_parse_args.return_value.run = True
 
         result = await entrypoint()
 
         assert not result
         mock_apply_plan.assert_called_once()
 
 
+@pytest.mark.asyncio
```
rgildein, line 96: Can we add also test-case when we will test behavior running also sub sub step? You accidentally broke than in your previous commit and no tests were broken. Thanks

rgildein (commented):
Only couple of comment to our unit tests.

tests/unit/test_cli.py
gabrielcocenza: Solved

tests/unit/test_cli.py
gabrielcocenza: Solved

tests/unit/test_cli.py
gabrielcocenza: I've changed to use the `mocker` module.

tests/unit/test_cli.py
gabrielcocenza: I've added a third layer of sub_step to prove that works. Thanks

cou/cli.py
```diff
@@ -50,9 +50,8 @@ def parse_args(args: Any) -> argparse.Namespace:
         exit_on_error=False,
     )
     parser.add_argument(
-        "--dry-run",
-        default=False,
-        help="Do not run the upgrade just print out the steps.",
+        "--run",
```
esunar, line 53: I think we should add `default=False` to make it more understandable

esunar (approved):
Overall LGTM with one minor comment

rgildein (approved):
LGTM

PR #24, opened 2023-08-01
Move Application unit tests to its own module

gabrielcocenza:
- #22 moved Application data class object to its own module, but unit tests were not moved and were together with the analyze.

### DIFF
--- a/tests/unit/apps/test_app.py
+++ b/tests/unit/apps/test_app.py
@@ -1,0 +1,114 @@
+# Copyright 2023 Canonical Limited.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


esunar (approved):

tests/unit/apps/test_app.py
```diff
@@ -0,0 +1,114 @@
...
+        # application is from charm store
+        expected_charm_origin = "cs"
+        app_status = status["keystone_ussuri_cs"]
+
+    elif condition == "keystone_wallaby":
+        expected_units = units["units_wallaby"]
+        expected_charm_origin = "ch"
+        app_config = config["openstack_wallaby"]
+        app_status = status["keystone_wallaby"]
+        expected_os_origin = "cloud:focal-wallaby"
```
rgildein, line 76: not-blocking: I think it would make more sense to create 4 separate tests instead of parametrizing and than using if conditions.
Example:
```python
def assert_application(app, exp_name, exp_status, ...):
    assert app.name == exp_name

def test_keystone_ussuri(status, config, mocker, units):
    app_status = status["keystone_ussuri"]
    app_config = config["openstack_ussuri"]
    expected_charm_origin = "ch"
    expected_os_origin = "distro"
    expected_units = units["units_ussuri"]

    app = Application("keystone", app_status, app_config, "my_model")
     
    assert_application(app, "keystone", ....)
```

rgildein (approved):
It's not wrong approach, but I think it would be more readable to create separate tests.

tests/unit/apps/test_app.py
gabrielcocenza: Done

PR #25, opened 2023-08-02
Analyze and Applications can determine current and next OpenStack release

gabrielcocenza:
- added helpers on openstack utils to determine next openstack
  release, list of the expected upgrade order, capability of compare
  openstack releases by its codename.
- Analyze can show current and next openstack version of the cloud
- added series, current and next openstack release properties on
  Application.
- analyze skip subordinate charms and those with unknown OpenStack
  release version.
- raise exception on mismatched OpenStack versions in an app.
- skip non-openstack charms on current cloud OpenStack release from Analyze.

This PR should be merged after #24

### DIFF
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -150,6 +150,7 @@ class Application:
 
         :raises MismatchedOpenStackVersions: Raise MismatchedOpenStackVersions if units of
             an application are running mismatched OpenStack versions.
+        :return: OpenStackRelease object
         :rtype: OpenStackRelease
         """
         os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -24,6 +24,7 @@ from typing import Any, Optional
 from juju.client._definitions import ApplicationStatus
 from ruamel.yaml import YAML
 
+from cou.exceptions import MismatchedOpenStackVersions
 from cou.utils.juju_utils import extract_charm_name_from_url
 from cou.utils.openstack import OpenStackCodenameLookup, OpenStackRelease
 
@@ -82,7 +83,12 @@ class Application:
                 unit_os_version = max(compatible_os_versions)
                 self.units[unit]["os_version"] = unit_os_version
             else:
-                self.units[unit]["os_version"] = ""
+                self.units[unit]["os_version"] = None
+                logger.warning(
+                    "No compatible OpenStack versions were found to %s with workload version %s",
+                    self.name,
+                    workload_version,
+                )
 
     def __hash__(self) -> int:
         """Hash magic method for Application.
@@ -118,7 +124,7 @@ class Application:
                 "units": {
                     unit: {
                         "workload_version": details.get("workload_version", ""),
-                        "os_version": str(details.get("os_version", "")),
+                        "os_version": str(details.get("os_version")),
                     }
                     for unit, details in self.units.items()
                 },
@@ -146,14 +152,24 @@ class Application:
         :rtype: OpenStackRelease
         """
         os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
+        if not os_versions:
+            # TODO(gabrielcocenza) subordinate charms doesn't have units on ApplicationStatus and
+            # return an empty set. This should be handled by a future implementation of
+            # subordinate applications class.
+            return None
         if len(os_versions) == 1:
-            return list(os_versions)[0]
-        logger.warning(
-            "%s has more than one OpenStack version: %s. Setting current release to None.",
+            return os_versions.pop()
+        # NOTE (gabrielcocenza) on applications that use single-unit or paused-single-unit
+        # upgrade methods, more than one version can be found.
+        logger.error(
+            (
+                "Units of application %s are running mismatched OpenStack versions: %s. "
+                "Setting current release to None."
+            ),
             self.name,
             os_versions,
         )
-        return None
+        raise MismatchedOpenStackVersions()
 
     def _get_os_origin(self) -> str:
         """Get application configuration for openstack-origin or source.
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -19,7 +19,7 @@ import logging
 from collections import defaultdict
 from dataclasses import dataclass, field
 from io import StringIO
-from typing import Any
+from typing import Any, Optional
 
 from juju.client._definitions import ApplicationStatus
 from ruamel.yaml import YAML
@@ -53,11 +53,6 @@ class Application:
     :param units: Units representation of an application.
         E.g: {"keystone/0": {'os_version': 'victoria', 'workload_version': '2:18.1'}}
     :type units: defaultdict[str, dict]
-    :param current_os_release: Current OpenStack release codename of the application
-    :type current_os_release: str, defaults to ""
-    :param next_os_release: Next OpenStack release codename of the application
-    :param os_versions: OpenStack releases codename encountered on the application
-    :type os_versions: set
     """
 
     # pylint: disable=too-many-instance-attributes
@@ -71,9 +66,6 @@ class Application:
     os_origin: str = ""
     channel: str = ""
     units: defaultdict[str, dict] = field(default_factory=lambda: defaultdict(dict))
-    current_os_release: str = ""
-    next_os_release: str = ""
-    os_versions: set = field(default_factory=set)
 
     def __post_init__(self) -> None:
         """Initialize the Application dataclass."""
@@ -87,15 +79,10 @@ class Application:
             compatible_os_versions = OpenStackCodenameLookup.lookup(self.charm, workload_version)
             # NOTE(gabrielcocenza) get the latest compatible OpenStack version.
             if compatible_os_versions:
-                unit_os_version = compatible_os_versions[-1]
+                unit_os_version = max(compatible_os_versions)
                 self.units[unit]["os_version"] = unit_os_version
-                self.os_versions.add(unit_os_version)
             else:
                 self.units[unit]["os_version"] = ""
-
-        if len(self.os_versions) == 1:
-            self.current_os_release = list(self.os_versions)[0]
-            self.next_os_release = OpenStackRelease(self.current_os_release).next_release
 
     def __hash__(self) -> int:
         """Hash magic method for Application.
@@ -131,7 +118,7 @@ class Application:
                 "units": {
                     unit: {
                         "workload_version": details.get("workload_version", ""),
-                        "os_version": details.get("os_version", ""),
+                        "os_version": str(details.get("os_version", "")),
                     }
                     for unit, details in self.units.items()
                 },
@@ -150,6 +137,18 @@ class Application:
         :rtype: str
         """
         return self.status.series
+
+    @property
+    def current_os_release(self) -> Optional[OpenStackRelease]:
+        """Current OpenStack Release of the application.
+
+        :return: OpenStackRelease object
+        :rtype: OpenStackRelease
+        """
+        os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
+        if len(os_versions) == 1:
+            return list(os_versions)[0]
+        return None
 
     def _get_os_origin(self) -> str:
         """Get application configuration for openstack-origin or source.
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -148,6 +148,11 @@ class Application:
         os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
         if len(os_versions) == 1:
             return list(os_versions)[0]
+        logger.warning(
+            "%s has more than one OpenStack version: %s. Setting current release to None.",
+            self.name,
+            os_versions,
+        )
         return None
 
     def _get_os_origin(self) -> str:
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -25,10 +25,7 @@ from juju.client._definitions import ApplicationStatus
 from ruamel.yaml import YAML
 
 from cou.utils.juju_utils import extract_charm_name_from_url
-from cou.utils.openstack import (
-    OpenStackCodenameLookup,
-    determine_next_openstack_release,
-)
+from cou.utils.openstack import OpenStackCodenameLookup, OpenStackRelease
 
 logger = logging.getLogger(__name__)
 
@@ -98,7 +95,7 @@ class Application:
 
         if len(self.os_versions) == 1:
             self.current_os_release = list(self.os_versions)[0]
-            _, self.next_os_release = determine_next_openstack_release(self.current_os_release)
+            self.next_os_release = OpenStackRelease(self.current_os_release).next_release
 
     def __hash__(self) -> int:
         """Hash magic method for Application.
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -25,7 +25,10 @@ from juju.client._definitions import ApplicationStatus
 from ruamel.yaml import YAML
 
 from cou.utils.juju_utils import extract_charm_name_from_url
-from cou.utils.openstack import OpenStackCodenameLookup
+from cou.utils.openstack import (
+    OpenStackCodenameLookup,
+    determine_next_openstack_release,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -45,14 +48,19 @@ class Application:
     :param charm: Name of the charm.
     :type charm: str
     :param charm_origin: Origin of the charm (local, ch, cs and etc.), defaults to ""
-    :type charm_origin: str, optional
+    :type charm_origin: str, defaults to ""
     :param os_origin: OpenStack origin of the application. E.g: cloud:focal-wallaby, defaults to ""
-    :type os_origin: str, optional
+    :type os_origin: str, defaults to ""
     :param channel: Channel that the charm tracks. E.g: "ussuri/stable", defaults to ""
-    :type channel: str, optional
+    :type channel: str, defaults to ""
     :param units: Units representation of an application.
         E.g: {"keystone/0": {'os_version': 'victoria', 'workload_version': '2:18.1'}}
     :type units: defaultdict[str, dict]
+    :param current_os_release: Current OpenStack release codename of the application
+    :type current_os_release: str, defaults to ""
+    :param next_os_release: Next OpenStack release codename of the application
+    :param os_versions: OpenStack releases codename encountered on the application
+    :type os_versions: set
     """
 
     # pylint: disable=too-many-instance-attributes
@@ -66,6 +74,9 @@ class Application:
     os_origin: str = ""
     channel: str = ""
     units: defaultdict[str, dict] = field(default_factory=lambda: defaultdict(dict))
+    current_os_release: str = ""
+    next_os_release: str = ""
+    os_versions: set = field(default_factory=set)
 
     def __post_init__(self) -> None:
         """Initialize the Application dataclass."""
@@ -79,9 +90,15 @@ class Application:
             compatible_os_versions = OpenStackCodenameLookup.lookup(self.charm, workload_version)
             # NOTE(gabrielcocenza) get the latest compatible OpenStack version.
             if compatible_os_versions:
-                self.units[unit]["os_version"] = compatible_os_versions[-1]
+                unit_os_version = compatible_os_versions[-1]
+                self.units[unit]["os_version"] = unit_os_version
+                self.os_versions.add(unit_os_version)
             else:
                 self.units[unit]["os_version"] = ""
+
+        if len(self.os_versions) == 1:
+            self.current_os_release = list(self.os_versions)[0]
+            _, self.next_os_release = determine_next_openstack_release(self.current_os_release)
 
     def __hash__(self) -> int:
         """Hash magic method for Application.
@@ -127,6 +144,15 @@ class Application:
         with StringIO() as stream:
             yaml.dump(summary, stream)
             return stream.getvalue()
+
+    @property
+    def series(self) -> str:
+        """Ubuntu series of the application.
+
+        :return: Ubuntu series of application. E.g: focal
+        :rtype: str
+        """
+        return self.status.series
 
     def _get_os_origin(self) -> str:
         """Get application configuration for openstack-origin or source.
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -148,7 +148,8 @@ class Application:
     def current_os_release(self) -> Optional[OpenStackRelease]:
         """Current OpenStack Release of the application.
 
-        :return: OpenStackRelease object
+        :raises MismatchedOpenStackVersions: Raise MismatchedOpenStackVersions if units of
+            an application are running mismatched OpenStack versions.
         :rtype: OpenStackRelease
         """
         os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -164,7 +164,7 @@ class Application:
         logger.error(
             (
                 "Units of application %s are running mismatched OpenStack versions: %s. "
-                "Setting current release to None."
+                "This is not currently handled."
             ),
             self.name,
             os_versions,
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -51,7 +51,7 @@ class Analysis:
 
     @classmethod
     async def _populate(cls) -> List[Application]:
-        """Generate the applications model.
+        """Analyze the applications in the model.
 
         :return: Application objects with their respective information.
         :rtype: List[Application]
@@ -67,7 +67,12 @@ class Analysis:
             )
             for app, app_status in juju_status.applications.items()
         }
-        return sorted(apps, key=lambda app: UPGRADE_ORDER.index(app.charm))
+        upgradeable_apps = {app for app in apps if app.charm in UPGRADE_ORDER}
+        unknown_apps = apps - upgradeable_apps
+        upgradeable_apps_sorted = sorted(
+            upgradeable_apps, key=lambda app: UPGRADE_ORDER.index(app.charm)
+        )
+        return upgradeable_apps_sorted + list(unknown_apps)
 
     def __str__(self) -> str:
         """Dump as string.
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -17,20 +17,45 @@ from __future__ import annotations
 
 import logging
 import os
-from dataclasses import dataclass
-from typing import Iterable
+from collections import defaultdict
+from dataclasses import dataclass, field
+from typing import Iterable, List, Optional
 
 from cou.apps.app import Application
 from cou.utils.juju_utils import async_get_application_config, async_get_status
+from cou.utils.openstack import (
+    LTS_SERIES,
+    SPECIAL_CHARMS,
+    UPGRADE_ORDER,
+    CompareOpenStack,
+    determine_next_openstack_release,
+)
 
 logger = logging.getLogger(__name__)
 
 
 @dataclass
 class Analysis:
-    """Analyze result."""
+    """Analyze result.
+
+    :param apps: Applications in the model
+    :type apps:  Iterable[Application]
+    :param apps_to_upgrade: Applications to upgrade in the model
+    :type apps_to_upgrade: Optional[List[Application]]
+    :param os_versions: Dictionary containing OpenStack codenames and set of Applications
+    :type os_versions:  defaultdict[str, set]
+    """
 
     apps: Iterable[Application]
+    apps_to_upgrade: Optional[List[Application]] = None
+    os_versions: defaultdict[str, set] = field(default_factory=lambda: defaultdict(set))
+
+    def __post_init__(self) -> None:
+        """Initialize the Analysis dataclass."""
+        for app in self.apps:
+            if app.current_os_release:
+                self.os_versions[app.current_os_release].add(app)
+        self.apps_to_upgrade = self.determine_apps_to_upgrade()
 
     @classmethod
     async def create(cls) -> Analysis:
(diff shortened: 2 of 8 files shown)


esunar (approved):
LGTM with one question

cou/apps/app.py
```diff
@@ -80,10 +91,16 @@ def __post_init__(self) -> None:
...
             # NOTE(gabrielcocenza) get the latest compatible OpenStack version.
             if compatible_os_versions:
-                self.units[unit]["os_version"] = compatible_os_versions[-1]
+                unit_os_version = compatible_os_versions[-1]
+                self.units[unit]["os_version"] = unit_os_version
+                self.os_versions.add(unit_os_version)
             else:
                 self.units[unit]["os_version"] = ""
 
+        if len(self.os_versions) == 1:
```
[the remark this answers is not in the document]
gabrielcocenza: In that case `current_os_release` and `next_os_release` will have the value as empty string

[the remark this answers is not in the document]
gabrielcocenza: This behavior is described on unit tests and can be found at the [test_application_different_wl](https://github.com/gabrielcocenza/charmed-openstack-upgrader-1/blob/ac3394481b4d99d89a8d26d6256438500538c923/tests/unit/apps/test_app.py#L96-L131) unit test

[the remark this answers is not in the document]
esunar: Yes I saw it however this means we will not attempt to upgrade it. We need to discuss the details.

cou/utils/openstack.py
```diff
@@ -147,3 +229,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+
+    :param release: the release to use as the base
+    :type release: str
+    :returns: the release tuple immediately after the current one.
+    :rtype: Tuple[str, str]
+    :raises: ValueError if the current release doesn't actually exist
+    """
+    old_index = list(OPENSTACK_CODENAMES.values()).index(release)
+    new_index = old_index + 1
+    return list(OPENSTACK_CODENAMES.items())[new_index]
```
rgildein, line 247: This will raise an IndexError if I do `determine_next_openstack_release("antelope")` and I think we should raise other exception or the `("2023.1", "antelope")`.
Also we are using mostly values from `OPENSTACK_CODENAMES`, so why not switch keys and values here? Or maybe not define it as dict, but as two lists?

cou/utils/openstack.py
```diff
@@ -147,3 +229,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+    alphanumerically ordered.  e.g. OpenStack or Ubuntu releases AFTER the
+    z-wrap.
+    """
+
+    _list: Optional[List] = None
+
+    def __init__(self, item):
+        """Do init."""
+        if self._list is None:
+            raise ValueError("Must define the _list in the class definition!")
```
rgildein, line 263: We should use
```python
@property
@abstractmethod
def _list(self):
        pass
```

Instead of this.

cou/utils/openstack.py
```diff
@@ -147,3 +229,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+
+        assert s_newton > s_mitaka
+
+        :returns: <string>
+        """
+        # pylint: disable=unsubscriptable-object
+        return self._list[self.index]
+
+
+class CompareOpenStack(BasicStringComparator):
```
rgildein, line 317: Why we are defining the `OPENSTACK_CODENAMES` as ordered dict and than never using it full potential? Why not defining our own [UserList](https://docs.python.org/3/library/collections.html#collections.UserList) or [UserDict](https://docs.python.org/3/library/collections.html#collections.UserList) object. Like this we can provide whole functionality in single object and not making `determine_next_openstack_release`, `BasicStringComparator` and `CompareOpenStack`.

We can do
```python
OPENSTACK_CODENAMES.get("xena").next_release()
OPENSTACK_CODENAMES["xena"].next_relase()

OPENSTACK_CODENAMES["xena"] > OPENSTACK_CODENAMES["yoga"]
```

cou/utils/openstack.py
aieri: or even an Enum with a `next()` method (so you can catch being out of range if needed). Very rough sketch (would need to be polished):
```
class OpenStackVersion(Enum):
    ussuri = '2020.1'
    victoria = '2020.2'
    wallaby = '2021.1'

    def next(self):
        cur_index = list(self.__class__).index(self)
        try:
            return list(self.__class__)[cur_index+1]
        except IndexError:
            raise SomeCustomException
```

cou/utils/openstack.py
```diff
@@ -147,3 +230,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+        :type os_release: str
+        :raises ValueError: Raises ValueError if OpenStack codename is unknown.
+        """
+        self.os_release = os_release
+        try:
+            self.index = self.openstack_codenames.index(os_release)
+        except ValueError as exc:
+            raise ValueError(
+                f"OpenStack '{os_release}' is not in '{self.openstack_codenames}'"
+            ) from exc
```
rgildein, line 259: not-blocking: I think it would be nicer to os_release.setter, so the init will look cleaner + I can do `OpenStackRelease.os_release = "my new os release"`.

cou/utils/openstack.py
gabrielcocenza: Good idea. I've added this kind of functionality in the new commit

cou/utils/openstack.py
```diff
@@ -39,6 +39,88 @@
...
+        ("2018.2", "rocky"),
+        ("2019.1", "stein"),
+        ("2019.2", "train"),
+        ("2020.1", "ussuri"),
+        ("2020.2", "victoria"),
+        ("2021.1", "wallaby"),
+        ("2021.2", "xena"),
+        ("2022.1", "yoga"),
+        ("2022.2", "zed"),
+        ("2023.1", "antelope"),
```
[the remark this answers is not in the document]
gabrielcocenza: Added.

cou/utils/openstack.py
```diff
@@ -147,3 +230,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+        try:
+            self.index = self.openstack_codenames.index(os_release)
+        except ValueError as exc:
+            raise ValueError(
+                f"OpenStack '{os_release}' is not in '{self.openstack_codenames}'"
+            ) from exc
+
+    def __eq__(self, other):
+        """Do equals."""
+        assert isinstance(other, (str, self.__class__))
```
rgildein, line 263: This does not make sense to me, if I do `OpenStackRelease("yoga") == None` it raise AssertError.
```python
def __eq__(self, other):
    if not isinstance(other, Partial):
        return NotImplemented

    return self.index == other.index
```

cou/utils/openstack.py
```diff
@@ -147,3 +230,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+        return self.index == self.openstack_codenames.index(other)
+
+    def __ne__(self, other):
+        """Do not equals."""
+        return not self.__eq__(other)
+
+    def __lt__(self, other):
+        """Do less than."""
+        assert isinstance(other, (str, self.__class__))
```
rgildein, line 272: same here

cou/utils/openstack.py
```diff
@@ -147,3 +230,98 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+        return self.index < self.openstack_codenames.index(other)
+
+    def __ge__(self, other):
+        """Do greater than or equal."""
+        return not self.__lt__(other)
+
+    def __gt__(self, other):
+        """Do greater than."""
+        assert isinstance(other, (str, self.__class__))
```
rgildein, line 281: same here

tests/unit/utils/test_openstack.py
```diff
@@ -76,3 +76,75 @@ def test_generate_lookup(service):
...
+        result = OpenStackRelease(release_1) <= release_2
+        result_alternative = OpenStackRelease(release_1) <= OpenStackRelease(release_2)
+    elif comparison == "ge":
+        result = OpenStackRelease(release_1) >= release_2
+        result_alternative = OpenStackRelease(release_1) >= OpenStackRelease(release_2)
+    elif comparison == "gt":
+        result = OpenStackRelease(release_1) > release_2
+        result_alternative = OpenStackRelease(release_1) > OpenStackRelease(release_2)
+    assert result == expected_result
+    assert result_alternative == expected_result
```
rgildein, line 119: Please do not use something like this, where we define lot of if statements. It would be better to use multiple simple tests or if you really want you can use `eval`, or call directly `__eq__` function.

cou/utils/openstack.py
rgildein: And you did not even test this behavior with any unit tests.

rgildein (changes requested):
One request to fix compassion functions and request to change design of unit tests

cou/utils/openstack.py
gabrielcocenza: If we just use the index to compare it won't be possible to compare a string with the ` OpenStackRelease` object. I think it's nice to be able to do something like:

```python
OpenStackRelease("victoria") > "ussuri" 
```

Using this approach just comparing `OpenStackRelease` will be possible.

cou/steps/analyze.py
```diff
@@ -18,20 +18,44 @@
...
 
 
 @dataclass
 class Analysis:
-    """Analyze result."""
+    """Analyze result.
+
+    :param apps: Applications in the model
+    :type apps:  Iterable[Application]
+    :param apps_to_upgrade: Applications to upgrade in the model
```
aieri, line 43: Why are we determining a list of applications to upgrade here? I thought we had agreed in the spec that it would be simpler to have a predetermined upgrade process to iterate through, and have each Application class be able to handle a "upgrade to version X" call (which may be a noop if the application is already at the target version).

aieri (changes requested):

cou/steps/analyze.py
gabrielcocenza: True, and it will make things simpler indeed.

I've refactored the patch and removed apps to upgrade. We will still need to know the current and next OpenStack version for both Application and Analyze, so great part of the code is still valid.

Thanks

cou/utils/openstack.py
```diff
@@ -287,7 +264,28 @@ def __le__(self, other):
...
+    def os_release(self, value: str):
+        """Setter of OpenStack release codename.
+
+        :param value: OpenStack release codename.
+        :type value: str
+        :raises ValueError: Raise ValueError if codename is unknown.
+        """
+        if value not in self.openstack_codenames:
+            raise ValueError(f"OpenStack '{value}' is not in '{self.openstack_codenames}'")
+        self._os_release = value
```
rgildein, line 288: You need to update index here, since if I do `OpenStackRelease.os_release = "yoga"` is should also update to new index.

cou/apps/app.py
```diff
@@ -67,6 +72,9 @@ class Application:
     os_origin: str = ""
     channel: str = ""
     units: defaultdict[str, dict] = field(default_factory=lambda: defaultdict(dict))
+    current_os_release: str = ""
+    next_os_release: str = ""
```
aieri, line 76: isn't `next_os_release` a bit unnecessary? You could return an `OpenStackRelease` object in `current_os_release`, and the caller could use that to determine the next release

cou/apps/app.py
```diff
@@ -67,6 +72,9 @@ class Application:
...
+    os_versions: set = field(default_factory=set)
```
aieri, line 77: why cache the set of versions? this could easily be grabbed from units

cou/utils/openstack.py
gabrielcocenza: Sure, I'm returning NotImplemented, but I kept the comparison.

cou/utils/openstack.py
gabrielcocenza: Good catch. Thanks

aieri (changes requested):

cou/apps/app.py
gabrielcocenza: Good idea. Now the `OpenStackCodenameLookup` returns a list of `OpenStackRelease` and we can access the next release by `next_release` property

cou/apps/app.py
gabrielcocenza: It was to have this information in the object and be able to test like if we would have multiple OpenStack releases. However this is not necessary and adds some unnecessary lines of code. I've removed `os_versions` from `Application` and `Analysis` object.

cou/utils/openstack.py
```diff
@@ -147,3 +209,117 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+        """Do less than or equals."""
+        return not self.__gt__(other)
+
+    def __repr__(self):
+        """Return the representation of CompareOpenStack."""
+        return f"{self.__class__.__name__}<{self.os_release}>"
+
+    @property
+    def os_release(self) -> str:
+        """Return the next OpenStack release codename.
```
aieri, line 169: This does not return the next release
```suggestion
        """Return the OpenStack release codename.
```

cou/utils/openstack.py
```diff
@@ -147,3 +209,117 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+    def __le__(self, other):
+        """Do less than or equals."""
+        return not self.__gt__(other)
+
+    def __repr__(self):
+        """Return the representation of CompareOpenStack."""
+        return f"{self.__class__.__name__}<{self.os_release}>"
+
+    @property
+    def os_release(self) -> str:
```
aieri, line 168: `OpenStackRelease.os_release` seems quite redundant. How about simply `codename`?
```suggestion
    def codename(self) -> str:
```
You'd then have a more natural UX
```
my_openstack = OpenStackRelease('ussuri')
print(my_openstack.codename) # prints 'ussuri'
```

cou/utils/openstack.py
```diff
@@ -147,3 +209,117 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+
+        :return: OpenStack release codename.
+        :rtype: str
+        """
+        try:
+            os_release = self.openstack_codenames[self.index + 1]
+        except IndexError:
+            logger.warning("There is no OpenStack release after %s", self.os_release)
+            os_release = self.openstack_codenames[self.index]
+        return os_release
```
aieri, line 201: this would return the current release if you hit IndexError:
```
bobcat = OpenStackRelease('bobcat')
print(bobcat.next_release) # prints 'bobcat'
```
I think next_release should be `Null` if we're at the end of the list. So you could simply put your return in the `try`

cou/utils/openstack.py
```diff
@@ -147,3 +209,117 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+
+    @property
+    def release_year(self) -> str:
```
aieri, line 204: in the same spirit as above, the `release_` part is redundant. And `year` is not really correct when release numbers contain months as well.
```suggestion
    def date(self) -> str:
```

cou/utils/openstack.py
```diff
@@ -147,3 +209,117 @@ def lookup(cls, component: str, version: str) -> List[str]:
...
+class OpenStackRelease:
+    """Provides a class that will compare strings from an iterator type object.
+
+    Used to provide > and < comparisons on strings that may not necessarily be
+    alphanumerically ordered.  e.g. OpenStack or Ubuntu releases AFTER the
+    z-wrap.
+    """
+
+    openstack_codenames = list(OPENSTACK_CODENAMES.keys())
+    openstack_release_year = list(OPENSTACK_CODENAMES.values())
```
aieri, line 114: not really years, they include months

aieri (changes requested):

rgildein (dismissed):
LGTM, thanks

cou/utils/openstack.py
gabrielcocenza: I've changed to date

cou/utils/openstack.py
gabrielcocenza: Done

cou/steps/analyze.py
```diff
@@ -72,3 +87,22 @@ def __str__(self) -> str:
...
+        """Shows the current OpenStack release codename.
+
+        :return: OpenStack release codename
+        :rtype: str
+        """
+        os_sequence = sorted(self.os_versions.keys(), key=OpenStackRelease)
+        return os_sequence[0]
+
+    @property
+    def next_cloud_os_release(self) -> str:
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/steps/analyze.py
```diff
@@ -72,3 +87,22 @@ def __str__(self) -> str:
...
         return os.linesep.join([str(app) for app in self.apps])
+
+    @property
+    def current_cloud_os_release(self) -> str:
+        """Shows the current OpenStack release codename.
+
+        :return: OpenStack release codename
+        :rtype: str
+        """
+        os_sequence = sorted(self.os_versions.keys(), key=OpenStackRelease)
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/steps/analyze.py
```diff
@@ -46,11 +61,11 @@ async def create(cls) -> Analysis:
         return Analysis(apps=apps)
 
     @classmethod
-    async def _populate(cls) -> set[Application]:
+    async def _populate(cls) -> List[Application]:
         """Generate the applications model.
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/app.py
```diff
@@ -128,6 +129,27 @@ def __str__(self) -> str:
...
+    def current_os_release(self) -> Optional[OpenStackRelease]:
+        """Current OpenStack Release of the application.
+
+        :return: OpenStackRelease object
+        :rtype: OpenStackRelease
+        """
+        os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
+        if len(os_versions) == 1:
+            return list(os_versions)[0]
+        return None
```
rgildein, line 151: not-blocking: Should we provide some warning logs if there are two different os_versions? Returns `None` make sense in that case, but the logs would be nice.

tests/unit/utils/test_openstack.py
```diff
@@ -76,3 +76,163 @@ def test_generate_lookup(service):
...
+        ("xena", "2021.2", "yoga"),
+        ("bobcat", "2023.2", None),  # None when there is no next release
+    ],
+)
+def test_determine_next_openstack_release(os_release, release_year, next_os_release):
+    release = OpenStackRelease(os_release)
+    if next_os_release is None:
+        assert release.next_release is next_os_release
+    else:
+        assert release.next_release == next_os_release
```
rgildein, line 237: not-blocking: I do not think it's needed to define `is` when next_os_release is `None`.
```suggestion
    assert release.next_release == next_os_release
```

rgildein (dismissed):

cou/apps/app.py
gabrielcocenza: Warning logs added

cou/steps/analyze.py
```diff
@@ -71,3 +81,16 @@ def __str__(self) -> str:
         :rtype: str
         """
         return os.linesep.join([str(app) for app in self.apps])
+
+    @property
+    def current_cloud_os_release(self) -> Optional[OpenStackRelease]:
```
aieri, line 86: at the moment we only have a single type of Application class, and use `current_os_release` being `None` to mean several things: "not an OpenStack app", "an OpenStack app undergoing a per-unit upgrade", "an OpenStack app in a weird state". We then calculate the overall cloud version based only on the apps that have a consistent `current_os_release`. This may be ok in this WIP phase, but would lead to incorrect behavior later on (unless modified): imagine an upgrade that is almost complete, save for one last app running into an issue leading to mismatched versions. You re-run cou and suddenly get offered to start the next upgrade transition.

I think for now we should either raise an exception here or return None whenever there is at least one app with `current_os_release == None`.

Later on, when we have more sophisticated classes (e.g. `NotAnOpenStackApplication`, `NovaCompute`, etc) we can be smarter about whether it's ok or not to ignore an app.

aieri (changes requested):
Sorry, I thought of more things on my second pass... :innocent:

cou/apps/app.py
```diff
@@ -128,6 +129,32 @@ def __str__(self) -> str:
...
+        """
+        os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
+        if len(os_versions) == 1:
+            return list(os_versions)[0]
+        logger.warning(
+            "%s has more than one OpenStack version: %s. Setting current release to None.",
+            self.name,
+            os_versions,
+        )
+        return None
```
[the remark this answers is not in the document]
rgildein: I would suggest to raise an Exception instead of using assert here (if we want to do that)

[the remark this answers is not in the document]
gabrielcocenza: Now I'm raising an Exception with mismatch and returning `None` if it's an empty set (this will happen on subordinate charms).

cou/apps/app.py
```diff
@@ -128,6 +129,32 @@ def __str__(self) -> str:
...
+    @property
+    def current_os_release(self) -> Optional[OpenStackRelease]:
+        """Current OpenStack Release of the application.
+
+        :return: OpenStackRelease object
+        :rtype: OpenStackRelease
+        """
+        os_versions = {unit_values.get("os_version") for unit_values in self.units.values()}
+        if len(os_versions) == 1:
+            return list(os_versions)[0]
```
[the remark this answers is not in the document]

cou/apps/app.py
```diff
@@ -128,6 +129,32 @@ def __str__(self) -> str:
...
+        logger.warning(
+            "%s has more than one OpenStack version: %s. Setting current release to None.",
```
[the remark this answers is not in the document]

cou/steps/analyze.py
```diff
@@ -71,3 +81,16 @@ def __str__(self) -> str:
...
+        """Shows the current OpenStack release codename.
+
+        :return: OpenStack release codename
+        :rtype: OpenStackRelease
+        """
+        os_versions = set()
+        for app in self.apps:
+            if app.current_os_release:
+                os_versions.add(app.current_os_release)
+        return min(os_versions) if os_versions else None
```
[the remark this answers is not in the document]
gabrielcocenza: comments and logs added

cou/steps/analyze.py
gabrielcocenza: Now we are like this:
- `Non - OpenStack app` is skipped because there is no source or openstack-origin and a debug log was addded
- OpenStack app in a weird state raises exception. This is currently including the situation of upgrading per-unit
- skipping apps with `current_os_release == None`. This is currently happening on subordinate charms (until we implement a proper class) because there are no units on ApplicationStatus and also when we are not able to identify the OpenStack release because the workload version is unknown. On both cases I'm using warning logs.

cou/apps/app.py
```diff
@@ -128,6 +135,42 @@ def __str__(self) -> str:
...
+            # subordinate applications class.
+            return None
+        if len(os_versions) == 1:
+            return os_versions.pop()
+        # NOTE (gabrielcocenza) on applications that use single-unit or paused-single-unit
+        # upgrade methods, more than one version can be found.
+        logger.error(
+            (
+                "Units of application %s are running mismatched OpenStack versions: %s. "
+                "Setting current release to None."
```
aieri, line 167:
```suggestion
                "This is not currently handled."
```

cou/steps/analyze.py
```diff
@@ -71,3 +81,35 @@ def __str__(self) -> str:
...
+        for app in self.apps:
+            if app.os_origin:
+                # NOTE (gabrielcocenza) this is temporarily skipping subordinate charms
+                # until there is a specific class able to get the OpenStack release.
+                # It's also skipping applications with no identified OpenStack release.
+                if app.current_os_release:
+                    os_versions.add(app.current_os_release)
+                else:
+                    logger.warning(
+                        "Disconsidering %s to determine the minimum version of the cloud.",
```
aieri, line 104:
```suggestion
                        "Ignoring %s when determining the minimum version of the cloud.",
```

cou/steps/analyze.py
```diff
@@ -71,3 +81,35 @@ def __str__(self) -> str:
...
+                        app.name,
+                    )
+            else:
+                logger.debug(
+                    (
+                        "Disconsidering %s to determine the min version of the cloud "
```
aieri, line 110:
```suggestion
                        "Ignoring %s when determining the minimum version of the cloud: "
```

cou/steps/analyze.py
```diff
@@ -71,3 +81,35 @@ def __str__(self) -> str:
...
+                        "because isn't an OpenStack charm."
```
aieri, line 111:
```suggestion
                        "not an OpenStack charm."
```

tests/unit/steps/test_steps_analyze.py
```diff
@@ -55,31 +55,73 @@ async def test_analysis_dump(mocker, apps):
...
+    assert result.current_cloud_os_release == "ussuri"
+    # no OpenStack charms doesn't show on logs.
+    mock_logger.warning.assert_called_once_with(
+        "Disconsidering %s to determine the minimum version of the cloud.", app_mysql_router.name
+    )
+    mock_logger.debug.assert_called_once_with(
+        (
+            "Disconsidering %s to determine the min version of the cloud "
+            "because isn't an OpenStack charm."
+        ),
```
aieri, line 124: these will need to be updated with the changes suggested above

tests/unit/steps/test_steps_analyze.py
```diff
@@ -55,31 +55,73 @@ async def test_analysis_dump(mocker, apps):
...
+        no_openstack.name,
+    )
+    assert result.current_cloud_os_release.next_release == "victoria"
```
aieri, line 127: Isn't this out of scope for this test? Having `.next_release` return the right value should be tested (and is tested) in `test_openstack.py`.
If a regression occurs in `utils/openstack.py` this test may fail and confuse the developer.

tests/unit/steps/test_steps_analyze.py
```diff
@@ -55,31 +55,73 @@ async def test_analysis_dump(mocker, apps):
...
     apps = await Analysis._populate()
-    assert len(apps) == 3
-    assert {app.charm for app in apps} == {"keystone", "cinder", "rabbitmq-server"}
+    assert len(apps) == 4
+    # apps are on the UPGRADE_ORDER sequence and unknown in the end of the list
+    assert [app.charm for app in apps] == ["rabbitmq-server", "keystone", "cinder", "my-app"]
 
 
 @pytest.mark.asyncio
 async def test_analysis(mocker, apps):
```
aieri, line 96: This test packs a lot of logic! Ideally, tests should be as self-evident as possible, contain the least amount of logic possible, only contain one assertion per test (ideally), and be unashamedly verbose if needed. This is so that when something fails, you can more quickly understand what did.

iiuc you are testing that:
* the Analysis object is built correctly
* that the current_cloud_os_release property reacts correctly to three separate cases:
  * subordinates are skipped
  * non-OpenStack charms are skipped
  * OpenStack charms on the same version yield the correct result

I suggest splitting this test in at least three tests (or one with parametrization), focusing on `current_cloud_os_release`, and adding a case including OpenStack charms of different versions (to test the min algorithm).

aieri (changes requested):
We're really close now! I only have a few wording and test optimization suggestions.

tests/unit/steps/test_steps_analyze.py
gabrielcocenza: Done

tests/unit/steps/test_steps_analyze.py
gabrielcocenza: Good catch. Thanks!
I've splitted the unit tests to separate what we are testing.

cou/apps/app.py
```diff
@@ -128,6 +135,42 @@ def __str__(self) -> str:
...
+        """
+        return self.status.series
+
+    @property
+    def current_os_release(self) -> Optional[OpenStackRelease]:
+        """Current OpenStack Release of the application.
+
+        :return: OpenStackRelease object
+        :rtype: OpenStackRelease
+        """
```
agileshaw, line 153: Add `raise` in docstring
```suggestion
        :raises MismatchedOpenStackVersions: Raise MismatchedOpenStackVersions if units of an application are running mismatched OpenStack versions.
        """
```

tests/unit/steps/test_steps_analyze.py
```diff
@@ -55,31 +55,114 @@ async def test_analysis_dump(mocker, apps):
...
-async def test_generate_model(mocker, full_status, config):
+async def test_populate_model(mocker, full_status, config):
     mocker.patch.object(analyze, "async_get_status", return_value=full_status)
     mocker.patch.object(
         analyze, "async_get_application_config", return_value=config["openstack_ussuri"]
     )
-    # Initially, 3 applications are in the status (keystone, cinder and rabbitmq-server)
-    assert len(full_status.applications) == 3
+    # Initially, 4 applications are in the status: keystone, cinder, rabbitmq-server and my-app
+    # my-app is a unknown application in the model.
```
agileshaw, line 87:
```suggestion
    # my-app is an unknown application in the model.
```

agileshaw (changes requested):
Functionality-wise LGTM. I added some suggestions in-line for best practices and rewording

aieri (commented):
LGTM, I think we're ready to merge once Tianqi's suggestions are applied.

cou/utils/openstack.py
```diff
@@ -38,6 +38,189 @@
...
+    "designate-bind",
+    "glance",
+    "gnocchi",
+    "heat",
+    "manila",
+    "manila-ganesha",
+    "neutron-api",
+    "neutron-gateway",
+    "ovn-central",
+    "ovn-dedicated-chassis",
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/utils/openstack.py
```diff
@@ -38,6 +38,189 @@
...
+        ("train", "2019.2"),
+        ("ussuri", "2020.1"),
+        ("victoria", "2020.2"),
+        ("wallaby", "2021.1"),
+        ("xena", "2021.2"),
+        ("yoga", "2022.1"),
+        ("zed", "2022.2"),
+        ("antelope", "2023.1"),
+        ("bobcat", "2023.2"),
+    ]
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/utils/openstack.py
```diff
@@ -38,6 +38,189 @@
...
+)
+
+
+class OpenStackRelease:
+    """Provides a class that will compare strings from an iterator type object.
+
+    Used to provide > and < comparisons on strings that may not necessarily be
+    alphanumerically ordered.  e.g. OpenStack or Ubuntu releases AFTER the
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/app.py
gabrielcocenza: Done

tests/unit/steps/test_steps_analyze.py
gabrielcocenza: Done

cou/apps/app.py
agileshaw: I don't think we should remove the `return` though. Both `return` and `raise` should be included in this docstring

gabrielcocenza: Definitely. I've removed without intention. Thanks

agileshaw (approved):
LGTM :+1:

aieri (approved):

PR #26, opened 2023-08-04
Added extended model.wait_for_idle that handles disconnections.

esunar:


gabrielcocenza (approved):
LGTM. Just left some points that are no-blockers

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,98 @@ async def async_upgrade_charm(
...
+# pylint: disable=too-few-public-methods
+class JujuWaiter:
+    """Enhanced version of wait_for_idle.
+
+    Usage:
+        JujuWaiter(model).wait(120)
+    """
+
+    DEFAULT_TIMEOUT: int = 3600
+    MODEL_IDLE_PERIOD: int = 15
```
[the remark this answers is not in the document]
esunar: MODEL_IDLE_PERIOD increased to 30 seconds

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
         revision=revision,
         switch=switch,
     )
+
+
+# pylint: disable=too-few-public-methods
```
rgildein, line 402: I think that this should not be needed, since this class contains some methods.

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+        if timeout_seconds:
+            self.timeout = timedelta(seconds=timeout_seconds)
+        self.log.debug("Waiting to stabilize in %s seconds", self.timeout)
+
+        while True:
+            await self._ensure_model_connected()
+            try:
+                await self.model.wait_for_idle(
+                    idle_period=JujuWaiter.MODEL_IDLE_PERIOD,
+                    timeout=JujuWaiter.JUJU_IDLE_CHECK_PERIOD,
```
rgildein, line 451: Is this correct? Maybe this
```python
check_freq=JujuWaiter.JUJU_IDLE_CHECK_PERIOD, timeout=self.timeout
```

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+                    JujuWaiter.MODEL_IDLE_PERIOD,
+                )
+                return
+            except (
+                JujuMachineError,
+                JujuAgentError,
+                JujuUnitError,
+                JujuAppError,
+            ) as ex:
+                raise ex
```
rgildein, line 465: not-blocking:
```suggestion
                raise
```

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+            except JujuWaiter.TimeoutException as ex:
+                self.log.debug(
+                    "Unable to stabilize in %s seconds",
+                    datetime.now() - self.start_time,
+                )
+                raise ex
```
rgildein, line 471: not-blocking:
```suggestion
                raise
```

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+
+    async def _ensure_model_connected(self):
+        """Ensure that the model is connected."""
+        while _is_model_disconnected(self.model):
+            await _disconnect(self.model)
+            try:
+                self._check_time()
+                await self.model.connect_model(self.model_name)
+            except JujuWaiter.TimeoutException as ex:
+                raise ex
```
rgildein, line 488: not-blocking:
```suggestion
                raise
```

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+        :param model: model to wait
+        :type model: Model
+        """
+        self.model = model
+        self.model_name = self.model.info.name
+        self.timeout = timedelta(seconds=JujuWaiter.DEFAULT_TIMEOUT)
+        self.start_time = datetime.now()
+        self.log = logging.getLogger(self.__class__.__name__)
+
+    async def wait(self, timeout_seconds: Optional[int] = None) -> None:
```
rgildein, line 429: When we will need to use waiting without timeout? I think this is not good idea to wait forever.

tests/unit/utils/test_juju_utils.py
```diff
@@ -577,3 +578,83 @@ async def test_disconnect(self):
...
+
+    def raise_exception_sync(*args):
+        raise Exception()
+
+    async def test_normal(self):
+        waiter = model.JujuWaiter(self.model_connected)
+        await waiter.wait(10)
+
+    async def test_exception(self):
+        self.model_juju_exception.wait_for_idle = JujuWaiterTests.raise_juju_unit_exception
```
rgildein, line 626: Why we need this and not `JujuUnitError()` as side_effect?
```suggestion
        self.model_juju_exception.wait_for_idle.side_effect = JujuUnitError()
```

tests/unit/utils/test_juju_utils.py
```diff
@@ -577,3 +578,83 @@ async def test_disconnect(self):
...
+        raise model.JujuWaiter.TimeoutException()
+
+    async def raise_exception(**kwargs):
+        raise Exception()
+
+    def raise_timeout_exception_sync(*args):
+        raise model.JujuWaiter.TimeoutException()
+
+    def raise_exception_sync(*args):
+        raise Exception()
```
rgildein, line 619: I believe that these should be replaced with side_effect

cou/utils/juju_utils.py
rgildein, line 471: When this could happen? Libjuju will never raise your own exception.

rgildein (changes requested):

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+        while _is_model_disconnected(self.model):
+            await _disconnect(self.model)
+            try:
+                self._check_time()
+                await self.model.connect_model(self.model_name)
+            except JujuWaiter.TimeoutException as ex:
+                raise ex
+            except Exception as ex:
+                self.log.debug(
+                    "Model has unexpected exception while connecting, retrying", exc_info=ex
```
rgildein, line 491: I'm not sure if this will work as expected, since you provide exc_info, but never used that in massage or fromatter?

cou/utils/juju_utils.py
gabrielcocenza: True. This should be under `_check_time()`

cou/utils/juju_utils.py
esunar: It is correct we want it to be count as idle if it is idle more than `MODEL_IDLE_PERIOD` and we want it to timeout after `JUJU_IDLE_CHECK_PERIOD`

rgildein: Can we provided some description? Just to describe why we are using check period for timeout. Thanks

cou/utils/juju_utils.py
rgildein: I do not see purpose of this, since we have `_check_time` function, which is providing some logs and can provide this also, instead of catching it and raising it again.

cou/utils/juju_utils.py
esunar: It is initialized in constructor with some sane default
`self.timeout = timedelta(seconds=JujuWaiter.DEFAULT_TIMEOUT)`

tests/unit/utils/test_juju_utils.py
esunar: ok

cou/utils/juju_utils.py
esunar: it is raised by `_check_time`

cou/utils/juju_utils.py
esunar: ok

cou/utils/juju_utils.py
esunar: ok

cou/utils/juju_utils.py
rgildein: This is not good at all, because we are hiding that there is default value. It should be `timeout: int = DEFAULT_TIMEOUT`

cou/utils/juju_utils.py
rgildein: But `_check_time ` is outside of try. And as I mention there is not reason to catching it here a provide log like this. This log message belong to the function itself.

cou/utils/juju_utils.py
esunar: I've replaced with `exc_info=True`. [reference](https://docs.python.org/3.10/library/logging.html#logging.Logger.debug)

esunar:
> I think we should something like this example, since it will be more readable and easy to use.
> 
> ```python
> import logging
> 
> from juju.model import Model
> 
> DEFAULT_TIMEOUT: int = 3600
> MODEL_IDLE_PERIOD: int = 30
> JUJU_IDLE_CHECK_PERIOD: int = 40
> 
> logger = logging.getLogger(__name__)
> 
> class TimeoutException(Exception):
>     """Own timeout exception."""
> 
> class ModelConnectionError(Exception):
>     """Model not connected."""
> 
> async def _ensure_model_connected(self, mode: Model):
>     """Ensure that the model is connected."""
>     name = model.info.name
>     while _is_model_disconnected(model):
>         await _disconnect(model)
>         try:
>             await model.connect_model()
>         except Exception as ex:
>             logger.debug(
>                 "Model has unexpected exception while connecting, retrying %s", ex
>             )
>             raise ModelConnectionError from ex
> 
> def wait_for_idle_model(
>     model: Model, 
>     timeout: int = DEFAULT_TIMEOUT, 
>     idle_period: int = MODEL_IDLE_PERIOD,
>     check_period: int = JUJU_IDLE_CHECK_PERIOD,
> ) -> None:
>     """Wait for model to be in indle state."""
>     start_time = datetime.now()
>     timeout = timedelta(seconds=timeout)
>     logger.debug("Waiting to stabilize in %s seconds", timeout)
> 
>     while datetime.now() - start_time > timeout::
>         try:
>             await _ensure_model_connected(model)
>             await model.wait_for_idle(idle_period=idle_period, timeout=check_period)
>             logger.debug(
>                 "Model %s is idle for %s seconds.",
>                 self.model.info.name,
>                 idle_period,
>             )
>             return
>         except (
>             JujuMachineError,
>             JujuAgentError,
>             JujuUnitError,
>             JujuAppError,
>             ModelConnectionError,
>         ):
>             continue
>         except Exception as ex:
>             # We do not care exceptions other than Juju(Machine|Agent|Unit|App)Error because
>             # when juju connection is dropped you can have wide range of exceptions depending
>             # on the case
>             logger.debug("Unknown error while waiting to stabilize %s",ex)
> ```

This is for future encapsulation of juju connection class. It will be extended to incude all juju related functionality. Since it will create huge change I decided to start small.

rgildein (changes requested):

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+                    datetime.now() - self.start_time,
+                )
+                raise
+            except Exception as ex:
+                # We do not care exceptions other than Juju(Machine|Agent|Unit|App)Error because
+                # when juju connection is dropped you can have wide range of exceptions depending
+                # on the case
+                self.log.debug("Unknown error while waiting to stabilize", exc_info=ex)
+
+            self._check_time()
```
gabrielcocenza, line 478: Try except `JujuWaiter.TimeoutException` should catch this function

gabrielcocenza (changes requested):

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,101 @@ async def async_upgrade_charm(
...
+                JujuUnitError,
+                JujuAppError,
+            ):
+                raise
+            except JujuWaiter.TimeoutException:
+                self.log.debug(
+                    "Unable to stabilize in %s seconds",
+                    datetime.now() - self.start_time,
+                )
+                raise
```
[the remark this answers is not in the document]
esunar: removed.

cou/utils/juju_utils.py
esunar: No need to catch since it is re-thrown as is.

cou/utils/juju_utils.py
esunar: removed.

cou/utils/juju_utils.py
esunar: changed as suggested.

cou/utils/juju_utils.py
esunar: Added extra comments for the constants.

cou/utils/juju_utils.py
```diff
@@ -396,3 +397,107 @@ async def async_upgrade_charm(
...
+        while _is_model_disconnected(self.model):
+            await _disconnect(self.model)
+            try:
+                self._check_time()
+                await self.model.connect_model(self.model_name)
+            except JujuWaiter.TimeoutException:
+                raise
+            except Exception:
+                self.log.debug(
+                    "Model has unexpected exception while connecting, retrying", exc_info=True
```
agileshaw, line 493: I was wondering why we are suing `exc_info=True` here but at line 475 we use `exc_info=ex`? If I'm not mistaken, they have the same effect: attaching the exception instance to the log message. If that's true, could we unify the code for easier future maintenance?

agileshaw (changes requested):
Overall LGTM. Just some small suggestions inline.

cou/utils/juju_utils.py
esunar: Correct version is `exc_info=True` [reference](https://docs.python.org/3.10/library/logging.html#logging.Logger.debug)
corrected

gabrielcocenza (approved):
LGTM

rgildein (approved):
To me this is over engineered solution, but I do not want to block it.

PR #27, opened 2023-08-08
Separate the execution logic

esunar:
Extract the execution logic to a simple interface
Created a simple ExecutorFactory
Created a simple SerialExecutor by using the cli.apply_plan 
Used SerialExecutor in cli

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -75,7 +75,10 @@ def parse_args(args: Any) -> argparse.Namespace:
         "  3 - Current active juju model",
     )
     parser.add_argument(
-        "--non-interactive", help="Run upgrade without prompt.", action="store_true", default=False
+        "--interactive",
+        help="Run upgrade with prompt.",
+        action=argparse.BooleanOptionalAction,
+        default=True,
     )
     parser.add_argument(
         "-h",
@@ -136,7 +139,7 @@ async def entrypoint() -> None:
         print(analysis_result)
         upgrade_plan = await generate_plan(analysis_result)
         if args.run:
-            await execute(upgrade_plan, not args.non_interactive)
+            await execute(upgrade_plan, args.interactive)
         else:
             print(upgrade_plan)
 
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -22,10 +22,8 @@ import sys
 from datetime import datetime
 from typing import Any
 
-from colorama import Fore, Style
-
-from cou.steps import UpgradeStep
 from cou.steps.analyze import Analysis
+from cou.steps.execute import ExecutorFactory
 from cou.steps.plan import generate_plan
 from cou.utils import juju_utils as utils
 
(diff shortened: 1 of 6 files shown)


rgildein (commented):
Why we are creating the BaseExecutor, ExecutorFactory and SerialExecutor? Why creating class object if it's not necessary? We create just more friction, and even pylint is telling us not to do so why even use pylint if we doing disable.

esunar:
> Why we are creating the BaseExecutor, ExecutorFactory and SerialExecutor? Why creating class object if it's not necessary? We create just more friction, and even pylint is telling us not to do so why even use pylint if we doing disable.

At the time being, it seems over engineering. However it is a guideline to show how will the application evolve. For example in near future we will be using parallel execution. When we move to data plane we will be using both. When it get complicated it will be really hard to maintain in a module with bunch of functions.

cou/steps/execute.py
```diff
@@ -0,0 +1,90 @@
...
+
+async def _apply_plan(plan: UpgradeStep, interactive: bool) -> None:
+    """Apply the plan for upgrade.
+
+    :param plan: Plan to be executed on steps.
+    :type plan: UpgradeStep
+    :param interactive:
+    :type interactive: bool
+    """
+    result = "X"
```
aieri, line 64: nit: `X` looks at a first glance as a possible option. We could use an empty string here to represent 'no result'

setup.cfg
```diff
@@ -6,19 +6,19 @@ description = Application for upgrading Charmed Openstack
 long_description = file: README.md
 long_description_content_type = text/markdown; charset=UTF-8; variant=GFM
 author = Canonical DevOps Centres
-license = GPLv3
+license = Apache-2.0
```
aieri, line 9: A change in licence seems out of scope for this PR. Also: why?

aieri: the headers of other files in this PR are also still mentioning the GPL

aieri (changes requested):

setup.cfg
esunar: I'll remove that part and open another pr for license.

cou/cli.py
```diff
@@ -188,7 +136,7 @@ async def entrypoint() -> None:
         print(analysis_result)
         upgrade_plan = await generate_plan(analysis_result)
         if args.run:
-            await apply_plan(upgrade_plan, args.non_interactive)
+            await execute(upgrade_plan, not args.non_interactive)
```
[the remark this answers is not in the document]
esunar: used the second approach

cou/steps/execute.py
esunar: removed `X`

setup.cfg
esunar: removed

cou/steps/execute.py
```diff
@@ -61,8 +60,8 @@ async def _apply_plan(plan: UpgradeStep, interactive: bool) -> None:
     :param interactive:
     :type interactive: bool
     """
-    result = "X"
-    while result.casefold() not in AVAILABLE_OPTIONS:
+    result = ""
+    while result == "" or result.casefold() not in AVAILABLE_OPTIONS:
```
aieri, line 64: ah yes, because an empty string is considered to always be part of a string. You could (non blocking) define `AVAILABLE_OPTIONS` as a list and get rid of the `result == ""` comparison.

aieri (dismissed):
LGTM w/one optional suggestion

cou/steps/execute.py
esunar: done

PR #30, opened 2023-08-15
Add MySQL charms on OpenStack lookup

gabrielcocenza:
- the OpenStack lookup is ready for MySQL charms, but it was missing the specific charms.


aieri (approved):
lgtm

PR #33, opened 2023-08-25
Pre-upgrade package refresh

agileshaw:
A pre-upgrade step that ensures all packages are at their current latest version. The step is at the moment implemented at the Application level as we are currently only dealing with control plane components.

In addition to the commands recommended in the charm upgrade documentation, we also included the following changes:
1. add dpkg options as used in the latest charm upgrade functions (e.g. [charm-nova-compute](https://opendev.org/openstack/charm-nova-compute/src/commit/f6892d228c6e6eff72bbfc700a4d5a4a9a4d92e7/hooks/nova_compute_utils.py#L769))
2. autoremvoe old, dangling dependency packages.

This PR should be merged after https://github.com/canonical/charmed-openstack-upgrader/pull/21

### DIFF
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -430,7 +430,7 @@ class OpenStackApplication:
         return UpgradeStep(
             description=(
                 f"Upgrade software packages of '{self.name}' to the latest "
-                f"in '{self.current_os_release}' release"
+                f"'{self.current_os_release}' release"
             ),
             parallel=parallel,
             function=upgrade_packages,
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -153,7 +153,7 @@ class OpenStackApplication:
     :raises MismatchedOpenStackVersions: When units part of this application are running mismatched
         OpenStack versions.
     :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation
-    :raises PackageUpgradeError: When the upgrade command execution resulted in error.
+    :raises PackageUpgradeError: When the package upgrade fails.
     """
 
     # pylint: disable=too-many-instance-attributes
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -29,10 +29,12 @@ from cou.exceptions import (
     ApplicationError,
     HaltUpgradePlanGeneration,
     MismatchedOpenStackVersions,
+    PackageUpgradeError,
 )
 from cou.steps import UpgradeStep
 from cou.utils.juju_utils import (
     async_get_status,
+    async_run_on_unit,
     async_set_application_config,
     async_upgrade_charm,
 )
@@ -152,6 +154,7 @@ class OpenStackApplication:
     :raises MismatchedOpenStackVersions: When units part of this application are running mismatched
         OpenStack versions.
     :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation
+    :raises PackageUpgradeError: When the upgrade command execution resulted in error.
     """
 
     # pylint: disable=too-many-instance-attributes
@@ -342,6 +345,44 @@ class OpenStackApplication:
             )
             raise ApplicationError()
 
+    async def _run_package_upgrade_command(self, unit: str, command: str) -> None:
+        """Run package upgrade command and raise Exception if fails.
+
+        :param unit: The name of the unit where the command runs on.
+        :type str
+        :param command: The upgrade command to run.
+        :type str
+        :raises PackageUpgradeError: When the upgrade command execution resulted in error.
+        """
+        logger.info("Running '%s' on %s", command, unit)
+        result = await async_run_on_unit(
+            unit_name=unit, command=command, model_name=self.model_name
+        )
+        if str(result["Code"]) == "0":
+            logger.debug(result["Stdout"])
+        else:
+            logger.error(
+                "Error upgrading package on %s: %s",
+                unit,
+                result["Stderr"],
+            )
+            raise PackageUpgradeError()
+
+    async def _upgrade_packages(self) -> None:
+        """Run package updates and upgrades on each unit of an Application."""
+        status = await async_get_status()
+        app_status = status.applications.get(self.name)
+        for unit in app_status.units.keys():
+            update_command = "sudo apt update"
+            await self._run_package_upgrade_command(unit=unit, command=update_command)
+
+            dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
+            upgrade_command = f"sudo apt full-upgrade {dpkg_opts} -y"
+            await self._run_package_upgrade_command(unit=unit, command=upgrade_command)
+
+            autoremove_command = "sudo apt autoremove"
+            await self._run_package_upgrade_command(unit=unit, command=autoremove_command)
+
     def pre_upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
         """Pre Upgrade planning.
 
@@ -350,7 +391,10 @@ class OpenStackApplication:
         :return: Plan that will add pre upgrade as sub steps.
         :rtype: list[Optional[UpgradeStep]]
         """
-        return [self._get_refresh_charm_plan(target)]
+        return [
+            self._get_upgrade_current_release_packages_plan(),
+            self._get_refresh_charm_plan(target),
+        ]
 
     def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
         """Upgrade planning.
@@ -413,6 +457,23 @@ class OpenStackApplication:
             if step:
                 upgrade_steps.add_step(step)
         return upgrade_steps
+
+    def _get_upgrade_current_release_packages_plan(self, parallel: bool = False) -> UpgradeStep:
+        """Get Plan for upgrading software packages to the latest of the current release.
+
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return plan: Plan for upgrading software packages to the latest of the current release.
+        :type plan: UpgradeStep
+        """
+        return UpgradeStep(
+            description=(
+                f"Upgrade software packages of '{self.name}' to the latest "
+                f"in '{self.current_os_release}' release"
+            ),
+            parallel=parallel,
+            function=self._upgrade_packages,
+        )
 
     def _get_refresh_charm_plan(
         self, target: OpenStackRelease, parallel: bool = False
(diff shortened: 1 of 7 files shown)


cou/apps/app.py
```diff
@@ -338,6 +341,44 @@ async def _check_upgrade(self, target: OpenStackRelease) -> None:
...
+        app_status = status.applications.get(self.name)
+        for unit in app_status.units.keys():
+            update_command = "sudo apt update"
+            await self._run_package_upgrade_command(unit=unit, command=update_command)
+
+            dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
+            upgrade_command = f"sudo apt full-upgrade {dpkg_opts} -y"
+            await self._run_package_upgrade_command(unit=unit, command=upgrade_command)
+
+            autoremove_command = "sudo apt autoremove"
```
aieri, line 379: `apt` is generally geared toward interactive use, not scripting, so using `apt-get` would be more appropriate here. From the man:
```
SCRIPT USAGE AND DIFFERENCES FROM OTHER APT TOOLS
       The apt(8) commandline is designed as an end-user tool and it may change behavior between versions. While it tries not to break backward compatibility this is not guaranteed either if a change seems
       beneficial for interactive use.

       All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5) and specifically the
       Binary scope). So you should prefer using these commands (potentially with some additional options enabled) in your scripts as they keep backward compatibility as much as possible.
```

Also please add `-y` flags where needed otherwise the commands would never return.

cou/apps/app.py
```diff
@@ -338,6 +341,44 @@ async def _check_upgrade(self, target: OpenStackRelease) -> None:
             )
             raise ApplicationError()
 
+    async def _run_package_upgrade_command(self, unit: str, command: str) -> None:
```
esunar, line 344: `_run_package_upgrade_command` and `_upgrade_packages` should not be in this class. Because this class is `OpenStackApplication` and these functions does not apply just for `OpenStackApplication`. I think we should just move them out of class.

esunar (changes requested):
We should move upgrading functions out of this class and use them in this class.

cou/apps/app.py
agileshaw: Moved the function to the newly created `app_utils.py`

cou/apps/app.py
agileshaw: Changed `apt` to `apt-get`. 

The previous patch already has `-y` flag for upgrade command, but I realized that autoremove needs it as well. Added.

cou/utils/app_utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def upgrade_packages(units: Iterable[str], model_name: str) -> None:
+    """Run package updates and upgrades on each unit of an Application.
+
+    :param units: The list of unit names where the package upgrade runs on.
+    :type Iterable[str]
+    :param model_name: The name of the model that the applcation belongs to.
```
aieri, line 30:
```suggestion
    :param model_name: The name of the model that the application belongs to.
```

cou/utils/app_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    :type str
+    :raises PackageUpgradeError: When the upgrade command execution resulted in error.
+    """
+    for unit in units:
+        dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
+        command = (
+            "sudo apt-get update && "
+            f"sudo apt-get dist-upgrade {dpkg_opts} -y && "
+            "sudo apt-get autoremove -y"
+        )
```
aieri, line 40: nit: these could remain outside of the for loop to avoid being redefined multiple times

cou/utils/app_utils.py
```diff
@@ -0,0 +1,52 @@
...
+    :param model_name: The name of the model that the applcation belongs to.
+    :type str
+    :raises PackageUpgradeError: When the upgrade command execution resulted in error.
+    """
+    for unit in units:
+        dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
+        command = (
+            "sudo apt-get update && "
+            f"sudo apt-get dist-upgrade {dpkg_opts} -y && "
+            "sudo apt-get autoremove -y"
```
aieri, line 39: `juju run -u <unit>` runs commands as root. Is this not the case for `async_run_on_unit`?

cou/exceptions.py
```diff
@@ -56,6 +56,10 @@ class ApplicationError(Exception):
     """Exception when Application does something unexpected."""
 
 
+class PackageUpgradeError(Exception):
+    """Exception when pacakge upgrade fails."""
```
aieri, line 60:
```suggestion
    """Exception raised when a package upgrade fails."""
```

cou/apps/app.py
```diff
@@ -410,6 +415,25 @@ def generate_upgrade_plan(self, target: str) -> UpgradeStep:
...
+
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return plan: Plan for upgrading software packages to the latest of the current release.
+        :type plan: UpgradeStep
+        """
+        return UpgradeStep(
+            description=(
+                f"Upgrade software packages of '{self.name}' to the latest "
+                f"in '{self.current_os_release}' release"
```
aieri, line 429:
```suggestion
                f"Upgrade software packages of '{self.name}' to the latest "
                f"'{self.current_os_release}' release"
```

aieri (changes requested):
Overall LGTM, just a few small comments

cou/utils/app_utils.py
agileshaw: It is running as the root user indeed. I included `sudo` becuase we were initially considering package refreshes at machine level (`juju run --machine <machine>`  runs command as `ubuntu` user) and I forgot to make the change after switching to running commands against units. It is now removed.

cou/exceptions.py
```diff
@@ -33,27 +33,31 @@ def __init__(self, cmd: str, code: str, output: str, err: str):
...
     """Exception to halt the application upgrade at any moment."""
 
 
 class ApplicationError(Exception):
-    """Exception when Application does something unexpected."""
+    """Exception raised when Application does something unexpected."""
+
+
+class PackageUpgradeError(Exception):
+    """Exception raised when a pacakge upgrade fails."""
```
aieri, line 60:
```suggestion
    """Exception raised when a package upgrade fails."""
```

cou/exceptions.py
```diff
@@ -33,27 +33,31 @@ def __init__(self, cmd: str, code: str, output: str, err: str):
 
 
 class UnitNotFound(Exception):
-    """Unit not found in actual dict."""
+    """Exception raised when unit not found in the actual dict."""
```
aieri, line 36: Actually now that you've modified this docstring I realize it's kinda confusing... which "actual dict"? You're also missing an article:
```suggestion
    """Exception raised when a unit is not found in the actual dict."""
```

cou/exceptions.py
```diff
@@ -33,27 +33,31 @@ def __init__(self, cmd: str, code: str, output: str, err: str):
...
 
 
 class JujuError(Exception):
-    """Exception when libjuju does something unexpected."""
+    """Exception raised when libjuju does something unexpected."""
 
 
 class MismatchedOpenStackVersions(Exception):
-    """Exception when more than one OpenStack version are found in the Application."""
+    """Exception raised when more than one OpenStack version are found in the Application."""
```
aieri, line 44:
```suggestion
    """Exception raised when more than one OpenStack version is found in the Application."""
```

aieri (changes requested):
Looks good, we're down to just a few typos now.

esunar (approved):

cou/exceptions.py
agileshaw: Reworded

cou/utils/app_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    :param model_name: The name of the model that the application belongs to.
+    :type str
+    :raises PackageUpgradeError: When the upgrade command execution resulted in error.
+    """
+    dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
+    command = f"apt-get update && apt-get dist-upgrade {dpkg_opts} -y && apt-get autoremove -y"
+
+    for unit in units:
+        logger.info("Running '%s' on '%s'", command, unit)
+        result = await async_run_on_unit(unit_name=unit, command=command, model_name=model_name)
```
gabrielcocenza, line 39: Wouldn't be good to use the timeout and maybe a try except for `JujuError` ?

gabrielcocenza (commented):
I've left two comments, but generally LGTM

gabrielcocenza (dismissed):
Thanks. LGTM

esunar (approved):

cou/utils/app_utils.py
```diff
@@ -0,0 +1,53 @@
...
+            result = await async_run_on_unit(
+                unit_name=unit, command=command, model_name=model_name, timeout=600
+            )
+            if str(result["Code"]) == "0":
+                logger.debug(result["Stdout"])
+            else:
+                logger.error("Error upgrading package on %s: %s", unit, result["Stderr"])
+                raise PackageUpgradeError()
+        except JujuError as exc:
+            logger.error("Failed running package upgrade on %s: %s", unit, exc)
```
aieri, line 52:
```suggestion
            logger.error("Failed running package upgrades on %s: %s", unit, exc)
```

aieri (changes requested):
Just two tiny typos left

aieri (approved):

cou/utils/app_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+        try:
+            result = await async_run_on_unit(
+                unit_name=unit, command=command, model_name=model_name, timeout=600
+            )
+            if str(result["Code"]) == "0":
+                logger.debug(result["Stdout"])
+            else:
+                logger.error("Error upgrading packages on %s: %s", unit, result["Stderr"])
+                raise PackageUpgradeError()
```
rgildein, line 50: Why we are raising error without any message?

agileshaw: The point of raising the exception is to stop the upgrade process (and give the calling code a chance to do something about it). We could add a message, I just didn't think it was necessary since we are logging the error.

rgildein: I think it is always nice to do so and maybe even leave logging to other catcher.

agileshaw: To close the loop here, after discussion we agree to rework the exception and log handling and open a separate PR for it.

PR #34, opened 2023-08-28
Limit pr workflow concurrency.

esunar:


gabrielcocenza (approved):
LGTM

agileshaw (commented):
I'm interested in the reason behind the [functional test error](https://github.com/canonical/charmed-openstack-upgrader/actions/runs/6004861003/job/16286377148) when the workflow is running more than once same time because I don't remember seeing it in other repos. If this reason is commonly affected, we should update the template in charm spec

esunar:
> I'm interested in the reason behind the [functional test error](https://github.com/canonical/charmed-openstack-upgrader/actions/runs/6004861003/job/16286377148) when the workflow is running more than once same time because I don't remember seeing it in other repos. If this reason is commonly affected, we should update the template in charm spec

We have one runner running these tests. If it is available it runs the test but if not it chooses a random runner provided by IS team. An those ones do not contain snapcraft.

agileshaw (approved):
After further discussion w/ @esunar, the functional test in this repo needs to use a specific self-hosted runner. Limiting concurrency prevents it from running elsewhere

PR #38, opened 2023-08-29
Subordinate charms handling.

esunar:
It is ready for initial review. 
All the subordinate charms that have openstack tracks are added.
Added a general purpose is_charm_supported function


cou/apps/subordinate.py
```diff
@@ -0,0 +1,74 @@
...
+        :rtype: UpgradeStep
+        """
+        plan = UpgradeStep(description=f"Upgrade {self.name}", parallel=False, function=None)
+        upgrade_charm_plan = self._get_upgrade_charm_plan(OpenStackRelease(target))
+        if upgrade_charm_plan:
+            plan.add_step(upgrade_charm_plan)
+
+        return plan
+
+    def _try_getting_channel(self, charm_channel: str) -> str:
```
gabrielcocenza, line 55: maybe something like this would be more clear:

```python
@property
def channel(self):
    return self._channel

@channel.setter
def channel(self, charm_channel):
    try:
         OpenStackRelease(charm_channel.split("/")[0])
         self._channel = charm_channel
    except ValueError as exc:
         ....
```

cou/apps/subordinate.py
```diff
@@ -0,0 +1,74 @@
...
+        try:
+            OpenStackRelease(charm_channel.split("/")[0])
+            return charm_channel
+        except ValueError as exc:
+            logger.error(
+                "Unable to determine the OpenStack version from channel: %s, %s",
+                charm_channel,
+                exc,
+            )
+            raise HaltUpgradePlanGeneration from exc
```
gabrielcocenza, line 74: IMO `ApplicationError` is more appropriate in that case. If an O7K subordinate don't have the channel as expected, I think we can stop the upgrading process

cou/utils/openstack.py
```diff
@@ -330,8 +352,8 @@ def find_compatible_versions(cls, charm: str, version: str) -> list[OpenStackRel
         return []
 
     @classmethod
-    def is_charm_supported(cls, charm: str) -> bool:
-        """Check if a charm is supported or not in the OpenStackCodenameLookup.
+    def is_primary_charm_supported(cls, charm: str) -> bool:
```
gabrielcocenza, line 355: maybe `is_charm_version_supported` represent better what we are trying to do because we support a group of charms, but some of then we don't support version, right?

cou/apps/subordinate.py
esunar: ok

cou/apps/subordinate.py
esunar: ok

cou/utils/openstack.py
esunar: ok

cou/apps/subordinate.py
```diff
@@ -0,0 +1,80 @@
...
+    def generate_upgrade_plan(self, target: str) -> UpgradeStep:
+        """Generate full upgrade plan for an Application.
+
+        :param target: OpenStack codename to upgrade.
+        :type target: str
+        :return: Full upgrade plan if the Application is able to generate it.
+        :rtype: UpgradeStep
+        """
+        plan = UpgradeStep(description=f"Upgrade {self.name}", parallel=False, function=None)
+        upgrade_charm_plan = self._get_upgrade_charm_plan(OpenStackRelease(target))
```
gabrielcocenza, line 49: Don't we need to also refresh the current channel?
It might be good also explaining that we are not updating packages because the principal unit will do it.

cou/apps/app.py
```diff
@@ -74,8 +79,8 @@ def create(
         :rtype: Optional[OpenStackApplication]
         """
         # pylint: disable=too-many-arguments
-        if OpenStackCodenameLookup.is_charm_supported(charm):
-            if status.subordinate_to:
+        if is_charm_supported(charm):
+            if status.subordinate_to and charm not in SUBORDINATES:
```
agileshaw, line 83: I don't think we need to check for `charm not in SUBORDINATES` here. If `is_charm_supported(charm)`, then we already know the charm is either principal or a known subordinate.

cou/apps/subordinate.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __post_init__(self) -> None:
+        """Initialize the Application dataclass."""
+        self.channel = self.status.charm_channel
+        self.charm_origin = self.status.charm.split(":")[0]
+        self.os_origin = self._get_os_origin()
+
+    @property
+    def current_os_release(self) -> OpenStackRelease:
+        """Assume latest version since principal already upgraded the packages."""
```
agileshaw, line 37: This docstring is not explanatory. We could change it to something like:
```suggestion
        """Infer the OS release from subordinate charm's channel.
        
        We cannot determine the OS release base on workload packages because the principal charm has already upgraded the packages.
        :return: OpenStackRelease object.
        :rtype: OpenStackRelease
        """
```

Also added the `:return` variable descirption in docstring

cou/apps/subordinate.py
```diff
@@ -0,0 +1,80 @@
...
+logger = logging.getLogger(__name__)
+
+
+@AppFactory.register_application(SUBORDINATES)
+class OpenStackSubordinateApplication(OpenStackApplication):
+    """Subordinate application class."""
+
+    def __post_init__(self) -> None:
+        """Initialize the Application dataclass."""
+        self.channel = self.status.charm_channel
```
agileshaw, line 31: There are `channel` getter/setter methods created later in the class, why do we still need a `self.channel` variable here?

cou/apps/subordinate.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            OpenStackRelease(charm_channel.split("/")[0])
+            self._channel = charm_channel
+        except ValueError as exc:
+            logger.error(
+                "Unable to determine the OpenStack version from channel: %s, %s",
+                charm_channel,
+                exc,
+            )
+            raise ApplicationError from exc
```
agileshaw, line 80: Since now we are adopting the CLI guideline, please add a message to the exception

agileshaw (changes requested):

cou/apps/subordinate.py
gabrielcocenza: We are validating `channel` and for that you can see that the setter accepts a value named `charm_channel`. That is why is still needed. This [article](https://www.programiz.com/python-programming/property) has an example of that.

cou/apps/app.py
esunar: hacluster charm is an example

cou/apps/subordinate.py
esunar: ok

cou/apps/subordinate.py
esunar: I need to revert this change because when creating object channel is set to "" and it creates exception since we do not allow this.

cou/apps/subordinate.py
esunar: done

cou/apps/subordinate.py
esunar: added refresh

cou/apps/subordinate.py
gabrielcocenza: I think using getter/setter is more appropriate. Maybe if we remove `channel` as a required parameter to instantiate `OpenStackApplication` and do it on `__post_init__` is a better solution. I believe the same behavior that you were using on ` OpenStackSubordinateApplication` should be applied on `OpenStackApplication`. Changing on `OpenStackApplication` would fix for both.

cou/utils/openstack.py
```diff
@@ -71,6 +71,28 @@
...
+    "cinder-purestorage",
+    "keystone-kerberos",
+    "keystone-ldap",
+    "keystone-saml-mellon",
+    "magnum-dashboard",
+    "manila-dashboard",
+    "manila-generic",
+    "masakari-monitors",
+    "neutron-api-plugin-ironic",
+    "neutron-api-plugin-ovn",
```
gabrielcocenza, line 90: I think [neutron-api-plugin-arista](https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html) can be also included here

cou/apps/subordinate.py
esunar: done

cou/utils/openstack.py
esunar: added

cou/apps/app.py
agileshaw: I was misled by the docstring change in `is_charm_version_supported` function. I though the lookup file only contains principal charms because of the word "primary". I suggestion changing `primary` to `core` in that function's docstring.

But still, I think this is a very confusing approach. `is_charm_supported == true` should indicate the charm is supported, so we should not need to have a condition afterwards to exclude some charms. I suggest either remove hacluster from the lookup file till we are ready to support it, or have a EXCLUDES list in openstack.py to exclude some charms and add `charm not in EXCLUDES` in `is_charm_supported` function.

cou/apps/subordinate.py
```diff
@@ -0,0 +1,89 @@
...
+        :type charm_channel: str
+        :raises ApplicationError: Exception raised when channel is not a valid OpenStack
+            channel.
+        """
+        try:
+            OpenStackRelease(charm_channel.split("/")[0])
+            self._channel = charm_channel
+        except ValueError as exc:
+            raise ApplicationError(
+                f"Unable to determine the OpenStack version from channel: {charm_channel}, {exc}"
```
agileshaw, line 88: According to the CLI guideline, `Cannot` is preferred in this case
```suggestion
                f"Cannot determine the OpenStack version from channel: {charm_channel}: {exc}"
```

cou/apps/app.py
esunar: I fixed the docstring. We can talk the details.

esunar: removed hacluster from csv file and no need for that block.

cou/apps/subordinate.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def generate_upgrade_plan(self, target: str) -> UpgradeStep:
+        """Generate full upgrade plan for an Application.
+
+        :param target: OpenStack codename to upgrade.
+        :type target: str
+        :return: Full upgrade plan if the Application is able to generate it.
+        :rtype: UpgradeStep
+        """
+        plan = UpgradeStep(description=f"Upgrade {self.name}", parallel=False, function=None)
```
aieri, line 54: This message should be made consistent with the one from the superclass

cou/apps/subordinate.py
```diff
@@ -0,0 +1,89 @@
...
+    def channel(self) -> str:
+        """Get charm channel of the application.
+
+        :return: Charm channel. E.g: ussuri/stable
+        :rtype: str
+        """
+        return self._channel
+
+    @channel.setter
+    def channel(self, charm_channel: str) -> None:
```
aieri, line 75: when do we need to set the channel?

cou/steps/plan.py
```diff
@@ -42,10 +45,46 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
         UpgradeStep(description="backup mysql databases", parallel=False, function=backup)
     )
 
-    upgrade_plan = UpgradeStep(
-        description="Application(s) upgrade plan", parallel=False, function=None
+    principle_upgrade_plan = await create_upgrade_group(
```
aieri, line 48:
```suggestion
    principal_upgrade_plan = await create_upgrade_group(
```

cou/steps/plan.py
```diff
@@ -42,10 +45,46 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
...
+        analysis_result=analysis_result,
+        description="Principal(s) upgrade plan",
+        target=target,
+        filter_function=lambda app: not isinstance(app, OpenStackSubordinateApplication),
     )
-    for app in analysis_result.apps:
+    plan.add_step(principle_upgrade_plan)
```
aieri, line 54:
```suggestion
    plan.add_step(principal_upgrade_plan)
```

cou/steps/plan.py
```diff
@@ -42,10 +45,46 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
...
+    :type target: str
+    :param description: Description of the upgrade step.
+    :type description: str
+    :param filter_function: Function to filter applications.
+    :type filter_function: Callable[[OpenStackApplication], bool]
+    :return: Upgrade group.
+    :rtype: UpgradeStep
+    """
+    group_upgrade_plan = UpgradeStep(description=description, parallel=False, function=None)
+    for app in (app for app in analysis_result.apps if filter_function(app)):
```
aieri, line 87: you called it filter_function, you might as well pass it to a filter ;)
```suggestion
    for app in filter(filter_function, analysis_result.apps):
```

aieri (changes requested):

cou/apps/subordinate.py
```diff
@@ -0,0 +1,89 @@
...
+from cou.utils.openstack import SUBORDINATES, OpenStackRelease
+
+logger = logging.getLogger(__name__)
+
+
+@AppFactory.register_application(SUBORDINATES)
+class OpenStackSubordinateApplication(OpenStackApplication):
+    """Subordinate application class."""
+
+    def __post_init__(self) -> None:
```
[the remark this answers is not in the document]
esunar: done

cou/apps/subordinate.py
```diff
@@ -0,0 +1,89 @@
...
+        """Initialize the Application dataclass."""
+        self.channel = self.status.charm_channel
+        self.charm_origin = self.status.charm.split(":")[0]
+        self.os_origin = self._get_os_origin()
+
+    @property
+    def current_os_release(self) -> OpenStackRelease:
+        """Infer the OS release from subordinate charm's channel.
+
+        We cannot determine the OS release base on workload packages because the principal charm
```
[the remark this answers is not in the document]
esunar: ok

cou/apps/subordinate.py
esunar: ok

cou/steps/plan.py
esunar: ok

cou/steps/plan.py
esunar: ok

cou/steps/plan.py
esunar: ok

cou/utils/openstack.py
```diff
@@ -330,8 +353,8 @@ def find_compatible_versions(cls, charm: str, version: str) -> list[OpenStackRel
         return []
 
     @classmethod
-    def is_charm_supported(cls, charm: str) -> bool:
-        """Check if a charm is supported or not in the OpenStackCodenameLookup.
+    def is_charm_version_supported(cls, charm: str) -> bool:
```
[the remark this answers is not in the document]
esunar: ok

cou/utils/openstack.py
```diff
@@ -342,3 +365,14 @@ def is_charm_supported(cls, charm: str) -> bool:
         if not cls._OPENSTACK_LOOKUP:
             cls._OPENSTACK_LOOKUP = cls._generate_lookup(cls._DEFAULT_CSV_FILE)
         return cls._OPENSTACK_LOOKUP.get(charm, False)
```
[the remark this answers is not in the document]
esunar: fixed

cou/utils/openstack.py
```diff
@@ -318,7 +341,7 @@ def find_compatible_versions(cls, charm: str, version: str) -> list[OpenStackRel
         :rtype: list[str]
         """
         compatible_os_releases: list[OpenStackRelease] = []
-        if cls.is_charm_supported(charm):
+        if cls.is_charm_version_supported(charm):
             for openstack_release, version_range in cls._OPENSTACK_LOOKUP[charm].items():
```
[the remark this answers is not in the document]
esunar: fixed

cou/utils/openstack.py
```diff
@@ -342,3 +365,14 @@ def is_charm_supported(cls, charm: str) -> bool:
...
+
+def is_charm_supported(charm: str) -> bool:
+    """Check if a charm upgrade is supported.
+
+    :param charm: Name of the charm.
+    :type charm: str
+    :return: True if supported, else False
+    :rtype: bool
+    """
+    return OpenStackCodenameLookup.is_charm_version_supported(charm) or charm in SUBORDINATES
```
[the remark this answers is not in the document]
esunar: now it is consistent

aieri (changes requested):
LGTM, we're only missing moving the setter/getter of channel to the superclass as we had agreed

agileshaw (commented):
For the functional code part, I just have one small suggestion inline. Haven't had the time to check the test part, will look into it next week.

cou/apps/subordinate.py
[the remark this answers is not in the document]
esunar: We cannot move them to superclass yet. Since OpenStack charms those are not principals sill exits. We cannot move them to superclass after pr #42

cou/utils/openstack.py
```diff
@@ -318,27 +341,37 @@ def find_compatible_versions(cls, charm: str, version: str) -> list[OpenStackRel
...
     @classmethod
-    def is_charm_supported(cls, charm: str) -> bool:
-        """Check if a charm is supported or not in the OpenStackCodenameLookup.
+    def lookup(cls, charm: str) -> dict:
+        """Check if a core OpenStack charm is supported or not in the OpenStackCodenameLookup.
 
         This function also generate the lookup if _OPENSTACK_LOOKUP is empty.
         :param charm: name of the charm
         :type charm: str
         :return: If supported return True, else False
```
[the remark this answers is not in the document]
esunar: done

tests/unit/apps/test_app.py
```diff
@@ -440,10 +440,11 @@ def test_upgrade_plan_application_already_disable_action_managed(status, config)
         "Change charm config of 'my_keystone' 'openstack-origin' to 'cloud:focal-victoria'",
         "Check if the workload of 'my_keystone' has been upgraded",
     ]
-    assert upgrade_plan.description == "Upgrade plan for 'my_keystone' from: ussuri to victoria"
+    assert upgrade_plan.description == "Upgrade plan for 'my_keystone' to victoria"
     assert_plan_description(upgrade_plan, steps_description)
 
 
+@unittest.skip("subordinates are removed")
```
gabrielcocenza, line 447: Shouldn't we remove this unit test instead of skipping it?

tests/unit/conftest.py
```diff
@@ -150,6 +150,13 @@ def status():
     mock_mysql_router.subordinate_to = ["keystone"]
     mock_mysql_router.units = {}
 
+    # subordinate application
```
gabrielcocenza, line 153:
```suggestion
    # OpenStack subordinate application
```

tests/unit/steps/test_steps_plan.py
```diff
@@ -34,30 +34,25 @@ async def test_generate_plan(mocker, apps):
...
+    assert len(plan.sub_steps) == 3
 
     sub_step_back_up = plan.sub_steps[0]
     assert sub_step_back_up.description == "backup mysql databases"
     assert not sub_step_back_up.parallel
     assert sub_step_back_up.function == backup
 
     sub_step_upgrade_plan = plan.sub_steps[1]
-    assert sub_step_upgrade_plan.description == "Application(s) upgrade plan"
+    assert sub_step_upgrade_plan.description == "Principal(s) upgrade plan"
```
gabrielcocenza, line 45: I think it would be good to include `keystone-ldap` on `apps` to also show `Subordinate(s) upgrade plan`

tests/unit/apps/test_subordinate.py
```diff
@@ -0,0 +1,75 @@
...
+        plan.sub_steps[0].description
+        == "Refresh 'my_keystone_ldap' to the latest revision of 'ussuri/stable'"
+    )
+    assert (
+        plan.sub_steps[1].description
+        == "Upgrade 'my_keystone_ldap' to the new channel: 'victoria/stable'"
+    )
+
+
+def test_try_getting_channel(status):
```
gabrielcocenza, line 62:
```suggestion
def test_channel_setter(status):
```

tests/unit/apps/test_subordinate.py
```diff
@@ -0,0 +1,75 @@
...
+    app_status = status["keystone-ldap"]
+    app = OpenStackSubordinateApplication(
+        "my_keystone_ldap", app_status, {}, "my_model", "keystone-ldap"
+    )
+
+    app.channel = "ussuri/stable"
+    assert app.channel == "ussuri/stable"
```
gabrielcocenza, line 69: maybe we could have two tests using `parametrize`. One showing that `<openstack>/release` doesn't raise `ApplicationError` and other that does.

gabrielcocenza (changes requested):

tests/unit/apps/test_app.py
esunar: I can remove it after pr #42 is merged.

cou/apps/subordinate.py
```diff
@@ -0,0 +1,87 @@
...
+        :type charm_channel: str
+        :raises ApplicationError: Exception raised when channel is not a valid OpenStack
+            channel.
+        """
+        try:
+            OpenStackRelease(charm_channel.split("/")[0])
+            self._channel = charm_channel
+        except ValueError as exc:
+            raise ApplicationError(
+                f"Cannot determine the OpenStack version from channel: {charm_channel}: {exc}"
```
rgildein, line 86: I think we should remove `exc` form this message. It will be present in trace logs, since we are using `raise ApplicationError(...) from exc`.
Although we are not providing trace logs right now, but reported an issue #43 for it.

rgildein (commented):
overall I think it's ok

tests/unit/steps/test_steps_plan.py
esunar: done

tests/unit/apps/test_subordinate.py
esunar: done

tests/unit/apps/test_subordinate.py
esunar: done

cou/apps/subordinate.py
esunar: done

gabrielcocenza (dismissed):
LGTM

cou/apps/subordinate.py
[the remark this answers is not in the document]
aieri: Will be tracked in https://github.com/canonical/charmed-openstack-upgrader/issues/45

aieri (approved):

tests/unit/apps/test_app.py
agileshaw: I don't understand why we need to wait for #42 to be merged before removing this test since subordinates-related changes should be handled in this PR.

Even if we decide to leave the test here for now, the skip message is really confusing; should say something like "subordinate check is removed" or "added supports for some subordinates" etc

tests/unit/apps/test_subordinate.py
```diff
@@ -0,0 +1,88 @@
...
+    app = OpenStackSubordinateApplication(
+        "my_keystone_ldap", app_status, {}, "my_model", "keystone-ldap"
+    )
+
+    if throws_exception:
+        with pytest.raises(ApplicationError):
+            app.channel = channel
+    else:
+        app.channel = channel
+        assert app.channel == channel
```
agileshaw, line 88: It's better to separate the success and failure cases into two test functions. The current logic is a bit coupled.

agileshaw (changes requested):
I have two comments inline. I think we are ready to merge the PR after addressing them.

tests/unit/apps/test_subordinate.py
esunar: done

tests/unit/apps/test_app.py
esunar: removed

tests/unit/apps/test_subordinate.py
```diff
@@ -60,29 +60,40 @@ def test_generate_upgrade_plan(status):
...
-        ("victoria/stable", False),
-        ("xena/stable", False),
-        ("yoga/stable", False),
-        ("wallaby/stable", False),
-        ("wallaby/edge", False),
-        ("focal/edge", True),
-        ("latest/edge", True),
-        ("latest/stable", True),
-        ("something/stable", True),
+        ("ussuri/stable"),
```
agileshaw, line 65: Parameterize on a single variable doesn't need the parenthesis. Could you please remove them? (and also for the `invalid` test function)
```
@pytest.mark.parametrize( "channel", ["ussuri/stable", "victoria/stable", ...])

esunar: done

agileshaw (approved):
LGTM, thanks. Let's wait for the CI to pass and then merge.

aieri (approved):

PR #41, opened 2023-08-31
Improve exception catching

rgildein:
Add new COUException and define new way how we are catching exceptions in CLI.

Additionally, this PR adjusts exception messages to align with CLI guideline ([tone of voice](https://discourse.ubuntu.com/t/tone-of-voice/18883) and [punctuation](https://discourse.ubuntu.com/t/punctuation/18884))

### DIFF
--- a/cou/apps/app.py
+++ b/cou/apps/app.py
@@ -148,11 +148,10 @@ class OpenStackApplication:
         E.g: {"keystone/0": {'os_version': 'victoria', 'workload_version': '2:18.1'}}
     :type units: defaultdict[str, dict]
     :raises ApplicationError: When there are no compatible OpenStack release for the
-        workload version
+        workload version.
     :raises MismatchedOpenStackVersions: When units part of this application are running mismatched
         OpenStack versions.
-    :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation
-    :raises PackageUpgradeError: When the package upgrade fails.
+    :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation.
     """
 
     # pylint: disable=too-many-instance-attributes
@@ -184,7 +183,7 @@ class OpenStackApplication:
                 self.units[unit]["os_version"] = unit_os_version
             else:
                 raise ApplicationError(
-                    f"'{self.name}' with workload version {workload_version} has no compatible"
+                    f"'{self.name}' with workload version {workload_version} has no compatible "
                     "OpenStack release in the lookup."
                 )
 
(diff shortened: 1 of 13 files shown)


esunar (approved):

cou/apps/app.py
```diff
@@ -377,7 +362,11 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
                 str(self.current_os_release),
                 target,
             )
-            raise HaltUpgradePlanGeneration()
+            raise HaltUpgradePlanGeneration(
+                f"Application '{self.name}' already running {self.current_os_release} that is "
+                f"equal or greater version than {target}. Ignoring."
```
aieri, line 367:
```suggestion
                f"Application '{self.name}' already running {self.current_os_release}, which is "
                f"equal or greater than {target}. Ignoring."
```

cou/utils/juju_utils.py
```diff
@@ -253,17 +254,18 @@ async def async_get_unit_from_name(
     try:
         if model is None:
             model = await _async_get_model(model_name)
+
         units = model.applications[app].units
     except KeyError as exc:
-        msg = f"Application: {app} does not exist in current model"
-        logger.error(msg)
-        raise UnitNotFound(unit_name) from exc
+        raise UnitNotFound(f"Application {app} not found in model.") from exc
```
aieri, line 260: We know the model name so why not:
```suggestion
        raise UnitNotFound(f"Application {app} not found in model {model_name}.") from exc
```

cou/utils/juju_utils.py
```diff
@@ -510,8 +509,10 @@ async def _ensure_model_connected(self) -> None:
...
         """Check time.
 
-        :raises JujuWaiter.TimeoutException: if timeout occurs
+        :raises TimeoutException: if timeout occurs
         """
         if datetime.now() - self.start_time > self.timeout:
             self.log.debug("MODEL IS NOT IDLE in: %d seconds", self.timeout)
-            raise JujuWaiter.TimeoutException()
+            raise TimeoutException(
+                f"Model {self.model_name} is not stabilized for {self.timeout} seconds."
```
aieri, line 517:
```suggestion
                f"Model {self.model_name} has not stabilized after {self.timeout} seconds."
```

cou/utils/juju_utils.py
```diff
@@ -510,8 +509,10 @@ async def _ensure_model_connected(self) -> None:
     def _check_time(self) -> None:
         """Check time.
 
-        :raises JujuWaiter.TimeoutException: if timeout occurs
+        :raises TimeoutException: if timeout occurs
         """
         if datetime.now() - self.start_time > self.timeout:
             self.log.debug("MODEL IS NOT IDLE in: %d seconds", self.timeout)
```
aieri, line 515:
```suggestion
            self.log.debug("Model %s is not idle after %d seconds.", self.model_name, self.timeout)
```

aieri (changes requested):
Small improvements suggested

cou/utils/juju_utils.py
agileshaw: I didn't use `model_name` here because `model_name` is an optional input. I have re-arranged the function a bit to move model object assignment outside the try-except block (we were not trying to catch any exception for this part anyway). This way we can simply call `model.name` to get the actual name.

gabrielcocenza (approved):
LGTM

aieri (approved):

PR #42, opened 2023-08-31
Auxiliary charms

gabrielcocenza:
Add auxiliary charms support


cou/utils/auxiliary_focal.csv
```diff
@@ -0,0 +1,8 @@
+OpenStack charms,ussuri,victoria,wallaby,xena,yoga
```
esunar, line 1: We should remove there charms from openstack_lookup.csv

cou/apps/app.py
```diff
@@ -394,6 +394,18 @@ def post_upgrade_plan(self, target: OpenStackRelease) -> list[UpgradeStep]:
         """
         return [self._get_reached_expected_target_plan(target)]
 
+    def upgrade_description(self, target: OpenStackRelease) -> str:
+        """Top description for upgrading the application.
```
esunar, line 398: Not critical: I think this level of function splitting is overkill.

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+    def os_origin_config(self, target: OpenStackRelease) -> Optional[OpenStackRelease]:
+        """Identify the OpenStack release set on openstack-origin or source config.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: OpenStackRelease object or None if the app doesn't have os_origin config.
+        :rtype: Optional[OpenStackRelease]
+        """
+        os_origin_parsed = None
+        if self.os_origin is not None:
```
esunar, line 75:
```
if !self.os_origin
    return ....
```

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+            # Ex: "cloud:focal-ussuri" will result in "ussuri"
+            os_origin_parsed = self.os_origin.rsplit("-", maxsplit=1)[-1]
+            match os_origin_parsed:
```
esunar, line 78: why don't we use if else?

cou/utils/openstack.py
```diff
@@ -342,3 +360,74 @@ def is_charm_supported(cls, charm: str) -> bool:
...
+                        "zed":"quincy",
+                        "2023.1":"quincy"
+                    },
+                },
+            }
+        :rtype: defaultdict
+        """
+        ubuntu_series = ["focal", "jammy"]
+        track_to_openstack_mapping: defaultdict[Any, Any] = defaultdict(lambda: defaultdict(str))
+        for series in ubuntu_series:
```
esunar, line 399: Too nested and hard to understand.
Easiest way is to separate file reading from parsing.
Since we will not have much items in these csv files, reading into memory and parsing into model makes it more readable.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ skip_glob = [
 
 [tool.pylint]
 max-line-length = 99
+min-similarity-lines=10
```
esunar, line 39: Why this was required?

esunar (changes requested):

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @property
+    def expected_current_channel(self) -> str:
+        """Return the expected current channel based on the current OpenStack release.
+
+        Note that this is not necessarily equal to the "channel" property since it is
+        determined based on the workload version.
+
+        :return: The expected current channel for the application.
+        :rtype: Optional[str]
```
agileshaw, line 98:
```suggestion
        :rtype: str
```

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+            description = f"Migration of '{self.name}' from charmstore to charmhub"
+            switch = f"ch:{self.charm}"
+            channel = self.expected_current_channel
+        elif self.channel != self.expected_current_channel:
+            logger.warning(
+                "'%s' has the channel set to: %s that is different from the expected channel: %s",
+                self.name,
+                self.channel,
+                self.expected_current_channel,
+            )
```
agileshaw, line 141: IIRC our discussion, we will not refresh the charm if its current channel doesn't match the expectation. So I think we should `return None` here

cou/utils/openstack.py
```diff
@@ -199,6 +205,18 @@ def next_release(self) -> Optional[str]:
             logger.warning("There is no OpenStack release after %s", self.codename)
             return None
 
+    @property
+    def previous_release(self) -> Optional[str]:
+        """Return the previous OpenStack release codename.
+
+        :return: OpenStack release codename.
+        :rtype: str
```
agileshaw, line 213:
```suggestion
        :rtype: Optional[str]
```

cou/utils/openstack.py
```diff
@@ -199,6 +205,18 @@ def next_release(self) -> Optional[str]:
...
+        """
+        if self.index == 0:
+            logger.warning("There is no OpenStack release before %s", self.codename)
```
agileshaw, line 216:
```suggestion
            logger.warning("Cannot find an OpenStack release before %s", self.codename)
```

cou/utils/openstack.py
```diff
@@ -199,6 +205,18 @@ def next_release(self) -> Optional[str]:
             logger.warning("There is no OpenStack release after %s", self.codename)
```
agileshaw, line 205:
```suggestion
            logger.warning("Cannot find an OpenStack release after %s", self.codename)
```

cou/apps/app.py
agileshaw: I also agree that this is over-engineering

agileshaw (changes requested):
I have done a partial review and left some comments. Will look into the changes more thoroughly next week.

cou/apps/app.py
rgildein: I think it's ok since it can be overwrite during inherited, as it's done in  this [line](https://github.com/canonical/charmed-openstack-upgrader/pull/42/files#diff-a29275aa9b0cacf5c69098e1afcce54ff8c5c0c6cd0480d12a5b483939047b34R40). It's not used only in one place.

cou/apps/auxiliary.py
rgildein: Using `is not None` it's ok, but I agree that we could simple return the value, instead of defining `os_origin_parsed = None` and than do `OpenStackRelease(os_origin_parsed) if os_origin_parsed else None`.
Like this we are doing multiple if statements instead of single one.

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+                case "distro":
+                    # find the OpenStack release based on ubuntu series
+                    os_origin_parsed = LTS_SERIES[self.series]
+                case "":
+                    # if it's empty we consider the previous release from the target.
+                    # Ex: rabbitmq-server has empty "source" and receive target "victoria".
+                    # In that case it will be considered as ussuri and with the upgrade,
+                    # "source" config will be changed to "cloud:focal-victoria".
+                    os_origin_parsed = target.previous_release
+        return OpenStackRelease(os_origin_parsed) if os_origin_parsed else None
```
rgildein, line 88:
```suggestion
        if self.os_origin is None:
            return None

        # Ex: "cloud:focal-ussuri" will result in "ussuri"
        os_origin_parsed = self.os_origin.rsplit("-", maxsplit=1)[-1]
        if os_origin_parsed == "distro":
            # find the OpenStack release based on ubuntu series
            os_origin_parsed = LTS_SERIES[self.series]
        else:
            # if it's empty we consider the previous release from the target.
            # Ex: rabbitmq-server has empty "source" and receive target "victoria".
            # In that case it will be considered as ussuri and with the upgrade,
            # "source" config will be changed to "cloud:focal-victoria".
            os_origin_parsed = target.previous_release
        
        return OpenStackRelease(os_origin_parsed)
```

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+        :type target: OpenStackRelease
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return: Plan for refreshing the charm.
+        :rtype: Optional[UpgradeStep]
+        """
+        switch = None
+        channel = self.channel
+
+        description = f"Refresh '{self.name}' to the latest revision of " f"'{self.channel}'"
```
rgildein, line 129:
```suggestion
        description = f"Refresh '{self.name}' to the latest revision of '{self.channel}'"
```

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,180 @@
...
+        yoga.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation
+        :return: Plan that will add upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        os_origin_config = self.os_origin_config(target)
+        if self.current_os_release >= target and os_origin_config and os_origin_config >= target:
```
rgildein, line 170: If I', correct we can can compare `None >= target` or not? If yes the `and os_origin_config` is not needed.

cou/utils/openstack.py
```diff
@@ -199,6 +205,18 @@ def next_release(self) -> Optional[str]:
...
+    def previous_release(self) -> Optional[str]:
+        """Return the previous OpenStack release codename.
+
+        :return: OpenStack release codename.
+        :rtype: str
+        """
+        if self.index == 0:
+            logger.warning("There is no OpenStack release before %s", self.codename)
+            return None
+        return self.openstack_codenames[self.index - 1]
```
rgildein, line 218:
```suggestion
            return None

        return self.openstack_codenames[self.index - 1]
```

cou/utils/openstack.py
rgildein: I agree with Erhan here.

cou/apps/auxiliary.py
gabrielcocenza: It won't throw an exception, but comparing None with target will always result to false. I think we can remove `and os_origin_config`

cou/utils/openstack.py
gabrielcocenza: From the stand up, we come in the conclusion that is better to write a global variable to this.

cou/apps/auxiliary.py
gabrielcocenza: I refactored to use if else instead of match. However, we still need to keep `return OpenStackRelease(os_origin_parsed) if os_origin_parsed else None` because `target.previous_release` can be `None` and pylint will complain if we return direct

cou/apps/auxiliary.py
gabrielcocenza: Done

cou/apps/auxiliary.py
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,181 @@
...
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        os_origin_config = self.os_origin_config(target)
+        if self.current_os_release >= target and os_origin_config >= target:
+            logger.info(
+                "Application: '%s' already configured for %s release. Ignoring.",
+                self.name,
+                target,
+            )
+            raise HaltUpgradePlanGeneration()
```
esunar, line 177: Info message should be part of exception.

cou/steps/plan.py
```diff
@@ -47,6 +48,8 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
     )
     for app in analysis_result.apps:
         try:
+            if isinstance(app, AuxiliaryOpenStackApplication):
```
esunar, line 51: Just debug for every charm with its type. Putting condition just for debugging is not a good practice.
`logger.debug("'%s' is an %s charm", app.name, type(app))`

gabrielcocenza: I did this to somehow use `AuxiliaryOpenStackApplication`. I've removed this log and added some comments explaining that I need to import this object in order to register the applications.

tests/unit/apps/test_auxiliary.py
```diff
@@ -0,0 +1,154 @@
...
+    plan = app.generate_upgrade_plan(target)
+
+    steps_description = [
+        f"Upgrade software packages of '{app.name}' from the current APT repositories",
+        f"Refresh '{app.name}' to the latest revision of '3.8/stable'",
+        f"Change charm config of '{app.name}' 'source' to 'cloud:focal-victoria'",
+        f"Check if the workload of '{app.name}' has been upgraded",
+    ]
+
+    assert_plan_description(plan, steps_description)
```
agileshaw, line 48: Can we move plan generation testing to a separate function or group all the assert together? The current layout is an anti-pattern (should be "arrange-act-assert "instead of "arrange-act-assert-act-assert")?

tests/unit/apps/test_auxiliary.py
```diff
@@ -0,0 +1,154 @@
...
+    assert app.os_origin_config(OpenStackRelease(target)) is None
+
+    app = AuxiliaryOpenStackApplication(
+        "rabbitmq-server",
+        status["rabbitmq_server"],
+        {"source": {"value": ""}},
+        "my_model",
+        "rabbitmq-server",
+    )
+    assert app.os_origin_config(OpenStackRelease(target)) == "ussuri"
```
agileshaw, line 115: Similar as above, could we separate it into two tests to avoid anti-pattern? Each function should focus on testing 1 single thing.

cou/utils/openstack.py
```diff
@@ -125,6 +125,97 @@
...
+    :return:  Dictionary containing the services and charms
+    :rtype: dict
+    """
+    mapping_charms = mapping.copy()
+    ubuntu_series = mapping_charms.keys()
+    for series in ubuntu_series:
+        for charm_type, charms in CHARM_TYPES.items():
+            for charm in charms:
+                if charm_type in mapping_charms[series].keys():
+                    mapping_charms[series][charm] = mapping_charms[series][charm_type]
```
agileshaw, line 149: Having heavily nested "for" loops is not a good practice. We should discuss and come up with some  ways to simplify the logic

agileshaw (changes requested):

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,172 @@
...
+        :type os_release: OpenStackRelease
+        :raises ApplicationError: When there is no track compatible.
+        :return: The track of the auxiliary charm.
+        :rtype: str
+        """
+        try:
+            return OPENSTACK_TO_TRACK_MAPPING[self.series][self.charm][os_release.codename]
+        except KeyError as exc:
+            raise ApplicationError(
+                f"Not possible to find the track for '{self.charm}' on {os_release.codename}"
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def os_origin_config(self, target: OpenStackRelease) -> Optional[OpenStackRelease]:
+        """Identify the OpenStack release set on openstack-origin or source config.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: OpenStackRelease object or None if the app doesn't have os_origin config.
+        :rtype: Optional[OpenStackRelease]
+        """
+        # that means that the charm doesn't have source or openstack-origin config.
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+    def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
+        """Upgrade planning.
+
+        Auxiliary charms have multiple compatible OpenStack releases. In that case,
+        we check also the OpenStack origin from the "source" or "openstack-origin" to know
+        if it's necessary to change it. E.g: rabbitmq-server with workload version 3.8 is
+        compatible from ussuri to yoga. Even that is considered as yoga, we need to set the
+        source accordingly with the OpenStack components that might be in a lower version than
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,172 @@
...
+        :type target: OpenStackRelease
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation
+        :return: Plan that will add upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        os_origin_config = self.os_origin_config(target)
+        if self.current_os_release >= target and os_origin_config >= target:
+            msg = (
+                f"Application: '{self.name}' already configured for release equal or greater "
+                f"version than {target}. Ignoring."
```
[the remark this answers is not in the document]
gabrielcocenza: Done

tests/unit/apps/test_auxiliary.py
gabrielcocenza: I've separated into two unit tests

tests/unit/apps/test_auxiliary.py
gabrielcocenza: Done

cou/utils/openstack.py
gabrielcocenza: what you think about something like this?

```python
def add_charms(mapping: dict) -> dict:
    mapping_charms = mapping.copy()
    ubuntu_series = mapping_charms.keys()
    for series in ubuntu_series:
        for charm_type, charms in CHARM_TYPES.items():
            new_mapping = {
                charm: mapping_charms[series][charm_type]
                for charm in charms
                if charm_type in mapping_charms[series].keys()
            }
            mapping_charms[series].update(new_mapping)
    return mapping_charms
```

In the end of the day it's 3 for loops but it will feel like 2.

gabrielcocenza: what you think about something like this?

```python
def add_charms(mapping: dict) -> dict:
    mapping_charms = mapping.copy()
    ubuntu_series = mapping_charms.keys()
    for series in ubuntu_series:
        for charm_type, charms in CHARM_TYPES.items():
            new_mapping = {
                charm: mapping_charms[series][charm_type]
                for charm in charms
                if charm_type in mapping_charms[series].keys()
            }
            mapping_charms[series].update(new_mapping)
    return mapping_charms
```

In the end of the day it's 3 for loops but it will feel like 2.

gabrielcocenza: based on our conversation on the cou-sync, I've created a new parameter `family` during the creation of the application, so we can get rid of this function

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,165 @@
...
+        yoga.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation
+        :return: Plan that will add upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        os_origin_config = self.os_origin_config(target)
+        if self.current_os_release >= target and os_origin_config >= target:
```
aieri, line 155: why was it not necessary to check the origin setting for OpenStack control plane apps?

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,165 @@
...
+        :rtype: str
+        """
+        try:
+            return OPENSTACK_TO_TRACK_MAPPING[self.series][self.family][os_release.codename]
+        except KeyError as exc:
+            raise ApplicationError(
+                f"Cannot find a track of '{self.charm}' for {os_release.codename}"
+            ) from exc
+
+    def os_origin_config(self, target: OpenStackRelease) -> Optional[OpenStackRelease]:
```
aieri, line 47: it took me a bit to understand this function. You want to understand which OpenStack release is contained in the APT sources configured in this charm. Could we rename it to something more obvious?
Additionally, shouldn't this also be part of cou.utils.openstack? The resolution algorithm does not change among applications.

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,165 @@
...
+        os_origin_parsed: Optional[str] = self.os_origin.rsplit("-", maxsplit=1)[-1]
+        if os_origin_parsed == "distro":
+            # find the OpenStack release based on ubuntu series
+            os_origin_parsed = LTS_SERIES[self.series]
+        elif os_origin_parsed == "":
+            # if it's empty we consider the previous release from the target.
+            # Ex: rabbitmq-server has empty "source" and receive target "victoria".
+            # In that case it will be considered as ussuri and with the upgrade,
+            # "source" config will be changed to "cloud:focal-victoria".
+            os_origin_parsed = target.previous_release
```
aieri, line 69: this is a big assumption... what if a user has configured a PPA?

cou/apps/auxiliary.py
```diff
@@ -0,0 +1,165 @@
...
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        os_origin_config = self.os_origin_config(target)
+        if self.current_os_release >= target and os_origin_config >= target:
+            msg = (
+                f"Application: '{self.name}' already configured for release equal or greater "
+                f"than {target}. Ignoring."
+            )
+            logger.info(msg)
+            raise HaltUpgradePlanGeneration(msg)
```
aieri, line 161: I know you're aligning with the superclass (and that code is already merged), but two things occurred to me:
1. this function's role is only to produce a plan (if possible). Technically, it's not `upgrade_plan` that is choosing to ignore already upgraded apps, but rather the caller
2. similarly, an exception name should represent what happened, not what the `except` higher up in the stack may or may not do with it. This exception should rather be named `PlanGenerationError` if there's a genuine failure, or just return `None` if nothing needs to be done (and the plan generation - in a sense - succeeded)

cou/utils/openstack.py
```diff
@@ -214,14 +285,27 @@ def next_release(self) -> Optional[str]:
...
             return None
 
+    @property
+    def previous_release(self) -> Optional[str]:
+        """Return the previous OpenStack release codename.
+
+        :return: OpenStack release codename.
+        :rtype: Optional[str]
+        """
+        if self.index == 0:
```
aieri, line 303: using the EAFP style would better align with the `next_release` property

aieri (changes requested):

cou/apps/auxiliary.py
gabrielcocenza: I tried to explain in the docstring, but basically on the control plane apps we have a 1:1 map between workload_version and OpenStack release. On the auxiliary charms we don't. E.g: version 3.8 on rabbitmq-server compatible with ussuri to yoga. In that case, if we want to set `source` of rmq to the apt repository of the current cloud version, we need to know where the "source" is pointing.

Moreover, we are checking if the workload has been upgraded. So in case a control plane changes the `openstack-origin` and the workload doesn't change, an error will happen.

gabrielcocenza: However, this logic on OpenStack control plane would make no harm and we can use the same method for both classes. It's actually a good idea passing it to the super class and not need to overwrite the method

cou/utils/openstack.py
gabrielcocenza: The problem of using `EAFP` on `previous_release` is that an index -1 is possible and would result into the latest OpenStack release. E.g: you ask for the previous release of `diablo` and that would result in `caracal`. There is no `IndexError` for negative indexes.

cou/apps/auxiliary.py
gabrielcocenza: What do you think about renaming to `os_origin_from_apt_sources`? 
It's indeed doesn't change among apps. I think it's better if we move to `OpenStackApplication` instead of `cou.utils.openstack`.

cou/apps/app.py
```diff
@@ -91,31 +100,31 @@ def create(
...
 
         Use this method as decorator to register Applications that
         cannot be described appropriately by the OpenStackApplication class.
 
-        Example:
-        ceph_types = ["ceph-mon", "ceph-fs", "ceph-radosgw", "ceph-osd"]
-
-        @AppFactory.register_application(ceph_types)
-        class Ceph(OpenStackApplication):
+        @AppFactory.register_application(["ceph", "rabbitmq-server"])
```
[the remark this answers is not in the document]
gabrielcocenza: The benefit it's not populating charms from a family like we were doing on  this [comment](https://github.com/canonical/charmed-openstack-upgrader/pull/42#discussion_r1317257747). Now the application knows witch family they came from and this helps the auxiliary charms find their right channel track

cou/apps/auxiliary.py
```diff
@@ -71,20 +69,21 @@ def os_origin_config(self, target: OpenStackRelease) -> Optional[OpenStackReleas
...
-                    # if it's empty we consider the previous release from the target.
-                    # Ex: rabbitmq-server has empty "source" and receive target "victoria".
-                    # In that case it will be considered as ussuri and with the upgrade,
-                    # "source" config will be changed to "cloud:focal-victoria".
-                    os_origin_parsed = target.previous_release
+        if not self.os_origin:
+            return None
+
+        # Ex: "cloud:focal-ussuri" will result in "ussuri"
+        os_origin_parsed: Optional[str] = self.os_origin.rsplit("-", maxsplit=1)[-1]
```
rgildein, line 60: not-blocking: This comment is only my preferences of how you we can extract parameters from tuple, list, etc. For me it's more readable. 
```suggestion
        *_, os_origin_parsed: Optional[str] = self.os_origin.rsplit("-", maxsplit=1)
```

cou/apps/auxiliary.py
rgildein: I think that for now it's ok, but we should provide warning to let user know that we are doing such assumption.

cou/utils/openstack.py
rgildein: I could not see the `family` parameter? Will you pushed new version?

cou/apps/auxiliary.py
gabrielcocenza: This could lose our `idempotency`. Eg;  running the cou until change the current channel, abort and run it again. In that case we would start throwing exceptions all the time.

cou/apps/auxiliary.py
gabrielcocenza: In case user has configured ppa, this would raise an exception, because the string on the ppa won't match an valid OpenStack release codename. i'm handling this better in the new commit

cou/apps/auxiliary.py
gabrielcocenza: we can discuss that on the COU sync, but IMO the current approach makes sense.

If we return None in `upgrade_plan`, this means that `pre_upgrade_plan` and `post_upgrade_plan` can generate steps to execute and I think we don't want this.

I don't think this is an error from plan generation. It's just an application saying that for the target passed, we should do nothing to upgrade it.

cou/apps/app.py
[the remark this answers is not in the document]
aieri: when we had that discussion, I did not understand what `add_charms` was for. I have now checked it out and iiuc you're fundamentally trying to manage in a reasonable way a 3-dimensional data structure with aliases: on one hand you have to map series, o7k codenames, and charm names to a track, on the other you have several duplicates so you want to have aliases.
Additionally, you want the mappings to be easy to update when there's a new release.

I think it may be valuable to reframe this as a data structure/storage problem rather than a method/decorator problem, as solving this from the Python logic side makes things less readable.

I would suggest exploring using csv tables (like you did for workload mappings) as they are easy to update so you may not care anymore about adding support for charm name aliases. csvs are only 2D of course, so you'd have to choose how to step one dimension down, for example by having per-LTS tables. You could still then reconstruct a sort of 3D structure by loading them into a single multilevel dictionary (to be queried as `mapping[series][codename][charm]` or any other level permutation.

Another alternative would be using pandas mulitindexes, but I don't think it's worth it for our usecase.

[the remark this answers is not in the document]
gabrielcocenza: Based on our conversation from yesterday, it was decided to create a single csv file with all charms and not need to populate by charm families.

cou/apps/auxiliary.py
aieri: I like the proposal, but isn't the unpacking operator unnecessary since `maxsplit=1` ?

cou/apps/auxiliary.py
```diff
@@ -16,89 +16,62 @@
...
 
     def _get_refresh_charm_plan(
         self, target: OpenStackRelease, parallel: bool = False
     ) -> Optional[UpgradeStep]:
         """Get plan for refreshing the current channel.
 
         This function also identifies if charm comes from charmstore and in that case,
-        makes the migration.
+        makes the migration. This method overwrite OpenStackApplication method because
+        a channel track from auxiliary charms can have multiple OpenStack releases.
```
aieri, line 74:
```suggestion
        makes the migration. This method overwrites the corresponding method from the  OpenStackApplication superclass because channels of "auxiliary" charms can be compatible with multiple OpenStack releases.
```

cou/utils/openstack.py
```diff
@@ -464,3 +409,48 @@ def is_charm_supported(charm: str) -> bool:
...
+    See the following url for more details:
+    https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html
+
+    :return: Dictionary containing the tracks by charm name, series and OpenStack release.
+    :rtype: dict[namedtuple, str]
+    """
+    track_mapping = {}
+    with open(
+        Path(__file__).parent / "openstack_to_track_mapping.csv",
+        encoding=encodings.utf_8.getregentry().name,
```
aieri, line 429: `encodings.utf_8.getregentry().name` is just... 'utf-8'. Why not use the string directly?

aieri (changes requested):

cou/apps/auxiliary.py
aieri: yeah leaving it in `OpenStackApplication` feels more right after all. You chosen name is better, but perhaps we can further refine it. How about `apt_source_codename` and making it a property (if you drop the target parameter)?

cou/apps/auxiliary.py
aieri: not really; whether `pre_upgrade_plan` and `post_upgrade_plan` steps will be generated (and executed) depends on how `generate_upgrade_plan` (`upgrade_plan`'s caller) will handle either `None` or an exception, which is my point: knowledge of the caller's implementation should not leak into your called method. I don't disagree with wanting to halt plan generation if we reach line 161 (I'd have to think more about it), I disagree with using an exception to signal a not-an-error situation.

`upgrade_plan` generates a plan to upgrade an application. If you ask an application to upgrade to a version it's already on, the plan is a no-op, and a no-op is not an error, so I think no exception should be raised.

[Here](https://stackoverflow.com/questions/1313812/raise-exception-vs-return-none-in-functions) is a pretty good discussion around the semantics of using returned objects vs raising exceptions to interact with callers.

cou/utils/openstack.py
aieri: duh of course... :clown_face: :facepalm:

cou/apps/auxiliary.py
aieri: in any case, changes to these lines should be handled in a separate PR

cou/apps/auxiliary.py
aieri: you could handle that case by checking if the current channel equals the target one (in which case there'd be nothing to do), but still throw an exception in other situations

cou/utils/openstack.py
rgildein: I ask the same question @esunar and he provide sufficient answer.

cou/utils/openstack.py
```diff
@@ -464,3 +409,48 @@ def is_charm_supported(charm: str) -> bool:
...
+    ) as csv_file:
+        csv_reader = csv.DictReader(csv_file, delimiter=",")
+        for row in csv_reader:
+            track_key = TrackKeys(
+                charm=row["charm"], series=row["series"], os_release=row["os_release"]
+            )
+            track_mapping[track_key] = row["track"]
```
rgildein, line 436: Not suggesting to any change. My original idea was to define this part directly in python e.g.
```python
OPENSTACK_TO_TRACK_MAPPING = {
    TrackKeys("ceph-mon", "focal", "ussuri"): "octopus",
    ...
}
```

cou/utils/openstack.py
```diff
@@ -464,3 +409,48 @@ def is_charm_supported(charm: str) -> bool:
...
+    return track_mapping
+
+
+OPENSTACK_TO_TRACK_MAPPING = _generate_track_mapping()
+
+
+def openstack_to_track(charm: str, series: str, os_release: OpenStackRelease) -> Optional[str]:
```
rgildein, line 443: not-blocking: I think that this function is unnecessary. Since you can use
```python
>>> OPENSTACK_TO_TRACK_MAPPING[("ceph-mon", "focal", "ussuri")]
'octopus'
>>> OPENSTACK_TO_TRACK_MAPPING.get(("ceph-mon", "focal", "ussuri"))
'octopus'
```

rgildein (dismissed):
Overall it looks good to me.

cou/utils/openstack.py
aieri: True, but it provides a nicer interface for callers that are outside this module. If you remove it, the caller needs to import `OPENSTACK_TO_TRACK_MAPPING` and understand its internal structure.

Consider also that pretty early on when we were discussing using or not using zaza utils, we decided to try to keep the "openstack utilities" as separate as possible from the rest of COU, so that we would at some point be able to release them as a separate package (whether that will actually ever happen is a separate question)

cou/apps/auxiliary.py
gabrielcocenza: I guess I found a way to have a unified `_get_refresh_charm_plan` for both `OpenStackApplication` and `OpenStackAuxiliaryApplication`.

For that I've changed the code to have a property named `channel_codename` that tries to get the OpenStack release by the channel. On auxiliary charms a channel can have multiple Openstack releases and in that case getting the maximum version seems to work.

cou/apps/auxiliary.py
gabrielcocenza: Done

cou/apps/auxiliary.py
gabrielcocenza: Created #55 to have a track of this issue.

cou/apps/app.py
```diff
@@ -276,6 +281,49 @@ def current_os_release(self) -> OpenStackRelease:
...
+        # that means that the charm doesn't have "source" or "openstack-origin" config.
+        if self.origin_setting is None:
+            return None
+
+        # Ex: "cloud:focal-ussuri" will result in "ussuri"
+        *_, os_origin_parsed = self.os_origin.rsplit("-", maxsplit=1)
+        if os_origin_parsed == "distro":
+            # find the OpenStack release based on ubuntu series
+            os_origin_parsed = DISTRO_TO_OPENSTACK_MAPPING[self.series]
+        elif os_origin_parsed == "":
```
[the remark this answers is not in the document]

cou/apps/app.py
```diff
@@ -362,20 +394,14 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
...
-                ),
-                self.name,
-                str(self.current_os_release),
-                target,
-            )
-            raise HaltUpgradePlanGeneration(
-                f"Application '{self.name}' already running {self.current_os_release} which is "
-                f"equal or greater than {target}. Ignoring."
+        os_origin_apt_sources = self.os_origin_from_apt_sources(target)
+        if self.current_os_release >= target and os_origin_apt_sources >= target:
```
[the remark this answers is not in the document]
gabrielcocenza: I didn't change the logic in the `upgrade_plan` and right now this comparison `os_origin_apt_sources >= target` will result into False if `os_origin_apt_sources` is None and will run the upgrade.

I don't think this is problematic, but if you think it is we can change the logic. In that case we would have to deal separately cases like `vault` that doesn't have "openstack-origin" or "source" from charms that are setting with empty string

cou/apps/auxiliary.py
```diff
@@ -16,89 +16,62 @@
...
         :return: The expected current channel for the application.
         :rtype: str
         """
-        return f"{self.openstack_to_track(self.current_os_release)}/stable"
+        track = openstack_to_track(self.charm, self.series, self.current_os_release)
+        if track:
+            return f"{track}/stable"
+
+        raise ApplicationError(
+            f"Cannot find a track of '{self.charm}' for {self.current_os_release.codename}"
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -16,89 +16,62 @@
...
+        )
 
     def target_channel(self, target: OpenStackRelease) -> str:
         """Return the channel based on the target passed.
```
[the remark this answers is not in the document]
gabrielcocenza: Done

cou/apps/auxiliary.py
```diff
@@ -128,38 +101,9 @@ def _get_refresh_charm_plan(
         return UpgradeStep(
             description=description,
             parallel=parallel,
-            function=async_upgrade_charm,
+            function=upgrade_charm,
             application_name=self.name,
             channel=channel,
```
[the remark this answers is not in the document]
gabrielcocenza: Please see this [conversation topic](https://github.com/canonical/charmed-openstack-upgrader/pull/42#discussion_r1323650333) to see if you agree having a single `_get_refresh_charm_plan` for both classes

cou/utils/openstack.py
gabrielcocenza: I ended up removing this function :grimacing:. I've added a reverse map to be able to know the compatible OpenStack releases of a certain track. Let me know if I should re-add this function

rgildein: I see point of easy to use, but i do not think that COU will be used as package, so that's I still think it's not important. For developer is good to be able to understand the structure and still I will check what function is returning to understand it. As developer that will be good chance that I use auto import and that will import `OPENSTACK_TO_TRACK_MAPPING` instead of this function. 

Now I read second paragraph and see that we want to use openstack utils as potential package. I do not have strong opinion here, so I think it's up to you @gabrielcocenza.

rgildein (dismissed):

cou/utils/openstack.py
rgildein: After checking again ho function is used, maybe we can pack more to this function e.g. 
```python
def get_track(charm: str, series: str, os_release: OpenStackRelease) -> Optional[str]:
    track = OPENSTACK_TO_TRACK_MAPPING.get(self.charm, self.series, target)
    if track:
        return f"{track}/stable"
    
    return track
```
That will make more sense to me than packed basically `OPENSTACK_TO_TRACK_MAPPING.get(self.charm, self.series, target)` in function.

esunar (dismissed):

cou/apps/app.py
```diff
@@ -276,6 +281,62 @@ def current_os_release(self) -> OpenStackRelease:
             f"{os_versions}. This is not currently handled."
         )
 
+    @property
+    def apt_source_codename(self) -> Optional[OpenStackRelease]:
+        """Identify the OpenStack release set on "openstack-origin" or "source" config.
+
+        :raises ValueError: If os_origin_parsed it's not a valid OpenStack release or os_origin
```
agileshaw, line 288:
```suggestion
        :raises ApplicationError: If os_origin_parsed it's not a valid OpenStack release or os_origin
```

agileshaw (commented):
Just one docstring change needed, otherwise LGTM

cou/utils/openstack.py
esunar: Since all "UTF-8", "UTF8", "ut", "utf8", "utf-8", "utf-8", "utf_8" and a bunch of other strings can be used as 'utf-8'. Unfortunately python does not provide a simple constant.

cou/apps/app.py
aieri: also s/it's/is/

cou/apps/app.py
```diff
@@ -281,12 +281,12 @@ def current_os_release(self) -> OpenStackRelease:
...
-    def os_origin_from_apt_sources(self, target: OpenStackRelease) -> Optional[OpenStackRelease]:
+    @property
+    def apt_source_codename(self) -> Optional[OpenStackRelease]:
         """Identify the OpenStack release set on "openstack-origin" or "source" config.
 
-        :raises ValueError: If os_origin_parsed it's not a valid OpenStack release.
-        :param target: OpenStack release as target to upgrade.
-        :type target: OpenStackRelease
+        :raises ValueError: If os_origin_parsed it's not a valid OpenStack release or os_origin
+            is on an unexpected format (ppa, url and etc).
```
aieri, line 289:
```suggestion
            is in an unexpected format (ppa, url, etc).
```

cou/apps/app.py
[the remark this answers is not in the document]
aieri: thank you I think it's a lot cleaner now

cou/apps/app.py
```diff
@@ -296,33 +296,46 @@ def os_origin_from_apt_sources(self, target: OpenStackRelease) -> Optional[OpenS
...
-        try:
-            return OpenStackRelease(os_origin_parsed) if os_origin_parsed else None
-        except ValueError as exc:
+        elif self.os_origin == "":
+            return None
+
+        else:
             # probably because user set a ppa or an url
             raise ApplicationError(
                 f"'{self.name}' has invalid '{self.origin_setting}': {self.os_origin}"
```
aieri, line 319:
```suggestion
                f"'{self.name}' has an invalid '{self.origin_setting}': {self.os_origin}"
```

cou/apps/app.py
```diff
@@ -394,7 +407,7 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
         :return: Plan that will add upgrade as sub steps.
         :rtype: list[Optional[UpgradeStep]]
         """
-        os_origin_apt_sources = self.os_origin_from_apt_sources(target)
+        os_origin_apt_sources = self.apt_source_codename
         if self.current_os_release >= target and os_origin_apt_sources >= target:
```
aieri, line 411: At this point, the extra variable seems unnecessary
```suggestion
        if self.current_os_release >= target and self.apt_source_codename >= target:
```

cou/apps/app.py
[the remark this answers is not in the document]
aieri: no, I think you're right, it should be safe to upgrade a charm that has no origin set (i.e. distro)

cou/utils/openstack.py
aieri: weird...

aieri (changes requested):
We're down to really minor typos now, +1 on the logic. I don't love we're passing raw OPENSTACK_TO_TRACK structures around instead of offering query functions, but it's ok.

aieri (approved):

esunar (approved):

agileshaw (approved):
LGTM

PR #44, opened 2023-09-05
refactor usage of model name and CURRENT_MODEL

rgildein:
- improve juju_utils
- drop `async` prefix from all functions
- pass model_name from CLI to everywhere


cou/apps/app.py
```diff
@@ -32,11 +32,7 @@
 )
 from cou.steps import UpgradeStep
 from cou.utils.app_utils import upgrade_packages
-from cou.utils.juju_utils import (
-    async_get_status,
-    async_set_application_config,
-    async_upgrade_charm,
-)
+from cou.utils.juju_utils import get_status, set_application_config, upgrade_charm
```
esunar, line 35: I prefer to use all juju_utils function with a common prefix such as `jutils` or `utils` or `ut` it makes fırther refactoring easier

cou/utils/juju_utils.py
```diff
@@ -109,11 +107,11 @@ async def _disconnect(model: Model) -> None:
     :param model: the juju.model.Model object.
     :type model: Model
     """
-    if model is not None:
-        try:
+    try:
+        if model is not None:
```
esunar, line 111: We can skip this check, we do not care the exception anyway.

esunar (commented):
Overall LGTM. Just a few comments.

cou/utils/juju_utils.py
```diff
@@ -88,17 +84,19 @@ async def _async_get_model(model_name: Optional[str] = None) -> Model:
...
     """
     global CURRENT_MODEL
 
     model = CURRENT_MODEL
     if model is not None and _is_model_disconnected(model):
         await _disconnect(model)
-        model = None
-    if CURRENT_MODEL is None:
+
+    if model is None:
```
agileshaw, line 95: A disconnected model object is not `None`, so we are not able to enter this block after executing L93. We may unintentionally return a disconnected model at the end of this function

agileshaw (changes requested):
Aside from the agreed module import change (e.g. `from cou.utils import juju_utils`), I have left a few comments inline.

cou/apps/app.py
rgildein: I will use `from cou.utils import juju_utils` everywhere.

cou/utils/juju_utils.py
rgildein: To be honest I'm confused from this function completely. I'm reverting it back and I will reworked it in next PR where we use Model object instead of model_name pass to all other functions. Is that ok for you?

cou/apps/app.py
```diff
@@ -53,7 +49,7 @@ def create(
         name: str,
         status: ApplicationStatus,
         config: dict,
-        model_name: str,
+        model_name: str | None,
```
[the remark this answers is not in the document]
esunar: I don't think it is a severe issue however we have used Optional throughout the code. It would be nice if we all use same.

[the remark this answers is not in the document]
agileshaw: Since there's not yet a plan of deprecating Optional, I think we can continue using it and keep it consistent throughout the code.

cou/utils/juju_utils.py
agileshaw: Yes, that sounds great. Thanks.

agileshaw (approved):
LGTM, thanks

esunar (approved):

PR #46, opened 2023-09-05
VersionRange raise ValueError if upper is not bigger than lower.

### LINKED ISSUE #37: If upper, lower and compared versions are same it returns false
https://github.com/canonical/charmed-openstack-upgrader/blob/a4101a6972c1191628d95d1b24e2246a632054b5/cou/utils/openstack.py#L244

For example:

```
arr = [VersionRange("6.0.0", "6.0.0")]
if "6.0.0" in arr:
    print("I am in range")
else:
    print("I am not in range")

```

gabrielcocenza:
Closes #37

### DIFF
--- a/cou/utils/openstack.py
+++ b/cou/utils/openstack.py
@@ -254,9 +254,9 @@ class VersionRange:
     upper: str
 
     def __post_init__(self) -> None:
-        """Initialize the VersionRange dataclass and check values."""
+        """Initialize the VersionRange dataclass and check its values."""
         if Version(self.lower) >= Version(self.upper):
-            raise ValueError("Upper version it is not bigger than lower version.")
+            raise ValueError("The upper bound version is not higher than the lower bound version.")
 
     def __contains__(self, version: str) -> bool:
         """Magic method to check if a version is within the range.
--- a/cou/utils/openstack.py
+++ b/cou/utils/openstack.py
@@ -253,6 +253,11 @@ class VersionRange:
     lower: str
     upper: str
 
+    def __post_init__(self) -> None:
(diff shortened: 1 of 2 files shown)


cou/utils/openstack.py
```diff
@@ -253,6 +253,11 @@ class VersionRange:
     lower: str
     upper: str
 
+    def __post_init__(self) -> None:
+        """Initialize the VersionRange dataclass and check values."""
+        if Version(self.lower) >= Version(self.upper):
+            raise ValueError("Upper version it is not bigger than lower version.")
```
agileshaw, line 259:
```suggestion
            raise ValueError("The upper bound version is not higher than the lower bound version.")
```

cou/utils/openstack.py
```diff
@@ -253,6 +253,11 @@ class VersionRange:
     lower: str
     upper: str
 
+    def __post_init__(self) -> None:
+        """Initialize the VersionRange dataclass and check values."""
```
agileshaw, line 257:
```suggestion
        """Initialize the VersionRange dataclass and check its values."""
```

agileshaw (changes requested):
Overall LGTM, just some wording changes suggested inline

cou/utils/openstack.py
gabrielcocenza: Done

cou/utils/openstack.py
gabrielcocenza: Done

agileshaw (approved):
+1 from me. Thanks :)

esunar (approved):

PR #51, opened 2023-09-07
CLI Overhaul

agileshaw:
- Redesign CLI to align with guidelines and improves UI/UX
- Suppress traceback in console
- Renaming package to canonical-openstack-upgrader

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,20 +10,20 @@ different repository, and only use some of the code and ideas found here.
 ```bash
 rm *.snap
 snapcraft clean
-sudo snap remove charmed-openstack-upgrader --purge
+sudo snap remove canonical-openstack-upgrader --purge
 make build
-sudo snap install ./charmed-openstack-upgrader.snap --devmode
-sudo snap connect charmed-openstack-upgrader:juju-client-observe snapd
-sudo snap connect charmed-openstack-upgrader:dot-local-share-juju snapd
-sudo snap connect charmed-openstack-upgrader:local-cou snapd
+sudo snap install ./canonical-openstack-upgrader.snap --devmode
+sudo snap connect canonical-openstack-upgrader:juju-client-observe snapd
+sudo snap connect canonical-openstack-upgrader:dot-local-share-juju snapd
+sudo snap connect canonical-openstack-upgrader:local-cou snapd
 ```
 
 Then you can use ```cou```
 
 ## Environment Variables
 
-- `COU_DATA`: Main directory for Charmed Openstack Upgrader configuration files. Currently only log files are kept under
-  COU_DATA/logs
+- `COU_DATA`: Main directory for Canonical Openstack Upgrader configuration files. Currently only log files are kept under
+  COU_DATA/logs.
 - `JUJU_MODEL`: Active juju model to operate on.
 
 ## Supported Upgrade Paths
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -12,26 +12,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Entrypoint to the 'charmed-openstack-upgrader'."""
+"""Entrypoint to the 'canonical-openstack-upgrader'."""
 import argparse
 import logging
 import logging.handlers
-import os
-import pathlib
 import sys
-from datetime import datetime
-from typing import Any
+from typing import Any, Iterable, Optional
+
+import pkg_resources
 
 from cou.exceptions import COUException
+from cou.logging import setup_logging
+from cou.steps import UpgradeStep
 from cou.steps.analyze import Analysis
 from cou.steps.execute import execute
 from cou.steps.plan import generate_plan
 from cou.utils import juju_utils as utils
 
-COU_DIR_LOG = pathlib.Path(os.getenv("COU_DATA", ""), "log")
 AVAILABLE_OPTIONS = "cas"
+VERBOSITY_LEVEL = {0: "ERROR", 1: "WARNING", 2: "INFO", 3: "DEBUG", 4: "NOTSET"}
 
 logger = logging.getLogger(__name__)
+
+
+class CapitalisedHelpFormatter(argparse.HelpFormatter):
+    """Capitalize usage prefix."""
+
+    def add_usage(
+        self,
+        usage: Optional[str],
+        actions: Iterable[argparse.Action],
+        groups: Iterable[argparse._MutuallyExclusiveGroup],
+        prefix: Optional[str] = None,
+    ) -> None:
+        """Add usage with capitalized prefix.
+
+        :param usage: usage message.
+        :type usage: Optional[str]
+        :param actions: actions.
+        :type actions: Iterable[argparse.Action]
+        :param groups: Arguments to be parsed.
+        :type groups: Iterable[argparse._MutuallyExclusiveGroup]
+        :param prefix: Arguments to be parsed.
+        :type prefix: Optional[str]
+        """
+        if prefix is None:
+            prefix = "Usage: "
+        super().add_usage(usage, actions, groups, prefix)
 
 
 def parse_args(args: Any) -> argparse.Namespace:
@@ -42,112 +69,169 @@ def parse_args(args: Any) -> argparse.Namespace:
     :return: Arguments parsed to the cli execution.
     :rtype: argparse.Namespace
     """
+    # Configure top level argparser and its options
     parser = argparse.ArgumentParser(
-        description="Charmed OpenStack Upgrader(cou) is an application to upgrade Charmed "
+        description="Canonical OpenStack Upgrader(cou) is an application to upgrade Canonical "
         "OpenStack. Application identifies the lowest OpenStack version on the components and "
         "upgrade to the next version.",
-        formatter_class=argparse.RawDescriptionHelpFormatter,
+        formatter_class=CapitalisedHelpFormatter,
+        usage="%(prog)s [options] <command>",
         exit_on_error=False,
         add_help=False,
+        allow_abbrev=False,
     )
     parser.add_argument(
-        "--run",
-        help="Use this flag to run the upgrade, otherwise just print out the upgrade steps.",
-        action="store_true",
-        default=False,
+        "--version",
+        "-V",
+        action="version",
+        default=argparse.SUPPRESS,
+        help="Show version details.",
+        version=pkg_resources.require("canonical_openstack_upgrader")[0].version,
     )
     parser.add_argument(
-        "--log-level",
-        default="INFO",
-        dest="loglevel",
-        choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
-        type=str.upper,
-        help="Set the logging level. Defaults to INFO. This only affects stdout. The logfile "
-        "will be always DEBUG. The file location is in COU_DATA/log. You can set the "
-        "COU_DATA via environment variable. However it needs to be plugged to snap.",
-    )
-    parser.add_argument(
-        "--model",
-        default=None,
-        dest="model_name",
-        help="Set the model to operate on. If not set it gets the model name in this order:\n"
-        "  1 - Environment variable JUJU_MODEL,"
-        "  2 - Environment variable MODEL_NAME,"
-        "  3 - Current active juju model",
-    )
-    parser.add_argument(
-        "--interactive",
-        help="Run upgrade with prompt.",
-        action=argparse.BooleanOptionalAction,
-        default=True,
-    )
-    parser.add_argument(
-        "-h",
         "--help",
+        "-h",
         action="help",
         default=argparse.SUPPRESS,
         help="Show this help message and exit.",
     )
 
-    return parser.parse_args(args)
+    # Configure subcommand and their common flags
+    subparsers = parser.add_subparsers(
+        title="Commands:",
+        dest="command",
+    )
+    base_subparser = argparse.ArgumentParser(add_help=False)
+    base_subparser.add_argument(
+        "--model",
+        default=None,
+        dest="model_name",
+        type=str,
+        help="Set the model to operate on. If not set it gets the model name in this order:\n"
+        "  1 - Environment variable JUJU_MODEL,"
+        "  2 - Environment variable MODEL_NAME,"
+        "  3 - Current active juju model",
+    )
+    group = base_subparser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--verbose",
+        "-v",
+        default=0,
+        action="count",
+        dest="verbosity",
+        help="Increase logging verbosity in STDOUT. Repeat the 'v' in the short option "
+        "for more detail. Maximum verbosity is obtained with 4 (or more) "
+        "v's, i.e. -vvvv. \nNote that this doesn't affect the verbosity in logfile, "
+        "which will always have the maximum verbosity.",
+    )
+    group.add_argument(
+        "--quiet",
+        "-q",
+        default=False,
+        action="store_true",
+        dest="quiet",
+        help="Disable output in STDOUT.",
+    )
+
+    # Arg parser for "cou plan" sub-command
+    subparsers.add_parser(
+        "plan",
+        description="Show the steps for upgrading the cloud to the next release.",
+        help="Show the steps for upgrading the cloud to the next release.",
+        usage="cou run [options]",
+        parents=[base_subparser],
+    )
+
+    # Arg parser for "cou run" sub-command
+    run_parser = subparsers.add_parser(
+        "run",
+        description="Run the cloud upgrade.",
+        help="Run the cloud upgrade.",
+        usage="cou run [options]",
+        parents=[base_subparser],
+    )
+    run_parser.add_argument(
+        "--interactive",
+        help="Run upgrade with prompt.",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
+
+    # It no sub-commands or options are given, print help message and exit
+    if len(args) == 0:
+        parser.print_help()
+        sys.exit(0)
+
+    try:
+        return parser.parse_args(args)
+    except argparse.ArgumentError as exc:
+        print(f"Unrecognized sub-command: {exc}.")
+        parser.print_help()
+        sys.exit(1)
 
 
-def setup_logging(log_level: str = "INFO") -> None:
-    """Do setup for logging.
+def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
+    """Get a log level based on input options.
 
-    :returns: Nothing: This function is executed for its side effect
-    :rtype: None
+    :param quiet: Whether to run COU in quiet mode.
+    :type quiet: bool
+    :param verbosity: Verbosity level based on user's input.
+    :type verbosity: int
+    :return: Log level.
+    :rtype: str
     """
-    log_formatter_file = logging.Formatter(
-        fmt="%(asctime)s [%(name)s] [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    log_formatter_console = logging.Formatter(
-        fmt="%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    root_logger = logging.getLogger()
-    root_logger.setLevel("DEBUG")
+    if quiet:
+        return "CRITICAL"
+    return VERBOSITY_LEVEL[verbosity] if verbosity <= 4 else VERBOSITY_LEVEL[4]
 
-    # handler for the log file. Log level is DEBUG
-    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
-    file_name = f"{COU_DIR_LOG}/cou-{time_stamp}.log"
-    pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
-    log_file_handler = logging.FileHandler(file_name)
-    log_file_handler.setFormatter(log_formatter_file)
 
-    # handler for the console. Log level comes from the CLI
-    console_handler = logging.StreamHandler()
-    console_handler.setLevel(log_level)
-    console_handler.setFormatter(log_formatter_console)
-    # just cou logs on console
-    console_handler.addFilter(logging.Filter(__package__))
+async def get_plan(model_name: Optional[str] = None) -> UpgradeStep:
+    """Get the upgrade plan with steps.
 
-    root_logger.addHandler(log_file_handler)
-    root_logger.addHandler(console_handler)
-    logger.info("Logs of this execution can be found at %s", file_name)
+    :param model_name: Model name inputted by user.
+    :type model_name: Optional[str]
+    :return: Generated upgrade plan.
+    :rtype: UpgradeStep
+    """
+    model = await utils.async_set_current_model_name(model_name)
+    logger.info("Setting current model name: %s", model)
+    analysis_result = await Analysis.create()
+    logger.info(analysis_result)
+    upgrade_plan = await generate_plan(analysis_result)
+    print(upgrade_plan)
+    return upgrade_plan
+
+
+async def run_upgrade(model_name: Optional[str] = None, interactive: bool = True) -> None:
+    """Run cloud upgrade.
+
+    :param model_name: Model name inputted by user.
+    :type model_name: Optional[str]
+    :param interactive: Whether to run upgrade in the interactive mode.
+    :type interactive: bool
+    """
+    upgrade_plan = await get_plan(model_name)
+    await execute(upgrade_plan, interactive)
+    print("Upgrade completed.")
 
 
 async def entrypoint() -> None:
-    """Execute 'charmed-openstack-upgrade' command."""
+    """Execute 'canonical-openstack-upgrade' command."""
     try:
         args = parse_args(sys.argv[1:])
 
-        setup_logging(log_level=args.loglevel)
+        setup_logging(log_level=get_log_level(quiet=args.quiet, verbosity=args.verbosity))
 
-        model_name = await utils.async_set_current_model_name(args.model_name)
-        logger.info("Setting current model name: %s", model_name)
-
-        analysis_result = await Analysis.create()
-        print(analysis_result)
-        upgrade_plan = await generate_plan(analysis_result)
-        if args.run:
-            await execute(upgrade_plan, args.interactive)
-        else:
-            print(upgrade_plan)
+        match args.command:
+            case "run":
+                await run_upgrade(model_name=args.model_name, interactive=args.interactive)
+            case "plan":
+                await get_plan(args.model_name)
 
     except COUException as exc:
         logger.error(exc)
         sys.exit(1)
     except Exception as exc:  # pylint: disable=broad-exception-caught
-        logger.error("unexpected error occurred")
+        logger.error("Unexpected error occurred")
         logger.exception(exc)
         sys.exit(2)
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,8 @@
 [metadata]
-name = charmed-openstack-upgrader
+name = canonical-openstack-upgrader
 # See setup.py version generator
-url = https://github.com/esunar/charmed-openstack-upgrader
-description = Application for upgrading Charmed Openstack
+url = https://github.com/esunar/canonical-openstack-upgrader
+description = Application for upgrading Canonical Openstack
 long_description = file: README.md
 long_description_content_type = text/markdown; charset=UTF-8; variant=GFM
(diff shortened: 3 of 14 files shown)


esunar:
Should we add switches for --data-plane --control-plane ?

agileshaw:
@esunar I will add place-holder options for "control-plane" and "data-plane" aligning with our discussion result.

agileshaw:
Will add unit tests after getting +1s for the overall approach.

completions/bash/cou_bash_completion
```diff
@@ -0,0 +1,61 @@
+_cou_bash_completion() {
```
rgildein, line 1: Can this be done also for zsh? It works fine for bash.

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+
+async def get_upgrade_plan(model_name: Optional[str] = None) -> None:
+    """Get upgrade plan and print to console.
+
+    :param model_name: Model name inputted by user.
+    :type model_name: Optional[str]
+    """
+    upgrade_plan = await analyze_and_plan(model_name)
+    logger.info(upgrade_plan)
+    print(upgrade_plan)  # print plan to console even in quiet mode
```
rgildein, line 345: So if I run in in -vvv, I will see the plan twice?

rgildein (commented):
I was thinking that for console logging format we will use only message (no ascitime and levelname ) and use logging everywhere instead of print. I think it will be much cleaner than using some if statement to not outputs things twice (once with logging and once with print) or outputting it twice. 
Maybe there is better solution like `extra` argument for log. e.g. `logger.info("...", extra=...)`

cou/cli.py
esunar: I think it should be debug or trace.

esunar (dismissed):

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
-        fmt="%(asctime)s [%(name)s] [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    log_formatter_console = logging.Formatter(
-        fmt="%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    root_logger = logging.getLogger()
-    root_logger.setLevel("DEBUG")
+    if quiet:
+        return "CRITICAL"
+    return VERBOSITY_LEVEL[verbosity] if verbosity <= 4 else VERBOSITY_LEVEL[4]
```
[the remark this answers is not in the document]
agileshaw: Nice suggestion. Thanks :)

completions/bash/cou_bash_completion
agileshaw: Thanks for the suggestion. As discussed elsewhere, this will be handled later as a future enhancement. I opened an issue to track: #58

cou/cli.py
agileshaw: I personally think this can stay at info level, similar as the analyze result, because they provide cloud-related information rather than program implementation details.

setup.cfg
```diff
@@ -1,8 +1,8 @@
 [metadata]
-name = charmed-openstack-upgrader
+name = canonical-openstack-upgrader
 # See setup.py version generator
-url = https://github.com/esunar/charmed-openstack-upgrader
-description = Application for upgrading Charmed Openstack
+url = https://github.com/esunar/canonical-openstack-upgrader
```
aieri, line 4: why are we pointing to Erhan's fork?

setup.cfg
```diff
@@ -1,8 +1,8 @@
...
+description = Application for upgrading Canonical Openstack
```
aieri, line 5:
```suggestion
description = Application for upgrading Canonical Charmed OpenStack
```

setup.cfg
```diff
@@ -1,8 +1,8 @@
...
 long_description = file: README.md
 long_description_content_type = text/markdown; charset=UTF-8; variant=GFM
 author = Canonical DevOps Centres
```
aieri, line 8:
```suggestion
author = Canonical BootStack DevOps Centres
```

aieri: ... and maybe we drop the 'centres'?

snap/snapcraft.yaml
```diff
@@ -1,10 +1,10 @@
-name: charmed-openstack-upgrader
+name: canonical-openstack-upgrader
 base: core22
 adopt-info: cou
-summary: Charmed Openstack Upgrader
+summary: Canonical Openstack Upgrader
 license: Apache-2.0
 description: |
-  Snap for automatically upgrading Charmed Openstack and related
+  Snap for automatically upgrading Canonical Openstack and related
```
aieri, line 7:
```suggestion
  Snap for automatically upgrading Canonical Charmed OpenStack and related
```

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+        action="append",
+        help="Specify availability zones to upgrade.",
+        dest="availability_zones",
+    )
+
+    # Arg parser for "cou plan" sub-command
+    plan_parser = subparsers.add_parser(
+        "plan",
+        description="Show the steps for upgrading the cloud to the next release.",
+        help="Show the steps for upgrading the cloud to the next release.",
```
aieri, line 192: I want to avoid giving the impression we'd print juju cli commands
```suggestion
        description="Show the steps COU will take to upgrade the cloud to the next release.",
        help="Show the steps COU will take to upgrade the cloud to the next release.",
```

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+    subcommand_common_opts_parser = argparse.ArgumentParser(add_help=False)
+    subcommand_common_opts_parser.add_argument(
         "--model",
         default=None,
         dest="model_name",
+        type=str,
         help="Set the model to operate on. If not set it gets the model name in this order:\n"
         "  1 - Environment variable JUJU_MODEL,"
         "  2 - Environment variable MODEL_NAME,"
         "  3 - Current active juju model",
```
aieri, line 132: the newlines seem to be ignored on my terminal:
```
$ canonical-openstack-upgrader.cou plan --help
Usage: cou plan [options]

Show the steps for upgrading the cloud to the next release.

options:
  -h, --help            show this help message and exit
  --model MODEL_NAME    Set the model to operate on. If not set it gets the model name in this order: 1 - Environment variable JUJU_MODEL, 2 - Environment variable MODEL_NAME, 3 - Current active juju model
  --parallel            Run upgrade steps in parallel where possible.
  --verbose, -v         Increase logging verbosity in STDOUT. Repeat the 'v' in the short option for more detail. Maximum verbosity is obtained with 4 (or more) v's, i.e. -vvvv. Note that this doesn't affect the
                        verbosity in logfile, which will always have the maximum verbosity.
  --quiet, -q           Disable output in STDOUT.

Upgrade group:
  {control-plane,data-plane}
                        For more information about a upgrade group, run 'cou plan <upgrade-group>' -h.
    control-plane       Show the steps for upgrading the control-plane components.
    data-plane          Show the steps for upgrading the data-plane components.
```

completions/bash/cou_bash_completion
```diff
@@ -0,0 +1,61 @@
...
+    # Define subcommand-specific options and child commands
+    if [ $COMP_CWORD -eq 2 ]; then
+        cmd="${COMP_WORDS[1]}"
+        case "${cmd}" in
+            plan)
+                opts="--help --quite --verbose --parallel --model"
+                child_commands="control-plane data-plane"
+                ;;
+            run)
+                opts="--help --quite --verbose --parallel --model --interactive --no-interactive --i-really-mean-it"
```
aieri, line 26:
```suggestion
                opts="--help --quiet --verbose --parallel --model"
                child_commands="control-plane data-plane"
                ;;
            run)
                opts="--help --quiet --verbose --parallel --model --interactive --no-interactive --i-really-mean-it"
```

aieri: I'd also suggest dropping `--i-really-mean-it` (regardless of whether you agree on dropping the actual option): it'd be very unfortunate if someone caused downtime by mistake because they tabbed too hard ;)

completions/bash/cou_bash_completion
aieri, line 1: not a big deal, but autocompletion of the common options (e.g.  `--parallel`) stops if you select control or data plane.
```
$ canonical-openstack-upgrader.cou plan data-plane --
--availability-zone  --hostname           --machine            
```
Should common options be offered regardless? It's a little odd, especially for the control plane:
```
$ canonical-openstack-upgrader.cou plan control-plane # hitting tab prints nothing
$

$ canonical-openstack-upgrader.cou plan control-plane --help # lots of options here!
Usage: cou plan control-plane [options]

Show the steps for upgrading the control-plane components.

options:
  -h, --help          show this help message and exit
  --model MODEL_NAME  Set the model to operate on. If not set it gets the model name in this order: 1 - Environment variable JUJU_MODEL, 2 - Environment variable MODEL_NAME, 3 - Current active juju model
  --parallel          Run upgrade steps in parallel where possible.
  --verbose, -v       Increase logging verbosity in STDOUT. Repeat the 'v' in the short option for more detail. Maximum verbosity is obtained with 4 (or more) v's, i.e. -vvvv. Note that this doesn't affect the
                      verbosity in logfile, which will always have the maximum verbosity.
  --quiet, -q         Disable output in STDOUT.
```

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+    )
+
+    # Define options specific to data-plane child command and make them mutually exclusive
+    dp_subparser = argparse.ArgumentParser(add_help=False)
+    dp_upgrade_group = dp_subparser.add_mutually_exclusive_group()
+    dp_upgrade_group.add_argument(
+        "--machine",
+        "-m",
+        action="append",
+        help="Specify machines ids to upgrade.",
```
aieri, line 170:
```suggestion
        help="Specify machine ids to upgrade.",
```

aieri: Also, we don't clarify how to specify them (e.g. comma separated, or via multiple `-m` options).
I know we aren't actually reading those variables yet, so... should we even include these options? It would mislead users into thinking that data plane upgrades are already implemented.

aieri: perhaps we could merge this section of the code in a separate branch for now

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+        description="Run upgrade for the data-plane components.",
+        help="Run upgrade for the data-plane components.",
+        usage="cou plan data-plane [options]",
+        parents=[
+            subcommand_common_opts_parser,
+            dp_subparser,
+            run_args_parser,
+        ],
+        formatter_class=CapitalisedHelpFormatter,
+    )
```
aieri, line 271: we should mention that the upgrade will start only if the control plane has already been upgraded

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+
+    plan_subparser.add_parser(
+        "data-plane",
+        description="Show the steps for upgrading the data-plane components.",
+        help="Show the steps for upgrading the data-plane components.",
+        usage="cou plan data-plane [options]",
+        parents=[subcommand_common_opts_parser, dp_subparser],
+        formatter_class=CapitalisedHelpFormatter,
+    )
```
aieri, line 220: we should indicate that steps will not be shown if the control plane hasn't been upgraded

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+        default=False,
+    )
+    run_parser = subparsers.add_parser(
+        "run",
+        description="Run the cloud upgrade.",
+        help="Run the cloud upgrade.",
+        usage="cou run [options]",
+        parents=[subcommand_common_opts_parser, run_args_parser],
+        formatter_class=CapitalisedHelpFormatter,
     )
```
aieri, line 244: it's not clear here if one can run the whole upgrade at once (we must offer this option), i.e. if control-plane and data-plane subparsers are optional

aieri (changes requested):
I have made several comments and suggestions, but I'm overall ok with the implementation.
I do wonder however if using a third party import like click would have simplified managing the subcommands. It's not easy to have a mental picture of the CLI interface by simply reading the code.

README.md
```diff
@@ -10,20 +10,20 @@ different repository, and only use some of the code and ideas found here.
 ```bash
 rm *.snap
 snapcraft clean
```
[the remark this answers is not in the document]
agileshaw: I agree that it's better to suggest `make clean` here. Changes made in the new patch

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
-        fmt="%(asctime)s [%(name)s] [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    log_formatter_console = logging.Formatter(
-        fmt="%(asctime)s [%(levelname)s] %(message)s", datefmt="%Y-%m-%d %H:%M:%S"
-    )
-    root_logger = logging.getLogger()
-    root_logger.setLevel("DEBUG")
+    if quiet:
+        return "CRITICAL"
+    return VERBOSITY_LEVEL[min(verbosity, 4)]
```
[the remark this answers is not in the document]
agileshaw: I like this approach. Though it's requires more code, the readability and maintainability would be improved.

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+        help="Specify machine hostnames to upgrade.",
+        dest="hostnames",
+    )
+    dp_upgrade_group.add_argument(
+        "--availability-zone",
+        "--az",
+        action="append",
+        help="Specify availability zones to upgrade.",
+        dest="availability_zones",
+    )
```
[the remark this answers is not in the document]
agileshaw: It's not necessary because we are just casting them as strings. But I added it in the new patch for better clarity

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
-        action="help",
-        default=argparse.SUPPRESS,
-        help="Show this help message and exit.",
+    run_args_parser.add_argument(
+        "--i-really-mean-it",
+        help="Run upgrade steps even if they are considered as risky operations (e.g. upgrading "
+        "data-plane nodes with non-empty computes).",
+        action="store_true",
+        default=False,
+    )
```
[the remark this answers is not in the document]
agileshaw: Removed.

setup.cfg
agileshaw: Could be a leftover from the previous version of the repo. Corrected it in the new patch

setup.cfg
agileshaw: Done

cou/cli.py
agileshaw: Update. The newlines are displayed in the intended format now

completions/bash/cou_bash_completion
agileshaw: Done

completions/bash/cou_bash_completion
agileshaw: This was unintended and caused by a change I made after testing. I updated the bash completion and now the correct suggestions are showing.

cou/cli.py
agileshaw: My idea of creating these placeholder options was to construct as much CLI as possible so future developer only needs to read the values without needing to implement extra parsing functionalities. If you are worried about misleading users,  maybe we can let the program exit with "Data plane components upgrade is not yet supported." message when `data-plane` is specified.

cou/cli.py
agileshaw: Expanded the description to be more informative

cou/cli.py
```diff
@@ -12,140 +12,164 @@
...
-COU_DIR_LOG = pathlib.Path(os.getenv("COU_DATA", ""), "log")
 AVAILABLE_OPTIONS = "cas"
 
 logger = logging.getLogger(__name__)
+progress_indicator = Halo(spinner="line", placement="right")
 
 
-def parse_args(args: Any) -> argparse.Namespace:
-    """Parse cli arguments.
+class VerbosityLevel(Enum):
```
rgildein, line 39: Instead of this Enum and `get_log_level`, can we create "action" or specific "type" for `--verbose` argument.? Or simple use simple function in `parse_args` function. e.g.
```python
def verbosity(arg: int) -> int:
    match arg:
        case 0:
            return ERROR
        case 1:
            return logging.WARNING
        case 2:
            return logging.INFO
        case 3:
            return logging.DEBUG
        case _:
            raise argparse.ArgumentTypeError(f"{arg!r} is not valid verbosity")
```

cou/cli.py
```diff
@@ -12,140 +12,164 @@
...
+
+        :param value: value to get enum member
+        :type value: object
+        :return: return a member of VerbosityLevel
+        :rtype: Enum
+        :raises ValueError: Invalid value input.
+        """
+        if isinstance(value, int) and value > 4:
+            return cls.NOTSET
+        raise ValueError(f"{value} is not a valid member of VerbosityLevel.")
```
rgildein, line 68: If I'm correct this can be raised by `get_log_level` in `entrypoint`, and it is not catch properly. We will reach
```python
logger.error("Unexpected error occurred")
logger.exception(exc)
sys.exit(2)
```
What is not correct.

cou/commands.py
```diff
@@ -0,0 +1,302 @@
...
+                    plan_parser.print_help()
+                case "all":
+                    parser.print_help()
+            sys.exit(0)
+
+        return parsed_args
+    except argparse.ArgumentError as exc:
+        print(f"Error parsing arguments: {exc}\n")
+        print("See 'cou help' for more information.")
+        sys.exit(1)
```
rgildein, line 302: To me this should not be part of parse_args function. We are doing CLI logic here + sys.exit.

agileshaw: I agree that this fits best in entrypoint function, but since it need to access subparsers to print help messages, I implemented it on this level. 

A different solution would be let parse_args function, instead of returning parsed args, return both parser and subparser to entrypont, which I have implemented in the new patch. Please let me know if this is better.

cou/cli.py
aieri: do you mean by subclassing `argparse.Action`?

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+    :return: Arguments parser.
+    :rtype: argparse.ArgumentParser
+    :return: Arguments subparsers.
+    :rtype: argparse._SubParsersAction
+    """
+    # Configure top level argparser and its options
+    parser = argparse.ArgumentParser(
+        description="Charmed OpenStack Upgrader (cou) is an application to upgrade a Canonical "
+        "distribution of Charmed OpenStack. The application auto-detects the version of the "
+        "running cloud and will propose an upgrade to the next available version.",
```
aieri, line 324: This line is very long, you may need multiple `\n`s.
```suggestion
        description="Charmed OpenStack Upgrader (cou) is an application to upgrade a Canonical "
        "distribution of Charmed OpenStack.\nThe application auto-detects the version of the "
        "running cloud and will propose an upgrade to the next available version.",
```

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+        usage="cou plan control-plane [options]",
+        parents=[subcommand_common_opts_parser],
+        formatter_class=CapitalizeHelpFormatter,
+    )
+    plan_subparser.add_parser(
+        "data-plane",
+        description="Show the steps for upgrading the data-plane components. This is possible "
+        "only if control-plane has been fully upgrade. Otherwise an error will be thrown.",
+        help="Show the steps for upgrading the data-plane components. This is possible "
+        "only if control-plane has been fully upgrade. Otherwise an error will be thrown.",
```
aieri, line 206: these lines are very long and would benefit from a `\n`

aieri, line 206:
```suggestion
        description="Show the steps for upgrading the data-plane components. This is possible "
        "only if control-plane has been fully upgraded, otherwise an error will be thrown.",
        help="Show the steps for upgrading the data-plane components. This is possible "
        "only if control-plane has been fully upgraded, otherwise an error will be thrown.",
```

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+    dp_upgrade_group.add_argument(
+        "--availability-zone",
+        "--az",
+        action="append",
+        help="Specify availability zone(s) to upgrade. This option accepts a single availability "
+        "zone as well as a stringified comma-separated list for multiple AZs. This option can be "
+        "used repetitively.",
+        dest="availability_zones",
+        type=str,
```
aieri, line 157: these lines are too long, please add newlines

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+    # Define options specific to data-plane child command and make them mutually exclusive
+    dp_subparser = argparse.ArgumentParser(add_help=False)
+    dp_upgrade_group = dp_subparser.add_mutually_exclusive_group()
+    dp_upgrade_group.add_argument(
+        "--machine",
+        "-m",
+        action="append",
+        help="Specify machine id(s) to upgrade. This option accepts a single machine id as well "
+        "as a stringified comma-separated list for multiple machines. This option can be used "
+        "repetitively.",
```
aieri, line 135:
```suggestion
        help="Specify machine id(s) to upgrade. This option accepts a single machine id as well "
        "as a stringified comma-separated list of ids, and can be repeated multiple times.",
```

aieri: and so on for `--hostname` and `--availability-zone`

cou/cli.py
aieri: Yes, given the desire to do a release for the control-plane part I think we should not have the data-plane option at all.
I do agree that starting to design this is useful, hence the idea of merging these lines elsewhere (instead of throwing them away)

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+        usage="cou plan control-plane [options]",
+        parents=[subcommand_common_opts_parser, run_args_parser],
+        formatter_class=CapitalizeHelpFormatter,
+    )
+    run_subparser.add_parser(
+        "data-plane",
+        description="Run upgrade for the data-plane components. This is possible only if "
+        "control-plane has been fully upgrade. Otherwise an error will be thrown.",
+        help="Run upgrade for the data-plane components. This is possible only if "
+        "control-plane has been fully upgrade. Otherwise an error will be thrown.",
```
aieri, line 264:
```suggestion
        description="Run upgrade for the data-plane components. This is possible only if "
        "control-plane has been fully upgraded. Otherwise an error will be thrown.",
        help="Run upgrade for the data-plane components. This is possible only if "
        "control-plane has been fully upgraded. Otherwise an error will be thrown.",
```

cou/commands.py
```diff
@@ -0,0 +1,343 @@
...
+    run_args_parser.add_argument(
+        "--interactive",
+        help="Run upgrade with prompt.",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+    )
+    run_parser = subparsers.add_parser(
+        "run",
+        description="Run the cloud upgrade. If upgrade-group is unspecified, perform "
+        "upgrade for the entire cloud",
```
aieri, line 238:
```suggestion
        description="Run the cloud upgrade. If upgrade-group is unspecified, upgrade the whole cloud",
```

cou/cli.py
aieri: In theory, yes, `VerbosityLevel(-1)` would cause an exception, but as a user you can only instantiate `VerbosityLevel(0)` and up.

aieri (changes requested):

cou/cli.py
agileshaw: Yes. If the exception is ever raise, that means something unexpected happened and I think `exit(2)` is appropriate for it

cou/cli.py
agileshaw: I'm not against merging it to a separate branch. But if we do, I believe that we should push everything else "data-plane"-related (e.g. PR #54 ) to that branch also so the main branch is clean with just control-plane code.

cou/cli.py
agileshaw: I'm not quite sure what the advantages are with this approach comparing to using Enum. Could you elaborate more?

cou/cli.py
```diff
@@ -42,110 +71,333 @@ def parse_args(args: Any) -> argparse.Namespace:
...
+        await apply_plan(upgrade_plan, interactive)
+    print("Upgrade completed.")
 
 
 async def entrypoint() -> None:
-    """Execute 'charmed-openstack-upgrade' command."""
+    """Execute 'canonical-openstack-upgrade' command."""
     try:
         args = parse_args(sys.argv[1:])
+        # disable progress indicator when in quite mode to suppress its console output
```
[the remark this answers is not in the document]
agileshaw: Thanks! This is a typo that spell checker couldn't catch :)

cou/cli.py
agileshaw: For future reference, based on discussion elsewhere, we agreed to hide upgrade-group options until data-plane upgrade functionality is implemented.

cou/cli.py
rgildein: It can be more integrated with parsing arguments and CLI will failed directly in the parsing phase and not after it + it will raise and proper exception instead of ValueError which is catch by our try except [here](https://github.com/canonical/charmed-openstack-upgrader/blob/main/cou/cli.py).

rgildein: Now I reading next comment and realized that the `ValueError` is only raised if value is not an integer what from CLI call will never happened. I did not realized that.
It's more useful to catch argument error in `parse_args` function and not in `setup_logging(log_level=get_log_level(quiet=args.quiet, verbosity=args.verbosity))` call, but that argument error should never happen., so it's not important. It is only more understandable to be if it's integrated in parsing arguments, phase instead of in later phase of CLI.
Marking as resolved.

cou/cli.py
rgildein: This `ValueError` confused me, but you are right it could not happen from any CLI call.

rgildein (changes requested):
Only one issue with printing to STDOUT instead STDERR, otherwise LGTM

cou/commands.py
aieri: you've missed performing this change on `--hostname` and `--availability-zone`

aieri (changes requested):
I have resolved all outdated conversations of mine; there's one left for you to look at, as well as Robert's suggestions. Once those are done we can merge.

cou/commands.py
agileshaw: Thanks for the suggestion. I have moved the parsing exception handling back to `parse_args` function and uses `parser.error`/'parser.exit' instead.

completions/bash/cou_bash_completion
```diff
@@ -40,25 +46,28 @@ _cou_bash_completion() {
     cmd="${COMP_WORDS[1]}"
     _subcommand_args ${cmd}
 
-    if [ $COMP_CWORD -eq 2 ]; then
-        COMPREPLY=( $(compgen -W "${child_commands} ${opts}" -- "${cur}") )
-        return 0
-    fi
+    # TODO(txiao): Remove the line below and un-comment the rest after data-plane upgrade
```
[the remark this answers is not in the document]
agileshaw: Removed TODOs in bash completion and created an issue (#63) to track

tests/unit/test_cli.py
```diff
@@ -12,100 +12,191 @@
...
+    with patch(
+        "cou.cli.juju_utils.get_current_model_name",
+        new=AsyncMock(return_value=model_name),
+    ) as mock_get_current_model_name, patch(
+        "cou.cli.generate_plan", new=AsyncMock()
+    ) as mock_generate_plan, patch(
+        "cou.cli.Analysis.create",
+        new=AsyncMock(return_value=analysis_result),
+    ) as mock_analyze:
+        await cli.analyze_and_plan(model_name)
```
rgildein, line 80: not-blocking: I think it's more readable to use `mocker` or `patch` as decorator.

cou/commands.py
agileshaw: Done :)

rgildein (approved):

tests/unit/test_cli.py
agileshaw: Replaced complex context manager with decorators

esunar (approved):
LGTM

rgildein (approved):
LGTM

PR #50, opened 2023-09-07
Introduce COUModel with waiting and auto-reconnection

rgildein:
- replace JujuWaiter with COUModel

PR #53, opened 2023-09-11
define COUModel and our retry decorator

rgildein:
The COUModel object will be used and passed everywhere in COU. The retry decorator is mean to be used only for COUModel object functions like connect, get_status, etc.

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -22,6 +22,7 @@ from typing import Any, Callable, Dict, Optional
 from juju.action import Action
 from juju.application import Application
 from juju.client._definitions import FullStatus
+from juju.client.connector import NoConnectionException
 from juju.errors import JujuAgentError, JujuAppError, JujuMachineError, JujuUnitError
 from juju.model import Model
 from juju.unit import Unit
@@ -538,12 +539,12 @@ def retry(
 
     def _wrapper(func: Callable) -> Callable:
         @wraps(func)
-        async def wrapper(self: "COUModel", *args: Any, **kwargs: Any) -> Any:
+        async def wrapper(*args: Any, **kwargs: Any) -> Any:
             attempt: int = 0
             start_time = datetime.now()
             while (datetime.now() - start_time).seconds <= timeout:
                 try:
-                    return await func(self, *args, **kwargs)
+                    return await func(*args, **kwargs)
                 except ignored_exceptions:
                     # raising exception if no_retry_exception happen or TimeoutException
                     raise
@@ -577,9 +578,13 @@ class COUModel:
         self.logger = logging.getLogger(name)
 
     @property
-    def disconnected(self) -> bool:
+    def connected(self) -> bool:
         """Check if model is connected."""
-        return not (self._model.is_connected() and self._model.connection().is_open)
+        try:
+            connection = self._model.connection()
+            return connection is not None and connection.is_open
+        except NoConnectionException:
+            return False
 
     @property
     def name(self) -> str:
@@ -608,7 +613,7 @@ class COUModel:
     @retry(no_retry_exception=(BakeryException,))
     async def _get_model(self) -> Model:
         """Get juju.model.Model and make sure that's it's connected."""
-        if not self.disconnected:
+        if not self.connected:
             await self._connect()
 
         return self._model
@@ -627,6 +632,7 @@ class COUModel:
 
         return unit
 
+    @retry(no_retry_exception=(ApplicationNotFound,))
     async def get_application_config(self, name: str) -> Dict:
         """Return application configuration.
 
@@ -727,13 +733,10 @@ class COUModel:
         :param model_name: Name of model to query.
         :type model_name: Optional[str]
         """
-        model = await self._get_model()
-        app = model.applications.get(name)
-        if app is None:
-            raise ApplicationNotFound(f"application {name} was not found in model {self.name}")
-
+        app = await self._get_application(name)
         await app.set_config(configuration)
 
+    @retry(no_retry_exception=(UnitNotFound,))
     async def scp_from_unit(
         self,
         unit_name: str,
@@ -807,4 +810,14 @@ class COUModel:
             resources=resources,
             revision=revision,
             switch=switch,
+        )
+
+    async def wait_for_idle(self, timeout: int, apps: Optional[list[str]] = None) -> None:
+        """Wait for model to rich idle state."""
+        model = await self._get_model()
+        wait = retry(model.wait_for_idle, timeout=timeout)
+        await wait(
+            apps=apps,
+            idle_period=JujuWaiter.MODEL_IDLE_PERIOD,
+            timeout=JujuWaiter.JUJU_IDLE_TIMEOUT,
         )
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -40,6 +40,8 @@ JUJU_MAX_FRAME_SIZE: int = 2**30
 DEFAULT_TIMEOUT: int = 60
 DEFAULT_MAX_WAIT: int = 5
 DEFAULT_WAIT: float = 1.1
+DEFAULT_MODEL_RETRIES: int = 30
+DEFAULT_MODEL_RETRY_BACKOFF = 2
 
 CURRENT_MODEL: Optional[Model] = None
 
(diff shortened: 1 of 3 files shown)


rgildein:
alternative solution for #50

cou/utils/juju_utils.py
```diff
@@ -515,3 +522,110 @@ def _check_time(self) -> None:
...
+        async def wrapper(self: "COUModel", *args: Any, **kwargs: Any) -> Any:
+            start_time = datetime.now()
+            while (datetime.now() - start_time).seconds <= timeout:
+                try:
+                    return await func(self, *args, **kwargs)
+                except ignored_exceptions:
+                    # raising exception if no_retry_exception happen or TimeoutException
+                    raise
+                except Exception:
+                    logger.debug("function %s failed", func.__name__, exc_info=True)
```
esunar, line 556: We should add some wait here. If we fail too quick this may cause problems(too much hanging connections, max number of open files etc.)

cou/utils/juju_utils.py
```diff
@@ -509,3 +516,295 @@ def _check_time(self) -> None:
...
+    :param function: function to wrapped
+    :type function: Optional[Callable]
+    :param timeout: timeout in seconds
+    :type timeout: int
+    :param no_retry_exception: tuple of exception on which function will not be retried
+    :type no_retry_exception: Optional[tuple]
+    :return: wrapped function
+    :rtype: Callable
+    """
+    ignored_exceptions = (TimeoutException, *(no_retry_exception or ()))
```
esunar, line 537: I did not understand why we raising timeout exception. Assume we lost our network connection. It will create TimeoutException and we will not be able to retry. We should only raise TimeoutException from line 556.

rgildein: The connection lost will not create `TimeoutException` maybe `TimeoutError`, the `TimeoutException` is our own exception raised only by this retry decorator. Why it's needed is due nested decorating. For example we are using function `_get_model` in multiple other functions and if we decide to decorate any of it (as I did for `set_application_config`) you do not want to retry if function `_get_model` failed with `TimeoutException`.

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
             raise TimeoutException(
                 f"Model {self.model_name} has not stabilized after {self.timeout} seconds."
             )
+
+
+def retry(
+    function: Optional[Callable] = None,
+    timeout: int = DEFAULT_TIMEOUT,
+    no_retry_exception: Optional[tuple] = None,
```
aieri, line 528: and should we maybe rename it to a plural `no_retry_exceptions` since it accepts a tuple?

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
...
+        """
+        model = await self._get_model()
+        app = model.applications.get(name)
+        if app is None:
+            raise ApplicationNotFound(f"Application {name} was not found in model {model.name}.")
+
+        return app
+
+    async def _get_model(self) -> Model:
+        """Get juju.model.Model and make sure that's it's connected."""
```
aieri, line 621:
```suggestion
        """Get juju.model.Model and make sure that it is connected."""
```

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
...
+            force_series=force_series,
+            force_units=force_units,
+            path=path,
+            resources=resources,
+            revision=revision,
+            switch=switch,
+        )
+
+    async def wait_for_idle(self, timeout: int, apps: Optional[list[str]] = None) -> None:
+        """Wait for model to rich idle state."""
```
aieri, line 819:
```suggestion
        """Wait for model to reach an idle state."""
```

tests/unit/utils/test_juju_utils.py
```diff
@@ -662,6 +672,347 @@ async def test_extract_charm_name_not_existing_app(mocked_get_model):
...
+            obj.run()
+
+    test_model = TestModel()
+    await test_model.func()
+    obj.run.assert_called_once_with()
+
+
+@pytest.mark.asyncio
+async def test_retry_with_args():
+    """Tets retry as decorator with arguments."""
```
aieri, line 697:
```suggestion
    """Test retry as decorator with arguments."""
```

tests/unit/utils/test_juju_utils.py
```diff
@@ -662,6 +672,347 @@ async def test_extract_charm_name_not_existing_app(mocked_get_model):
...
+
+    test_model = TestModel()
+    await test_model.func()
+    obj.run.assert_called_once_with()
+
+
+@pytest.mark.asyncio
+@patch("asyncio.sleep", new=AsyncMock())
+async def test_retry_with_failures():
+    """Tets retry with some failures."""
```
aieri, line 713:
```suggestion
    """Test retry with some failures."""
```

aieri (changes requested):
A few minor comments, otherwise +1

cou/utils/juju_utils.py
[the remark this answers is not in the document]
rgildein: I did not realize that tuple is immutable and I can use it like default value. It's just a habit to always use None.
Thanks

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
...
+
+
+def retry(
+    function: Optional[Callable] = None,
+    timeout: int = DEFAULT_TIMEOUT,
+    no_retry_exception: Optional[tuple] = None,
+) -> Callable:
+    """Retry function for usage in COUModel.
+
+    :param function: function to wrapped
```
agileshaw, line 532:
```suggestion
    :param function: function to be wrapped
```

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
...
+        """Get juju.model.Model and make sure that's it's connected."""
+        if not self.connected:
+            await self._connect()
+
+        return self._model
+
+    async def _get_unit(self, name: str) -> Unit:
+        """Get juju.unit.unit from model.
+
+        :return: Unit
```
agileshaw, line 630: Need to add docstring for `name` parameter

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
             raise TimeoutException(
                 f"Model {self.model_name} has not stabilized after {self.timeout} seconds."
             )
+
+
+def retry(
+    function: Optional[Callable] = None,
```
agileshaw, line 526: I was wondering why do we need to accept `None` as a possible value for the `function` parameter?

cou/utils/juju_utils.py
```diff
@@ -509,3 +520,304 @@ def _check_time(self) -> None:
...
+
+    async def set_application_config(self, name: str, configuration: Dict[str, str]) -> None:
+        """Set application configuration.
+
+        :param name: Name of application
+        :type name: str
+        :param configuration: Dictionary of configuration setting(s)
+        :type configuration: Dict[str,str]
+        :param model_name: Name of model to query.
+        :type model_name: Optional[str]
```
agileshaw, line 738: Need to remove docstring about `model_name`

agileshaw (changes requested):

cou/utils/juju_utils.py
rgildein: Because if you want to use decorator without any configuration or with some additional configuration. 
example:
```python
>>> def decorator(func, debug=False):
...  def wrapper(*args, **kwargs):
...   print("decorated function", func.__name__, debug)
...   return func(*args, **kwargs)
...  return wrapper
... 
>>> @decorator
... def test1():
...  print("test1")
... 
>>> @decorator(debug=True)
... def test2():
...  print("test2")
... 
  File "<stdin>", line 1, in <module>
TypeError: decorator() missing 1 required positional argument: 'func'
>>> test1()
decorated function test1 False
test1
```

agileshaw: I see, thanks for the explanation.

esunar (dismissed):

esunar (approved):

agileshaw (approved):
LGTM

aieri (approved):
+1

PR #54, opened 2023-09-12
Added data plane apps.

esunar:
Unit test are added


cou/steps/analyze.py
```diff
@@ -100,7 +127,12 @@ def __str__(self) -> str:
         :return: String representation of Application objects.
         :rtype: str
         """
-        return os.linesep.join([str(app) for app in self.apps])
+        return (
+            "Control Plane:\n"
+            + "\n".join([str(app) for app in self.apps_control_plane])
+            + "Data Plane:\n"
+            + "\n".join([str(app) for app in self.apps_data_plane])
+        )
```
aieri, line 135: [non-blocking] this starts being a bit too complex for a single return. Perhaps it'd be better to construct the app lists in two variables and return a multiline f-string here.

cou/steps/analyze.py
```diff
@@ -111,4 +143,9 @@ def current_cloud_os_release(self) -> Optional[OpenStackRelease]:
         :return: OpenStack release codename
         :rtype: OpenStackRelease
         """
-        return min((app.current_os_release for app in self.apps), default=None)
+        control_plane = (app for app in self.apps_control_plane)
+        all_apps = chain(control_plane, self.apps_data_plane)
```
aieri, line 147: I don't understand why you need the generator expression and chain here. Can `apps_control_plane` and `apps_data_plane` be anything else than lists?

cou/utils/openstack.py
```diff
@@ -39,6 +39,8 @@
     "mysql": ["mysql-innodb-cluster", "mysql-router"],
 }
 
+DATA_PLANE_CHARMS = ["nova-compute", "ceph-osd", "swift-proxy", "swift-storage"]
```
aieri, line 42: just a note: I don't remember how we upgrade swift-proxy, I suggest getting in touch with Gui. Swift-proxy is arguably a data plane app, but its unavailability does not cause loss of connectivity for VMs, which is usually what we are worried about

aieri (changes requested):
+1 overall, except that you're not really unit testing the new split function. Perhaps you could slightly alter `test_analysis_create` instead of directly testing a piece of code that feels right as a private function.

cou/steps/analyze.py
esunar: Currently it seem more expressive to me. If it even gets complicated we can split it.

cou/steps/analyze.py
esunar: they are `Iterable`
Would you prefer something like this:
```
        return min(
            (app.current_os_release for app in [*self.apps_control_plane, *self.apps_data_plane]),
            default=None,
        )
```

esunar:
> +1 overall, except that you're not really unit testing the new split function. Perhaps you could slightly alter `test_analysis_create` instead of directly testing a piece of code that feels right as a private function.

unit tests are not complete, I also need to add some data plane applications to show it actually works.

cou/utils/openstack.py
esunar: After checking with Giuseppe I've removed "swift-proxy", "swift-storage" from data-plane.

cou/steps/analyze.py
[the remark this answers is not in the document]
rgildein: I agree that Gabriel solution is nicer, I did not understand purpose of `control_plane = (app for app in self.apps_control_plane)`, what is different than using `self.apps_control_plane`?

Since both apps_control_plane and apps_data_plane are defined as `Iterable[OpenStackApplication]`, can we use set instead of list?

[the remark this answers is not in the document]
gabrielcocenza: using set for `apps_control_plane` and `apps_data_plane` can mess with the upgrade order. I think they should stay as list

cou/steps/analyze.py
```diff
@@ -51,7 +54,31 @@ async def create(cls, model_name: Optional[str] = None) -> Analysis:
...
+
+        :param apps: List of applications to split.
+        :type apps:  Iterable[OpenStackApplication]
+        :return: Control plane and data plane application lists.
+        :rtype: tuple[list[OpenStackApplication], list[OpenStackApplication]]
+        """
+        data_plane = []
+        control_plane = []
+        for app in apps:
+            if app.charm in DATA_PLANE_CHARMS:
```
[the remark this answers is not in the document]
esunar: Yes when the machines are available they will be further processed.

cou/steps/plan.py
```diff
@@ -83,7 +83,7 @@ async def create_upgrade_group(
     :rtype: UpgradeStep
     """
     group_upgrade_plan = UpgradeStep(description=description, parallel=False, function=None)
-    for app in filter(filter_function, analysis_result.apps):
+    for app in filter(filter_function, analysis_result.apps_control_plane):
```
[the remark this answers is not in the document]
esunar: I think the data plane upgrade should be totally different from control plane. What I think is:
```
- top level plan
  - control plane
    - principals
      - ...
      - ...
      - ...
    - subordinates
      - ...
      - ...
      - ...
  - data plane
    - principals
      - machine 1
        - nova
        - ceph
      - machine 2
        - nova
        - cehp
    - subordinates
      - app1
      - app2
```

cou/steps/analyze.py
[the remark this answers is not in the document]
esunar: Generator and chaining is more effective since it does not create new data structure. But I am ok with any other solution.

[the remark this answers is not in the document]
agileshaw: Why do we care about the order here? Don't we just need to find a minimal openstack release out of `all_apps`?

rgildein (dismissed):
LGTM

tests/unit/conftest.py
```diff
@@ -109,6 +109,21 @@ def status():
         ]
     )
 
+    mock_nova_wallaby = mock.MagicMock()
+    mock_nova_wallaby.series = "focal"
+    mock_nova_wallaby.charm_channel = "wallaby/stable"
+    mock_nova_wallaby.charm = "ch:amd64/focal/keystone-638"
```
gabrielcocenza, line 115:
```suggestion
    mock_nova_wallaby.charm = "ch:amd64/focal/nova-compute-638"
```

tests/unit/conftest.py
```diff
@@ -109,6 +109,21 @@ def status():
         ]
     )
 
+    mock_nova_wallaby = mock.MagicMock()
```
rgildein, line 112: I think that we should use `spec` argument so it's obvious what we are trying to mock + it's safer if you used something what is not defined in the object.

rgildein (changes requested):
One change request, but it's looking good to me.

tests/unit/conftest.py
esunar: done

tests/unit/conftest.py
```diff
@@ -16,19 +16,21 @@
...
 
 from cou.apps.app import OpenStackApplication
+from cou.apps.auxiliary import OpenStackAuxiliaryApplication
 from cou.apps.subordinate import OpenStackSubordinateApplication
 
 
 @pytest.fixture
 def status():
-    mock_keystone_ussuri = mock.MagicMock()
+    mock_keystone_ussuri = mock.MagicMock(spec=ApplicationStatus)
```
gabrielcocenza, line 28: I didn't know about this feature. Very cool @rgildein

I think we should use `spec_set` instead of `spec` based in this [discussion](https://stackoverflow.com/questions/25323361/what-is-spec-and-spec-set)

`spec` still allows us to set attribute that is not specified, while with spec_set, it is not allowed to set unspecified attribute.

gabrielcocenza (changes requested):
I think we should change `spec` to `spec_set`

tests/unit/conftest.py
esunar: done

gabrielcocenza (approved):

aieri (approved):
lgtm, but please rebase, squash, and possibly extend the commit message

Issue #56, opened 2023-09-13
Dynamic versioning with git tag


agileshaw:
The version of the snap is statically set to '0.1' at the moment. We need to add dynamic versioning feature which uses git tag to determine its value.

PR #57, opened 2023-09-13
Set dynamic version for snap

agileshaw:
Python package tracks git tag for versioning. Passing this info to snap so it can be on the same version.

closes: #56


rgildein (approved):
LGTM

gabrielcocenza (approved):
LGTM

PR #59, opened 2023-09-14
Droppedn unnecesary mock requirements, using unittests.mock instead

rgildein:
There is no need to use `mock` package, since it can be replaced by `unitests.mock`.

### DIFF
--- a/tests/unit/utils/test_juju_utils.py
+++ b/tests/unit/utils/test_juju_utils.py
@@ -13,14 +13,13 @@
 # limitations under the License.
 import asyncio
 import os
+from unittest import mock
 from unittest.mock import MagicMock
 
 import aiounittest
-import mock
 import pytest
 from juju.errors import JujuUnitError
 from juju.model import Model
-from mock.mock import AsyncMock
 
 import cou.utils.juju_utils as utils
 from cou.exceptions import ApplicationNotFound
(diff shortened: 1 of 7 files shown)


esunar (approved):

esunar (approved):

tests/unit/utils/test_juju_utils.py
```diff
@@ -351,9 +351,9 @@ async def test_get_current_model_from_juju(self):
 
     async def test_get_full_juju_status(self):
         with mock.patch("cou.utils.juju_utils._get_model") as get_model:
-            mymodel = AsyncMock()
+            mymodel = mock.AsyncMock()
```
agileshaw, line 354: AsyncMock is already imported, why do we want to add `mock` import and use `mock.AsyncMock`?

rgildein: Not sure why, replacing it. Thanks

esunar (approved):

tests/unit/utils/test_juju_utils.py
```diff
@@ -13,10 +13,10 @@
 # limitations under the License.
 import asyncio
 import os
+from unittest import mock
 from unittest.mock import AsyncMock, MagicMock, patch
```
agileshaw, line 17: `mock` module itself is only used as `mock.patch`. Since we are already importing `patch` from `mock` in the second line, can we replace `mock.patch` with `patch` throughout the test code so we don't need to import `mock`?

agileshaw (changes requested):

tests/unit/utils/test_juju_utils.py
agileshaw: Ah I see. I wonder why it makes a difference here with different type of imports. After all they are referring to the same function. 

In that case, can we do it the other way around: replace the usage of `patch` with `mock.patch` and remove the importing of `patch`? I just want to keep its usage consistent

rgildein: I do not think that it's needed since I will drop all these tests in my next PR.

aieri (approved):

agileshaw (approved):

PR #60, opened 2023-09-14
Added auxiliary subordinate charms

gabrielcocenza:
This PR should be merged after #42


tests/unit/apps/test_subordinate_auxiliary.py
```diff
@@ -0,0 +1,78 @@
...
+#  limitations under the License.
+"""Tests of the Auxiliary Subordinate application class."""
+
+from cou.apps.subordinate_auxiliary import OpenStackAuxiliarySubordinateApplication
+from cou.steps import UpgradeStep
+from cou.utils.openstack import OpenStackRelease
+from tests.unit.apps.utils import assert_plan_description
+
+
+def test_subordinate_auxiliary(status):
```
agileshaw, line 22: Since we are using `auxiliary subordinate` everywhere else, I think we should keep it consistent in the tests.
```suggestion
def test_auxiliary_subordinate(status):
```

agileshaw (changes requested):
Overall LGTM. One small thing is that, since we are mainly using the name `auxiliary subordinate`, I suggestion keeping it consistent throughout, including the title of the PR.

tests/unit/apps/test_subordinate_auxiliary.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    assert app.channel == "8.0/stable"
+    assert app.charm_origin == "ch"
+    assert app.os_origin == ""
+    assert app.apt_source_codename is None
+    assert app.channel_codename == "yoga"
+    assert app.current_os_release == "yoga"
+
+
+def test_subordinate_auxiliary_upgrade_plan_to_victoria(status):
```
agileshaw, line 34: This function name also needs to be updated

tests/unit/apps/test_subordinate_auxiliary.py
```diff
@@ -0,0 +1,78 @@
...
+    plan = app.generate_upgrade_plan(target)
+
+    steps_description = [
+        f"Refresh '{app.name}' to the latest revision of '8.0/stable'",
+    ]
+
+    assert_plan_description(plan, steps_description)
+
+
+def test_subordinate_auxiliary_upgrade_charm(status, mocker):
```
agileshaw, line 49: This function name also needs to be updated

agileshaw (dismissed):

cou/apps/auxiliary_subordinate.py
```diff
@@ -0,0 +1,60 @@
...
+        """Infer the OpenStack release from subordinate charm's channel.
+
+        We cannot determine the OpenStack release base on workload packages because the principal
+        charm has already upgraded the packages.
+        :return: OpenStackRelease object.
+        :rtype: OpenStackRelease
+        """
+        return self.channel_codename
+
+    def generate_upgrade_plan(self, target: str) -> UpgradeStep:
```
aieri, line 37: this function is almost identical to `OpenStackSubordinateApplication.generate_upgrade_plan`, I wonder if there's a way to use it without having to duplicate code here

tests/unit/apps/test_auxiliary_subordinate.py
```diff
@@ -0,0 +1,78 @@
...
+"""Tests of the Auxiliary Subordinate application class."""
+
+from cou.apps.auxiliary_subordinate import OpenStackAuxiliarySubordinateApplication
+from cou.steps import UpgradeStep
+from cou.utils.openstack import OpenStackRelease
+from tests.unit.apps.utils import assert_plan_description
+
+
+def test_auxiliary_subordinate(status):
+    app = OpenStackAuxiliarySubordinateApplication(
```
aieri, line 23: you repeat this line several times. Could it be useful to turn it into a fixture?

aieri (approved):
LGTM, just a couple of questions

esunar (dismissed):

agileshaw (changes requested):

esunar (approved):

cou/apps/auxiliary_subordinate.py
```diff
@@ -0,0 +1,40 @@
...
+from cou.apps.subordinate import OpenStackSubordinateApplication
+from cou.utils.openstack import AUXILIARY_SUBORDINATES, OpenStackRelease
+
+
+@AppFactory.register_application(AUXILIARY_SUBORDINATES)
+class OpenStackAuxiliarySubordinateApplication(OpenStackAuxiliaryApplication):
+    """Auxiliary subordinate application class."""
+
+    _default_used = False
+    generate_upgrade_plan: Callable = OpenStackSubordinateApplication.generate_upgrade_plan
```
rgildein, line 29: I think we need to find other way to do this. For now it's ok to be merged, but we should report an issue to found new way we can use better inheriting, mixing or using separate functions.

tests/unit/apps/test_auxiliary_subordinate.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    steps_description = [
+        f"Refresh '{app.name}' to the latest revision of '8.0/stable'",
+        (
+            f"Upgrade '{app.name}' to the new channel: "
+            f"'{app.target_channel(OpenStackRelease(target))}'"
+        ),
+    ]
+    plan = app.generate_upgrade_plan(target)
+    assert_plan_description(plan, steps_description)
```
rgildein, line 71: not-blocking: I do not see point to doing this here, since this function is defined in different class and it's tested there.

rgildein (dismissed):
Can you please report an issue mention in comment for `generate_upgrade_plan`. Thanks

aieri (dismissed):
+1, just please file the issue about possibly using mixins that Robert mentioned

tests/unit/apps/test_auxiliary_subordinate.py
gabrielcocenza: Before the class had it's own method for `generate_upgrade_plan`, but now we can indeed remove. Thanks

cou/apps/auxiliary_subordinate.py
gabrielcocenza: Issue #64 created

esunar (approved):

agileshaw (approved):
LGTM

PR #62, opened 2023-09-18
Added refresh option for subordinate charms.

### LINKED ISSUE #61: Subordinates are not able to make ch migration
Currently, the logic implemented on `OpenStackSubordinateApplication` with `_default_used` and the channel setter doesn't allow the application to make the migration if necessary.

esunar:
resolves #61

### DIFF
--- a/tests/unit/apps/test_subordinate.py
+++ b/tests/unit/apps/test_subordinate.py
@@ -77,7 +77,6 @@ def test_channel_setter_valid(status, channel):
 
     app.channel = channel
     assert app.channel == channel
-    assert not app._default_used
 
 
 @pytest.mark.parametrize(
@@ -97,13 +96,3 @@ def test_channel_setter_invalid(status, channel):
 
     app.channel = channel
     assert app.channel == "ussuri/stable"
-    assert app._default_used
-
-
-def test_channel_default_used(status):
-    app_status = status["keystone-ldap"]
-    app = OpenStackSubordinateApplication(
-        "my_keystone_ldap", app_status, {}, "my_model", "keystone-ldap"
-    )
-
-    assert not app._default_used
--- a/tests/unit/apps/test_subordinate.py
+++ b/tests/unit/apps/test_subordinate.py
@@ -96,3 +96,56 @@ def test_channel_setter_invalid(status, channel):
 
     app.channel = channel
     assert app.channel == "ussuri/stable"
+
+
+@pytest.mark.parametrize(
+    "channel",
+    [
+        "focal/edge",
+        "latest/edge",
+        "latest/stable",
+        "something/stable",
+    ],
+)
+def test_generate_plan(status, channel):
+    app_status = status["keystone-ldap-cs"]
+    app = OpenStackSubordinateApplication(
+        "my_keystone_ldap", app_status, {}, "my_model", "keystone-ldap"
+    )
+
+    app.channel = channel
+    plan = app.generate_upgrade_plan("wallaby")
+    assert str(plan) == (
+        "Upgrade plan for 'my_keystone_ldap' to wallaby\n"
+        "\tMigration of 'my_keystone_ldap' from charmstore to charmhub\n"
+        "\tUpgrade 'my_keystone_ldap' to the new channel: 'wallaby/stable'\n"
+    )
+
+
+def test_generate_plan_stable1(status):
+    app_status = status["keystone-ldap-cs"]
+    app = OpenStackSubordinateApplication(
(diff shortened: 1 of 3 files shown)


rgildein (approved):
Code looks good to me, but I  do not understand how this solved the issue or what is actual issue. So more reviews are needed.

gabrielcocenza (changes requested):
It seems that this solves the issue, but please add a unit test simulating the ch migration for subordinates apps.

esunar:
> It seems that this solves the issue, but please add a unit test simulating the ch migration for subordinates apps.

done

tests/unit/apps/test_subordinate.py
```diff
@@ -96,3 +96,56 @@ def test_channel_setter_invalid(status, channel):
 
     app.channel = channel
     assert app.channel == "ussuri/stable"
+
+
+@pytest.mark.parametrize(
+    "channel",
+    [
+        "focal/edge",
```
gabrielcocenza, line 104: I think those channels are a little bit misleading. IIRC, when charms comes from cs it always will be set to `stable`. E.g:

```yaml
applications:
  ubuntu:
    charm: cs:ubuntu-24
    series: bionic
    os: ubuntu
    charm-origin: charmstore
    charm-name: ubuntu
    charm-rev: 24
    charm-channel: stable
    exposed: false
```

So there is no need to have a parameterize function with different channels.

tests/unit/apps/test_subordinate.py
```diff
@@ -96,3 +96,56 @@ def test_channel_setter_invalid(status, channel):
...
+        "latest/edge",
+        "latest/stable",
+        "something/stable",
+    ],
+)
+def test_generate_plan(status, channel):
```
gabrielcocenza, line 110: `test_generate_plan_ch_migration`

tests/unit/apps/test_subordinate.py
```diff
@@ -96,3 +96,56 @@ def test_channel_setter_invalid(status, channel):
...
+    app.channel = channel
+    plan = app.generate_upgrade_plan("wallaby")
+    assert str(plan) == (
+        "Upgrade plan for 'my_keystone_ldap' to wallaby\n"
+        "\tMigration of 'my_keystone_ldap' from charmstore to charmhub\n"
+        "\tUpgrade 'my_keystone_ldap' to the new channel: 'wallaby/stable'\n"
+    )
+
+
+def test_generate_plan_stable1(status):
```
gabrielcocenza, line 125: I think we can remove this test once the suggestions above are done

tests/unit/apps/test_subordinate.py
```diff
@@ -96,3 +96,56 @@ def test_channel_setter_invalid(status, channel):
...
+    app.channel = "ussuri/stable"
+    plan = app.generate_upgrade_plan("wallaby")
+    assert str(plan) == (
+        "Upgrade plan for 'my_keystone_ldap' to wallaby\n"
+        "\tMigration of 'my_keystone_ldap' from charmstore to charmhub\n"
+        "\tUpgrade 'my_keystone_ldap' to the new channel: 'wallaby/stable'\n"
+    )
+
+
+def test_generate_plan_stable2(status):
```
gabrielcocenza, line 140: rename this test to something like `test_generate_plan_victoria_to_wallaby`

tests/unit/apps/test_subordinate.py
gabrielcocenza: I saw that conftest is alreasy setting the channel right as I said here

gabrielcocenza (changes requested):

tests/unit/apps/test_subordinate.py
esunar: made some improvements

tests/unit/apps/test_subordinate.py
esunar: done

tests/unit/apps/test_subordinate.py
esunar: removed

tests/unit/apps/test_subordinate.py
esunar: added from_to list for plan generation

gabrielcocenza (approved):
LGTM. Non-blocker:I think with the new tests added we can remove [test_generate_upgrade_plan](https://github.com/canonical/charmed-openstack-upgrader/blob/main/tests/unit/apps/test_subordinate.py#L43)

aieri (approved):
Looks good, but please ensure the commit message explains why this is necessary

PR #65, opened 2023-09-18
Added machines to units in application.

esunar:
- OpenStackApplication.units converted to list(ApplicationUnit).
- If control plane application exists on data plane it moved to data plane

should be merged after #51


cou/apps/app.py
```diff
@@ -118,6 +117,16 @@ def decorator(application: type[OpenStackApplication]) -> type[OpenStackApplicat
         return decorator
 
 
+@dataclass
+class ApplicationUnit:
+    """Representation of singe JujuUnit of application."""
+
+    unit: str
```
rgildein, line 124: not-blocking: Maybe name would be more suitable?

cou/apps/app.py
```diff
@@ -170,20 +178,26 @@ def __post_init__(self) -> None:
         self.os_origin = self._get_os_origin()
         # subordinates don't have units
         units = getattr(self.status, "units", {})
-        for unit in units.keys():
-            workload_version = self.status.units[unit].workload_version
-            self.units[unit]["workload_version"] = workload_version
+        for application_unit in units.keys():
+            unit = self.status.units[application_unit]
```
rgildein, line 182:
```suggestion
        for name, unit in units.items():
```

cou/steps/analyze.py
```diff
@@ -78,6 +79,30 @@ def _split_control_plane_and_data_plane(
...
+        """Move control plane applications to data plane if any of its units is on data plane.
+
+        :param control_plane: List of control plane applications.
+        :type control_plane:  list[OpenStackApplication]
+        :param data_plane: List of data plane applications.
+        :type data_plane:  list[OpenStackApplication]
+        """
+        data_plane_machines = set().union(
+            *[{unit.machine for unit in app.units} for app in data_plane]
+        )
```
rgildein, line 95:
```suggestion
        data_plane_machines = {unit.machine for unit in app.units for app in data_plane}
```

cou/steps/analyze.py
```diff
@@ -78,6 +79,30 @@ def _split_control_plane_and_data_plane(
...
+
+        apps_to_move = []
+        for app in control_plane:
+            if len([unit for unit in app.units if unit.machine in data_plane_machines]) > 0:
+                apps_to_move.append(app)
+
+        for app in apps_to_move:
+            control_plane.remove(app)
+            data_plane.append(app)
```
rgildein, line 104: This is not good design pattern to update input arguments, this function should instead return two lists.

cou/apps/app.py
```diff
@@ -270,12 +284,13 @@ def current_os_release(self) -> OpenStackRelease:
...
         :rtype: OpenStackRelease
         """
-        os_versions = {unit_values["os_version"] for unit_values in self.units.values()}
+        os_versions = {unit.os_version for unit in self.units}
 
         if len(os_versions) == 1:
             return os_versions.pop()
         # NOTE (gabrielcocenza) on applications that use single-unit or paused-single-unit
         # upgrade methods, more than one version can be found.
+        # return OpenStackRelease("ussuri")
```
rgildein, line 293: I think this was used for debugging purpose, or not?

cou/apps/app.py
esunar: done

cou/steps/analyze.py
esunar: this does not work

cou/apps/app.py
esunar: done

cou/apps/app.py
esunar: any suggestions

cou/steps/analyze.py
esunar: done

cou/steps/analyze.py
rgildein:
```suggestion
        data_plane_machines = {unit.machine for app in data_plane for unit in app.units}
```
Sorry wrong order. Here is an example how you can do it in Python
```python
[f"{i}" for app in [[1,2,3], [3,4,5]] for i in app]
```

cou/apps/app.py
rgildein: `name`

cou/steps/analyze.py
```diff
@@ -53,6 +53,7 @@ async def create(cls, model_name: Optional[str] = None) -> Analysis:
         apps = await Analysis._populate(model_name)
 
         control_plane, data_plane = cls._split_control_plane_and_data_plane(apps)
+        control_plane, data_plane = cls._remove_data_plane_apps(control_plane, data_plane)
```
agileshaw, line 56: These two functions are both returning control_plane and data_plane apps and are not being used anywhere else, I think we should merge them.

cou/apps/app.py
agileshaw: +1 on renaming it to `name`

cou/apps/app.py
```diff
@@ -118,6 +117,16 @@ def decorator(application: type[OpenStackApplication]) -> type[OpenStackApplicat
         return decorator
 
 
+@dataclass
+class ApplicationUnit:
+    """Representation of singe JujuUnit of application."""
```
agileshaw, line 122:
```suggestion
    """Representation of a singe unit of an application."""
```
I don't think `JujuUnit` is approprite here because it doesn't contain the object; instead, it only provides a handful attributes of a unit

cou/steps/analyze.py
```diff
@@ -78,6 +79,32 @@ def _split_control_plane_and_data_plane(
                 control_plane.append(app)
         return control_plane, data_plane
 
+    @classmethod
+    def _remove_data_plane_apps(
```
agileshaw, line 83: As stated in another comment, I think we should merge this function into `_split_control_plane_and_data_plane`. But if in the end we decide to keep them separated, we should at least update the function title. Right now it is misleading: it sounds like you are deleting data plane apps where in fact you are postponing the upgrade of some control-plane components.

agileshaw (changes requested):

cou/steps/analyze.py
esunar: ok

cou/apps/app.py
esunar: ok

cou/apps/app.py
esunar: done

cou/steps/analyze.py
esunar: I do not agree with this. There is clearly a separate logic, so both deserve their own function.

cou/steps/analyze.py
esunar: I changed its name to `_move_to_data_plane_if_required`. But I do not agree with merging.

cou/steps/analyze.py
rgildein: I agree with Erhan here.

gabrielcocenza: I think we can have a middle term by calling ` _move_to_data_plane_if_required` inside `_split_control_plane_and_data_plane` and return control_plane and data_plane

cou/steps/analyze.py
```diff
@@ -78,6 +79,32 @@ def _split_control_plane_and_data_plane(
...
+            unit.machine for units in [app.units for app in data_plane] for unit in units
+        }
+
+        apps_to_move = []
+        for app in control_plane:
+            if len([unit for unit in app.units if unit.machine in data_plane_machines]) > 0:
+                apps_to_move.append(app)
+
+        control_plane_apps = [app for app in control_plane if app not in apps_to_move]
+        data_plane_apps = data_plane + apps_to_move
```
rgildein, line 105: I think it will be more readable if we moved this logic to previous for loop and remove app from copy of `control_plane` + add app to copy of `data_plane`. Or maybe even create those list from scratch. What do you say?

cou/steps/analyze.py
```diff
@@ -78,6 +79,32 @@ def _split_control_plane_and_data_plane(
                 control_plane.append(app)
         return control_plane, data_plane
 
+    @classmethod
+    def _move_to_data_plane_if_required(
```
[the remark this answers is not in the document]
rgildein: With sets we will mess with order and that is important, or not?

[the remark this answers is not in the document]
gabrielcocenza: Good point, we will mess with the order. However, this shows that the order should happen after splitting into control and data plane. In a scenario where cinder is collocated with ceph/nova , it should upgrade cinder first. When we append the list cinder will be the last one to upgrade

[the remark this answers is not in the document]
rgildein: Do you have any suggestion how we can ensure that the order happen in later phase?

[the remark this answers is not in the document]
gabrielcocenza: I think we can transfer this logic to the class method `create` or even create a function with this logic and call it for `control_plane` and `data_plane` 
https://github.com/canonical/charmed-openstack-upgrader/blob/ea08be15bcbd793a95769348b58dbcc63dc84374/cou/steps/analyze.py#L108-L120

gabrielcocenza (dismissed):
LGTM

tests/unit/conftest.py
```diff
@@ -61,7 +62,7 @@ def status():
         ]
     )
 
-    mock_cinder_on_nova = mock.MagicMock(spec_set=ApplicationStatus())
+    mock_cinder_on_nova = MagicMock(spec_set=ApplicationStatus())
```
agileshaw, line 65: Can we unify the approach for mocking unit (L#31) and application?

gabrielcocenza (approved):

agileshaw (approved):
LGTM

PR #69, opened 2023-09-18
COUModel implementation

rgildein:
Replace `model_name` with `COUModel` and drop all functions from `cou.utils.juju_utils`.


tests/unit/steps/test_steps_analyze.py
```diff
@@ -69,69 +70,61 @@ async def test_analysis_dump(mocker, apps):
...
     test_model.applications = {app_name: generate_app(app_name) for app_name in apps_name}
-    juju_model = mocker.patch("cou.utils.juju_utils._get_model")
-    juju_model.return_value = test_model
-    mocker.patch("cou.utils.juju_utils.get_status", return_value=full_status)
-    mocker.patch(
-        "cou.utils.juju_utils.get_application_config", return_value=config["openstack_ussuri"]
-    )
+    model.get_status = AsyncMock(return_value=full_status)
+    model.get_application_config = AsyncMock(return_value=config["openstack_ussuri"])
+
```
esunar, line 98:
```suggestion
    async def get_charm_name(value):
        return value
    model.get_charm_name = get_charm_name
```

cou/utils/juju_utils.py
```diff
@@ -641,6 +198,15 @@ async def _get_unit(self, name: str) -> Unit:
 
         return unit
 
+    async def check_model_name(self) -> None:
+        """Check model name.
+
+        This function should be used if None was used for model name, so it mean that
+        default model will be used.
```
agileshaw, line 205:
```suggestion
        Get default model's name if no model name was explicitly passed in.
```

cou/utils/juju_utils.py
```diff
@@ -653,16 +219,20 @@ async def get_application_config(self, name: str) -> Dict:
...
         :type application_name: str
         :return: Charm name
-        :rtype: Optional[str]
+        :rtype: str
         :raises: ApplicationNotFound
+        :raises: ApplicationError if charm_name is None
         """
         app = await self._get_application(application_name)
+        if app.charm_name is None:
+            raise ApplicationError(f"charm_name could not be obtain for app {application_name}")
```
agileshaw, line 234:
```suggestion
            raise ApplicationError(f"Cannot obtain charm_name for {application_name}")
```

cou/utils/juju_utils.py
```diff
@@ -44,127 +43,10 @@
...
-    :returns: Full juju status output
-    :rtype: FullStatus
-    """
-    model = await _get_model(model_name)
-    return await model.get_status()
-
-
-def _normalise_action_results(results: Dict[str, str]) -> Dict[str, str]:
+def _normalize_action_results(results: Dict[str, str]) -> Dict[str, str]:
     """Put action results in a consistent format.
```
agileshaw, line 50:
```suggestion
    """Unify action results format.
```

cou/cli.py
```diff
@@ -91,11 +91,12 @@ async def analyze_and_plan(model_name: Optional[str] = None) -> UpgradeStep:
     :return: Generated upgrade plan.
     :rtype: UpgradeStep
     """
-    model_name = model_name or await juju_utils.get_current_model_name()
+    model = COUModel(model_name)
+    await model.check_model_name()  # check model name and obtain if it was None
```
agileshaw, line 95: I'm not a fan of the pattern here. Why do we want to call `check_model_name` outside the class? Wouldn't it be better to let COUModel automatically handle the situation where model_name input is `None`?

Also, maybe we should think about whether to auto-connect to a model when instantiating a COUModel.

agileshaw (changes requested):

cou/cli.py
esunar: We cannot run async functions in `__init__`.

esunar: Alternative would be something like this:
```
    @classmethod
    async def create(cls, name: str):
        model = COUModel(name)
        await model.check_model_name()
        return await model._connect()
```

esunar (dismissed):
Overall the changes are good. One recommendation is we can delegate object creation to a static method.

cou/utils/juju_utils.py
rgildein: I changed it to `Set ...` since we are setting the `_name` value.

cou/cli.py
rgildein: We can get even loop and run it on that, but I'm not sure how it will behave or we can go with Erhan suggestion or simply define `create_model` function outside.

tests/unit/conftest.py
```diff
@@ -232,7 +234,54 @@ def units():
...
+            "openstack-origin": {"value": "distro"},
+            "action-managed-upgrade": {"value": True},
+        },
+        "openstack_wallaby": {"openstack-origin": {"value": "cloud:focal-wallaby"}},
+        "auxiliary_ussuri": {"source": {"value": "distro"}},
+        "auxiliary_wallaby": {"source": {"value": "cloud:focal-wallaby"}},
+    }
+
+
+async def get_status():
```
gabrielcocenza, line 249: We probably will need to be able to get status from more than one json file. I think this function should be ready to receive the file name as parameter

cou/cli.py
agileshaw: As discussed elsewhere, we agreed on Erhan's suggestion with one modificaton: remove `check_model_name` method.

tests/unit/conftest.py
rgildein: I agree, but this could be done in next PR when we actually use it.

tests/unit/conftest.py
```diff
@@ -26,12 +28,12 @@
 
 @pytest.fixture
 def status():
-    mock_keystone_ussuri = mock.MagicMock(spec_set=ApplicationStatus())
+    mock_keystone_ussuri = MagicMock(spec=ApplicationStatus())
```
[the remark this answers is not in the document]
rgildein: Thanks, I checked the documentation and you are right. I did not know about it and I changed it only due same failed tests. Changing back and I will use it everywhere.

cou/cli.py
rgildein: I added the staticmethod create  with simple `_connect` call and remove `check_model_name`.

gabrielcocenza (approved):
LGTM

tests/unit/utils/test_juju_utils.py
```diff
@@ -153,6 +153,20 @@ def test_coumodel_init(mocker):
     assert model.name == name
 
 
+@pytest.mark.asyncio
+async def test_coumodel_create(mocked_model):
+    """Test COUModel create function with no model_name defined."""
+    model = await juju_utils.COUModel.create(None)
```
agileshaw, line 159: Can we also test model creation with name input? Parameterize it with `None` should be enough

cou/cli.py
agileshaw: Looks good, thanks!

tests/unit/utils/test_juju_utils.py
rgildein: This is tested in `test_coumodel_init`, so that's why I skipped it. Do you think it's really needed?

agileshaw: This looks good. It was hidden from the changes so I wasn't aware it was already tested elsewhere.

agileshaw (approved):
LGTM

Issue #67, opened 2023-09-18
Get rid of  Appfactory.


rgildein:
The `AppFactory` class bring only more complexity for testing and it making some magic behind the scene.

This issue is opened for discussion.

PR #70, opened 2023-09-19
drop AppFactory

rgildein:
Remove usage of `AppFactory` and define dictionary with all classes. 

fixes: #67


rgildein:
closing because it's just for showcase

PR #71, opened 2023-09-19
Alternative AppFactory

esunar:


rgildein:
closing because it's just for showcase
related it with #67

PR #73, opened 2023-09-20
Add support for upgrading Ceph control-plane components

agileshaw:
Create CephMonApplication class to handle its specific upgrade requirement ([ref](https://docs.openstack.org/charm-guide/latest/project/issues/upgrade-issues.html#ceph-require-osd-release))


cou/apps/ceph.py
```diff
@@ -0,0 +1,80 @@
...
+        :return: Plan to check if application workload has been upgraded
+        :rtype: UpgradeStep
+        """
+        track: str = channel.split("/", maxsplit=1)[0]
+        set_command = f"sudo ceph osd require-osd-release {track}"
+
+        return UpgradeStep(
+            description=(f"Set '{track}' for require-osd-release option on ceph-mon units"),
+            parallel=parallel,
+            function=app_utils.run_on_all_units,
```
gabrielcocenza, line 76: From the [docs](https://docs.openstack.org/charm-guide/latest/project/issues/upgrade-issues.html#ceph-require-osd-release) it says:

> If it needs changing, it can be done manually on any ceph-mon unit

I think we can run just on the leader unit instead of all units. It looks overkill

cou/apps/ceph.py
```diff
@@ -0,0 +1,80 @@
...
+        """Post Upgrade planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: Plan that will add post upgrade as sub steps.
+        :rtype: list[UpgradeStep]
+        """
+        return [
+            self._get_reached_expected_target_plan(target),
+            self._get_set_require_osd_release_plan(self.target_channel(target)),
```
gabrielcocenza, line 55: If target channel is equal to the current channel (e.g: from ussuri to victoria both will have `octopus`) than we can skip this part and avoid a ssh into the units

cou/apps/ceph.py
```diff
@@ -0,0 +1,80 @@
...
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        return [
+            self._get_upgrade_current_release_packages_plan(),
+            self._get_refresh_charm_plan(target),
+            self._get_set_require_osd_release_plan(self.expected_current_channel),
```
gabrielcocenza, line 42: I think this can be problematic. Imagine the following situation:

Upgrading from Victoria to Wallaby:

- if for some reason the COU stops after setting `pacific` and the user runs again, it will change back to `octopus` which I think it's not ideal. I think we should always query and check if it's as expected

gabrielcocenza (changes requested):

cou/apps/ceph.py
agileshaw: I don't see how that could happen because setting to the next target is done at the very end of the upgrade process. So if it has been changed to `pacific` in this case and stops, the upgrade won't be planed for `ceph-mon` in the next run.

cou/apps/ceph.py
agileshaw: You are right, it seems that setting for one unit should be enough. Will update it in the next patch.

cou/apps/app.py
```diff
@@ -420,7 +421,7 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
             self._get_workload_upgrade_plan(target),
         ]
 
-    def post_upgrade_plan(self, target: OpenStackRelease) -> list[UpgradeStep]:
+    def post_upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
```
gabrielcocenza, line 424: nit: Why are we changing to Optional here?

cou/apps/ceph.py
```diff
@@ -0,0 +1,82 @@
...
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        return [
+            self._get_upgrade_current_release_packages_plan(),
+            self._get_refresh_charm_plan(target),
+            self._get_set_require_osd_release_plan(target, self.expected_current_channel),
+        ]
+
+    def post_upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
```
gabrielcocenza, line 44: Same here. I don't see why we are changing to optional, but if we do please change the docstrings

cou/apps/ceph.py
```diff
@@ -0,0 +1,82 @@
...
+        :type target: OpenStackRelease
+        :return: Plan that will add post upgrade as sub steps.
+        :rtype: list[UpgradeStep]
+        """
+        return [
+            self._get_reached_expected_target_plan(target),
+            self._get_set_require_osd_release_plan(target, self.target_channel(target)),
+        ]
+
+    def _get_set_require_osd_release_plan(
```
gabrielcocenza, line 57: I think `_get_set` is a little confusing. Maybe `_set_require_osd_release_plan` is a better naming?

cou/apps/app.py
agileshaw: Because the `_get_set_require_osd_release_plan` could return `None` to the `post_upgrade_plan` function in CephMonApplicationm. And since `post_upgrade_plan` is inherited from the parent class, their return type needs to be aligned.

cou/apps/ceph.py
agileshaw: Because the list could contain either UpgradeStep or None (from the `_get_set_require_osd_release_plan` function)

cou/apps/ceph.py
agileshaw: We should keep the naming consistent: `_get_*_plan`. But I agree, having both get and set in the title could be confusing. I have changed it to `_get_change_require_osd_release_plan` in the new patch. Please let me know if it is better.

cou/apps/ceph.py
```diff
@@ -0,0 +1,82 @@
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param channel: The channel to get ceph track from.
+        :type str
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return: Plan to check if application workload has been upgraded
+        :rtype: UpgradeStep
+        """
+        if self.expected_current_channel != self.target_channel(target):
```
gabrielcocenza, line 71: Imagine we get a cloud that it's on `ussuri` but for some reason they didn't update `require-osd-release` and it's as `natilus`. We are trying to upgrade to Victoria but this logic won't let update to octupus because the expected_current_channel and target_channel are equal.

I think this logic should be on `post_upgrade_plan` to decide or not calling ` _get_change_require_osd_release_plan`.

Do you know if there are consequences when calling `require-osd-release` as for example restarting the services? If we do have, maybe we should query the value first before calling it to change. If there is no consequences, then we should always call it on `pre_upgrade_plan` just to ensure that it's set right.

agileshaw: As discussed elsewhere, overwriting with the same value doesn't seem to have consequences. However, after careful consideration, check its value before setting is the safest and best practices. I have added this check in the step's function instead of a plan function because it requires async

cou/apps/ceph.py
```diff
@@ -0,0 +1,80 @@
...
+        :rtype: UpgradeStep
+        """
+        track: str = channel.split("/", maxsplit=1)[0]
+        return UpgradeStep(
+            description=(
+                f"Ensure require-osd-release option on ceph-mon units correctly set to '{track}'"
+            ),
+            parallel=parallel,
+            function=set_require_osd_release_option,
+            unit=list(self.units.keys())[0],
```
gabrielcocenza, line 77: nit: units will change with #65.
I would declare something like:
```python
ceph_mon_unit, *_ = self.units
```
then when we merge the PR described the change here would be passing `ceph_mon_unit.name` instead of `ceph_mon_unit`

cou/utils/app_utils.py
```diff
@@ -44,9 +44,61 @@ async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
...
+                    cmd=command, result=result
+                )
 
         except JujuError as exc:
-            raise PackageUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+            raise RunUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+
+
+async def set_require_osd_release_option(unit: str, model: COUModel, ceph_release: str) -> None:
+    """Run package updates and upgrades on each unit of an Application.
```
gabrielcocenza, line 56: Fix docsting

cou/apps/ceph.py
```diff
@@ -0,0 +1,80 @@
...
+    def _get_change_require_osd_release_plan(
+        self, channel: str, parallel: bool = False
+    ) -> UpgradeStep:
+        """Get plan to set correct value for require-osd-release option on ceph-mon.
+
+        :param channel: The channel to get ceph track from.
+        :type str
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return: Plan to check if application workload has been upgraded
```
gabrielcocenza, line 67: Fix docstring

cou/utils/app_utils.py
```diff
@@ -44,9 +44,61 @@ async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
...
+
+    :param unit: The unit name where the check command runs on.
```
gabrielcocenza, line 58:
```suggestion
    :param unit: The ceph-mon unit name where the check command runs on.
```

gabrielcocenza (changes requested):

cou/apps/ceph.py
agileshaw: TIL. I didn't know we can do this on a dictionary. Thanks!

cou/utils/app_utils.py
```diff
@@ -44,9 +44,61 @@ async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
...
+                )
 
         except JujuError as exc:
-            raise PackageUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+            raise RunUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+
+
+async def set_require_osd_release_option(unit: str, model: COUModel, ceph_release: str) -> None:
+    """Run package updates and upgrades on each unit of an Application.
+
```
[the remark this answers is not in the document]
agileshaw: I added it to the `_get_*_plan` function

gabrielcocenza (dismissed):
LGTM. Thanks

cou/utils/app_utils.py
```diff
@@ -44,9 +44,59 @@ async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
...
+                raise RunUpgradeError(f"Cannot upgrade packages on {unit}.") from CommandRunFailed(
+                    cmd=command, result=result
+                )
 
         except JujuError as exc:
-            raise PackageUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+            raise RunUpgradeError(f"Cannot upgrade packages on {unit}.") from exc
+
+
+async def set_require_osd_release_option(unit: str, model: COUModel, ceph_release: str) -> None:
```
aieri, line 55: if this function is specific to the ceph-mon class, why put it here?

cou/exceptions.py
```diff
@@ -63,8 +63,8 @@ class ApplicationError(COUException):
     """Exception raised when Application does something unexpected."""
 
 
-class PackageUpgradeError(COUException):
-    """Exception raised when a package upgrade fails."""
+class RunUpgradeError(COUException):
```
aieri, line 66: how is this change related to this PR?

cou/utils/app_utils.py
agileshaw: Because ideally in the future, I think we should have all upgrade related functions put in a single file, like the current `app_utils.py`. And the application class file should just contain the knowledge about which steps an application needs etc.

cou/utils/app_utils.py
```diff
@@ -44,9 +44,59 @@ async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
...
+
+    :param unit: The ceph-mon unit name where the check command runs on.
+    :type str
+    :param model: COUModel object
+    :type model: COUModel
+    :param ceph_release: The ceph release to set for require-osd-release.
+    :type str
+    :raises RunUpgradeError: When an upgrade fails.
+    """
+    check_command = "ceph osd dump | grep require_osd_release"
```
aieri, line 66: (minor) in order to limit potential issues with invoking chains of shell commands, you could only run `ceph osd dump` and parse the output in python

tests/unit/apps/test_ceph.py
```diff
@@ -0,0 +1,83 @@
...
+        "ceph-mon",
+        status["ceph-mon_xena"],
+        config["auxiliary_xena"],
+        model,
+        "ceph-mon",
+    )
+
+    plan = app.generate_upgrade_plan(target)
+
+    steps_description = [
```
aieri, line 49: Isn't it odd that we're validating generating the correct plan by checking textual descriptions instead of function calls?
Wouldn't it be more robust, expressive, and maintainable to assert that certain functions are called in a certain order (and with certain arguments)

aieri (commented):
Looking good overall, I have a few questions

cou/utils/app_utils.py
agileshaw: Updated.

tests/unit/apps/test_ceph.py
agileshaw: I created an `assert_plan` function and used it to verify the steps generated by ceph-mon class, including their descriptions, function calls, and parameters. If this solution is accepted, I will create an issue so tests for other application classes can align with it.

cou/utils/app_utils.py
```diff
@@ -56,21 +56,27 @@ async def set_require_osd_release_option(unit: str, model: COUModel, ceph_releas
...
     try:
         check_result = await model.run_on_unit(unit_name=unit, command=check_command, timeout=600)
         if str(check_result["Code"]) == "0":
             logger.debug(check_result["Stdout"])
-            current_require_osd_release = check_result["Stdout"].strip().split()[-1]
+
+            dump_output = check_result["Stdout"].strip().split("\n")
+            output_dict = {
+                key: value for line in dump_output for key, value in [line.strip().split()]
+            }
```
aieri: This starts being a bit complex. Just an idea:
```suggestion
            output_dict = dict([line.split(maxsplit=1) for line in dump_output])
```

...though I'm not too sure which version is simpler :)

Of course an alternative would be not building the full dict at all and simply iterate through the output until you hit the key you want. Something like:

```
for line in dump_output:
    if line.startswith('require_osd_release'):
        release = line.split()[1]
```

tests/unit/apps/utils.py
```diff
@@ -18,3 +18,12 @@ def assert_plan_description(upgrade_plan, steps_description):
...
         assert sub_step.description == description
+
+
+def assert_plan(upgrade_plan, expected_plan):
+    assert len(upgrade_plan.sub_steps) == len(expected_plan)
+    sub_steps_check = zip(upgrade_plan.sub_steps, expected_plan)
+    for sub_step, expected_step in sub_steps_check:
+        assert sub_step.description == expected_step["description"]
+        assert expected_step["function_name"] in sub_step.function.__dict__.values()
+        assert sub_step.params == expected_step["params"]
```
aieri, line 29: nit: it would be more readable to keep the same L/R order between actual and expected. Since the `in` keyword locks you into keeping the expected part on the L I would flip the first and third assertions.

aieri (changes requested):
Most comments are optional, except being able to handle dump outputs without `require_osd_release` (just in case)

cou/utils/app_utils.py
```diff
@@ -56,21 +56,27 @@ async def set_require_osd_release_option(unit: str, model: COUModel, ceph_releas
...
+
+            current_require_osd_release = output_dict["require_osd_release"]
```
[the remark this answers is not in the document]
agileshaw: I was thinking it could be just raised by funciton and catch by entry-point as a un-handled error. But I agree, it would be better to avoid that. I have added a default value `""` for `current_require_osd_release`, which will prevent `KeyError`.

cou/utils/app_utils.py
agileshaw: Thanks for the suggestion. I like the iteration until hitting target approach and has implemented it in the new patch.

gabrielcocenza (approved):
LGTM

aieri (approved):

PR #74, opened 2023-09-21
Changes in snapcraft.yaml to publish

gabrielcocenza:
- added instrunctions to manually add alias
- removed write access to .local/share/juju
- [juju-client-observe-interface](https://snapcraft.io/docs/juju-client-observe-interface) already has read access to .local/share/juju

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ different repository, and only use some of the code and ideas found here.
 # Setup
 
 ```bash
+# Instructions for local builds until we have automatic connections and alias
 make clean
 sudo snap remove charmed-openstack-upgrader --purge
 make build
--- a/README.md
+++ b/README.md
@@ -12,10 +12,10 @@ rm *.snap
 snapcraft clean
 sudo snap remove charmed-openstack-upgrader --purge
 make build
-sudo snap install ./charmed-openstack-upgrader.snap --devmode
+sudo snap install ./charmed-openstack-upgrader.snap --dangerous
 sudo snap connect charmed-openstack-upgrader:juju-client-observe snapd
-sudo snap connect charmed-openstack-upgrader:dot-local-share-juju snapd
-sudo snap connect charmed-openstack-upgrader:local-cou snapd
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -54,11 +52,7 @@ parts:
       craftctl set version=${VERSION}
 
 plugs:
-  dot-local-share-juju:
```
rgildein, line 57: I think that this is needed, because we are accessing it with lubjuju, or not?

rgildein: Simple reading access would be enough, but we need to be able to read from there.

gabrielcocenza: The [juju-client-observe-interface](https://snapcraft.io/docs/juju-client-observe-interface) already give us read access to .local/share/juju used in libjuju

rgildein: Perfect 👍

README.md
```diff
@@ -11,10 +11,10 @@ different repository, and only use some of the code and ideas found here.
 make clean
 sudo snap remove charmed-openstack-upgrader --purge
 make build
-sudo snap install ./charmed-openstack-upgrader.snap --devmode
+sudo snap install ./charmed-openstack-upgrader.snap --dangerous
 sudo snap connect charmed-openstack-upgrader:juju-client-observe snapd
-sudo snap connect charmed-openstack-upgrader:dot-local-share-juju snapd
-sudo snap connect charmed-openstack-upgrader:local-cou snapd
+sudo snap connect charmed-openstack-upgrader:dot-local-cou snapd
+sudo snap alias charmed-openstack-upgrader.cou cou
```
agileshaw, line 17: We should state that this is for local builds. It would be easier when we are ready to add `install from snapstore` section

agileshaw (changes requested):

agileshaw (approved):

rgildein (approved):

PR #79, opened 2023-09-22
Remove invalid model arguments

agileshaw:
Currently we are passing `model` as an argument to some COUModel methods, which is not correct and causes `unexpected keyword argument` error. Removed the argument to fix this issue.


rgildein (approved):

gabrielcocenza (commented):
~~I think we should also remove on line `517` on app.py file.~~
Edit: sorry you already removed

PR #81, opened 2023-09-22
Add pylint docparams extension to check docstrings.

gabrielcocenza:
Docstrings are hard to maintaining and review and can be easily outdated.
This PR adds docparams extension to be more strict with those checks and also fix the problems found in the project.
Documentation of the plugin can be found [here](https://pylint.pycqa.org/en/latest/user_guide/checkers/extensions.html#parameter-documentation-checker)


agileshaw (approved):
I like this added feature! LGTM. I suggest also updating charm-spec templates once the changes are accepted.

PR #85, opened 2023-09-25
Add private methods on docstring checks.

gabrielcocenza:
[pylint]((https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#no-docstring-rgx) for standard doesn't consider private methods to be checked for docstrings. This PR changes this behavior to not check just the magic methods.

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -96,7 +96,7 @@ def retry(
     :rtype: Callable
     """
 
-    def _wrapper(func: Callable) -> Callable:
+    def _wrapper(func: Callable) -> Callable:  # pylint: disable=W9011
         @wraps(func)
         async def wrapper(*args: Any, **kwargs: Any) -> Any:  # pylint: disable=W9011
             attempt: int = 0
@@ -180,9 +180,9 @@ class COUModel:
 
         :param name: Name of application
         :type name: str
+        :raises ApplicationNotFound: When cannot find Application in the model.
         :return: Application
         :rtype: Application
-        :raises: ApplicationNotFound
         """
         model = await self._get_model()
         app = model.applications.get(name)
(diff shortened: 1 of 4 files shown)


cou/utils/juju_utils.py
```diff
@@ -180,9 +180,9 @@ async def _get_application(self, name: str) -> Application:
 
         :param name: Name of application
         :type name: str
+        :raises ApplicationNotFound: When cannot find Application in the model.
```
dashmage, line 183: nit: This sounds a bit weird. Would it be better like `When Application is not found in the model`

cou/utils/juju_utils.py
```diff
@@ -203,9 +207,9 @@ async def _get_unit(self, name: str) -> Unit:
 
         :param name: Name of unit
         :type name: str
+        :raises UnitNotFound: When cannot find unit in the model.
```
dashmage, line 210: nit: Same here. This might be better as `When unit is not found in the model`

agileshaw (commented):
LGTM. Just two small rewording suggestions

cou/utils/juju_utils.py
```diff
@@ -96,7 +96,7 @@ def retry(
     :rtype: Callable
     """
 
-    def _wrapper(func: Callable) -> Callable:
+    def _wrapper(func: Callable) -> Callable:  # pylint: disable=W9011
```
rgildein, line 99: Why this disable is needed here?

gabrielcocenza: If we remove lint will complain that is missing docstring and I think there isn't much to document about it. Do you think worth adding docstring to this function?

agileshaw (approved):
LGTM

cou/utils/juju_utils.py
rgildein: It was not there before ant did not failed, that's why I'm asking.

agileshaw: Because pylint wasn't checking docstrings for private functions before. With Gabriel's patch here, it is now checking so we need to disable it for this wrapper function..

rgildein (approved):
LGTM

PR #88, opened 2023-09-26
[pytest] smoke func tests

rgildein:
- move functional tests to separate directory
- add smoke functional tests

This PR providing simple smoke tests against simple bundle, which can be run even on LXD. I chose to use pytest-operator since it's my framework of first choice and I'm more confident using it than zaza. However, after writing the tests, I needed to update the tox and Makefile, where I realized that using zaza smoke test might be more appropriate. 

There are two things that we can discuss/improve.
1) Using zaza instead of pytest-operator. I will provide another PR with example.
2) Using different application, although the current bundle could be deployed on LXD, but it is taking too long to be in the Idle state.


rgildein:
Since we are using zaza to configure cloud we could not make switch to use only pytest-operator and it does not make sense to use both zaza and pytest-operator.

PR #89, opened 2023-09-26
[zaza] smoke func tests

rgildein:
This PR providing smoke tests against simple bundle, which can be run even on LXD.


Makefile
```diff
@@ -53,7 +53,11 @@ dev-environment:
 	@echo "Creating virtualenv and installing pre-commit"
 	@tox -r -e dev-environment
 
-functional: build
+functional-smoke: build
+	@echo "Executing functional tests using built snap"
```
agileshaw, line 57:
```suggestion
	@echo "Executing smoke functional tests using built snap"
```

tests/functional/tests/model.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # changing configuration and validating it was changed
+        zaza.sync_wrapper(self.model.set_application_config)(TESTED_APP, new_config)
+        zaza.sync_wrapper(self.model.wait_for_idle)(120, apps=[TESTED_APP])
+        config = zaza.sync_wrapper(self.model.get_application_config)(TESTED_APP)
+
+        self.assertFalse(config["enable-auto-restarts"]["value"])
+
+    def test_upgrade_charm(self):
+        """Test upgrade charm to already deployed channel.
```
agileshaw, line 78:
```suggestion
        """Test upgrade charm to the latest revision of the current channel.
```

tests/functional/tests/model.py
```diff
@@ -0,0 +1,87 @@
...
+
+        This test does not actually perform the update of the charm, but rather just the test
+        results of such operation.
```
agileshaw, line 81: I don't think it's correct to claim that upgrade is not actually performed. Rather, because the deployed version is already the latest in the current channel, we won't observe any changes. So I suggest rewording it to the following:

```suggestion
        This test only checks the results of such an upgrade operation.
```

tests/functional/tests/model.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        status = zaza.model.get_status()
+        # get original channel, so we will not change it
```
agileshaw, line 84:
```suggestion
        # get the current channel, so we will not change it
```

tests/functional/tests/model.py
```diff
@@ -0,0 +1,87 @@
...
+        charm = zaza.sync_wrapper(self.model.get_charm_name)(TESTED_APP)
+        self.assertEqual(TESTED_APP, charm)
+
+    def test_get_status(self):
+        """Test COUModel.get_status."""
+        status = zaza.sync_wrapper(self.model.get_status)()
+        self.assertIn(TESTED_APP, status.applications)
+
+    def test_run_action(self):
+        """Test running action."""
```
agileshaw, line 39:
```suggestion
        """Test run action."""
```

tests/functional/tests/bundles/smoke.yaml
```diff
@@ -0,0 +1,12 @@
+variables:
+  openstack-origin: &openstack-origin distro
+
+series: focal
+
+applications:
+  rabbitmq-server:
+    charm: ch:rabbitmq-server
+    num_units: 1
+    channel: 3.9/stable
```
agileshaw, line 10: Should we maybe use `3.8/stable` channel so we can validate both same channel refreshing and channel switching?

agileshaw (changes requested):
Overall LGTM. I suggested some improvements inline

gabrielcocenza (commented):
IIUC this smoke test is more designed to test `COUModel` and indirectly python libjuju. I would rather prefer to run `cou run --no-interactive` and expect that applications on the bundle successfully upgraded. I think it would be a more solid functional test.

rgildein:
> IIUC this smoke test is more designed to test `COUModel` and indirectly python libjuju. I would rather prefer to run `cou run --no-interactive` and expect that applications on the bundle successfully upgraded. I think it would be a more solid functional test.

Since we choose to not use libjuju directly in cou and introduced the `COUModel` we need to tests it. The smoke tests should do COUModel testing, CLI testing and yes we can do even simple `cou run` if it works fine. Right now this PR is limited to test only `COUModel` and other tests can be provided later.

tests/functional/tests/bundles/smoke.yaml
rgildein: It would be nice to have, but since these will be done in func tests and do not think it's important. With this the charms will be changed and could not be put to previous state, so tests will failed in second run. Also these will take more time and smoke tests should be fast as possible + as Gabriel suggested we use smoke for running `cou run` and that will try to do same things.

I'm not saying it's not good, I'm opening just and discussion.

rgildein:
As we discussed I moved `smoke` bundle to `gate_tests` and remove aliases. Like this we can add more tests to smoke and gate at the same time, and smoke bundle will be run first before full tests.
example:
```yaml
tests:
  - tests.functional.tests.model.COUModelTest
  - tests.functional.tests.cli.CLITest1
  - tests.functional.tests.cli.CLITest2
  - ...
  - openstack:
    - tests.functional.tests.backup.BackupTest
```

tests/functional/tests/model.py
```diff
@@ -0,0 +1,86 @@
...
+        exp_path = os.path.join(tmp_dir, test_file)
+        zaza.model.run_on_unit(unit_name=TESTED_UNIT, command=f"echo 'test' > {path}")
+
+        zaza.sync_wrapper(self.model.scp_from_unit)(TESTED_UNIT, path, tmp_dir)
+        self.assertTrue(os.path.exists(exp_path))
+
+    def test_changing_app_configuration(self):
+        """Test change of app configuration.
+
+        This tests cover set and get configuration option along with waiting for model to be idle.
```
agileshaw, line 63:
```suggestion
        This test covers set and get configuration option along with waiting for model to be idle.
```

agileshaw (approved):
LGTM. We may add more tests for smoke test in the future (e.g. upgrading a minimal setup like Gabriel suggested), but this is a good first iteration.

gabrielcocenza (approved):
LGTM

PR #93, opened 2023-09-29
Refactor OVN charm applications

### LINKED ISSUE #83: ovn 20.03 not recognized as a valid ussuri version
Running cou against an ussuri stsstack bundle yields the following error:
```
$ charmed-openstack-upgrader.cou plan 
Analyzing cloud... ✖
2023-09-23 21:17:59 [ERROR] 'ovn-central' with workload version 20.03.2 has no compatible OpenStack release in the lookup.
```

20.03 is a valid ovn version for ussuri so planning should not fail.

Additionally, the error message is obscure for anyone who isn't intimately familiar with cou. Which lookup?

gabrielcocenza:
- added ovn-chassis subordinate charm
- considered  20.03 version for ussuri in the OpenStack lookup
- raise ApplicationError if ovn charm use channels lesser than 22.03 and suggest to the user to manually make the upgrade.

 Closes #83


gabrielcocenza:
I've manually upgrade my deployment and followed the steps on the [documentation](https://docs.openstack.org/charm-guide/latest/project/procedures/ovn-upgrade-2203.html). OVN charms didn't change the workload version without running those commands again:

```shell
juju exec -a ovn-chassis 'apt update && apt -y install \
   --only-upgrade openvswitch-common ovn-common'
juju exec -a ovn-central 'apt update && apt -y install \
   --only-upgrade openvswitch-common ovn-common'
```
I think the solution will be more robust if we check the workload version instead of channel.

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+
+        :raises ApplicationError:  When workload version is lesser than 22.03.0.
+        """
+        if Version(self.status.workload_version) < Version("22.03.0"):
+            raise ApplicationError(
+                (
+                    "It's recommended to upgrade OVN to 22.03 before upgrading the cloud. "
+                    "Follow the instructions at: "
+                    "https://docs.openstack.org/charm-guide/latest/project/procedures/"
+                    "ovn-upgrade-2203.html"
```
agileshaw, line 92: Can we also mentioned that `{self.status.workload_version} is not currently handled` in addition to the recommandation?

agileshaw (changes requested):
Aside from the two small changes requested in-line, I'm still thinking whether we should have a table for compatible but non-supported versions instead of implementing an OVN-specific check function. In the future, we could easily run into a similar situation where COU cannot handle all compatible versions, so a more generic solution may be preferable. Let's discuss this synchronously.

cou/utils/openstack_lookup.csv
```diff
@@ -18,7 +18,7 @@ swift,2.24.0,2.26.0,2.26.0,2.27.0,2.27.0,2.28.0,2.28.0,2.29.0,2.29.0,2.30.0
 ceph,15.2.0,16.2.0,15.2.0,16.2.0,16.2.0,17.2.0,16.2.0,17.2.0,17.2.0,18.2.0
 gnocchi,4.3.0,4.4.0,4.4.0,4.4.1,4.4.0,4.4.1,4.4.1,4.4.2,4.4.1,4.4.2
 designate-bind,9.16.1,9.18.12,9.16.1,9.18.12,9.16.1,9.18.12,9.16.1,9.18.12,9.16.1,9.18.12
-ovn,22.03,22.09,22.03,22.09,22.03,22.09,22.03,22.09,22.03,22.09
+ovn,20.03,22.09,20.03,22.09,20.03,22.09,20.03,22.09,20.03,22.09
```
agileshaw, line 21: As discussed elsewhere, we will confirm the lower bound for each release instead of putting all of them as `20.03`. From what we are seeing [here](https://docs.openstack.org/charm-guide/latest/project/procedures/ovn-upgrade-2203.html#updated-neutron-api-plugin-ovn-charm), it seems, at least for yoga, it's lower bound should be 22.03

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        self._check_ovn_workload_version()
+        return [
+            self._get_upgrade_current_release_packages_plan(),
+            self._get_refresh_charm_plan(target),
+        ]
```
aieri, line 46: you could perhaps be more explicit about code reuse with `super()`

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def _check_ovn_workload_version(self) -> None:
+        """Check whether is necessary to manually upgrade OVN to 22.03.
+
+        :raises ApplicationError: When workload version is lesser than 22.03.0.
```
aieri, line 51:
```suggestion
        """Check whether it is necessary to manually upgrade OVN to 22.03.

        :raises ApplicationError: When workload version is lower than 22.03.0.
```

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        if any((Version(unit.workload_version) < Version("22.03.0") for unit in self.units)):
+            raise ApplicationError(
+                (
+                    "It's recommended to upgrade OVN to 22.03 before upgrading the cloud. "
```
aieri, line 56: "recommended" makes it sound like it's optional; I suggest saying that it is _necessary_ and that versions lower than 22.03 are not supported

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+        :type target: OpenStackRelease
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        self._check_ovn_workload_version()
+        return [
+            self._get_refresh_charm_plan(target),
+        ]
+
+    def _check_ovn_workload_version(self) -> None:
```
aieri, line 81: can we avoid this code duplication? For example you could move the loop outside of this function, make it a static method in the `OvnPrincipalApplication` class, and reuse it here

cou/apps/ovn.py
```diff
@@ -0,0 +1,94 @@
...
+    def pre_upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
+        """Pre Upgrade planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        self._check_ovn_workload_version()
+        return [
```
aieri, line 77: as above, `super()` may make what we're tweaking here clearer

cou/steps/plan.py
```diff
@@ -67,7 +76,12 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
...
-            app, (OpenStackSubordinateApplication, OpenStackAuxiliarySubordinateApplication)
+            app,
+            (
+                OpenStackSubordinateApplication,
+                OpenStackAuxiliarySubordinateApplication,
+                OvnSubordinateApplication,
+            ),
         ),
     )
     plan.add_step(control_plan_subordinate_upgrade_plan)
```
aieri, line 87: Typo (but unrelated to your submission)
```suggestion
    plan.add_step(control_plane_subordinate_upgrade_plan)
```

cou/steps/plan.py
```diff
@@ -57,7 +61,12 @@ async def generate_plan(analysis_result: Analysis) -> UpgradeStep:
         description="Control Plane principal(s) upgrade plan",
         target=target,
         filter_function=lambda app: not isinstance(
-            app, (OpenStackSubordinateApplication, OpenStackAuxiliarySubordinateApplication)
+            app,
+            (
+                OpenStackSubordinateApplication,
+                OpenStackAuxiliarySubordinateApplication,
+                OvnSubordinateApplication,
```
aieri, line 68: is this necessary? `OvnSubordinateApplication` is a subclass of `OpenStackAuxiliarySubordinateApplication`, so `isinstance(OvnSubordinateApplication(), OpenStackAuxiliarySubordinateApplication)` is supposed to return `True`

cou/utils/openstack_lookup.csv
aieri: correct, 22.03 is what's in the jammy repos, 20.03 is in the focal repos

tests/unit/apps/test_ovn.py
```diff
@@ -0,0 +1,197 @@
...
+        "ovn-chassis",
+    )
+    assert app.channel == "22.03/stable"
+    assert app.os_origin == ""
+    assert app.apt_source_codename is None
+    assert app.channel_codename == "yoga"
+    assert app.current_os_release == "yoga"
+
+
+def test_ovn_workload_lesser_22(status, config, model):
```
aieri, line 56: I think this test should be split into two (principal/subordinate)

aieri (changes requested):

cou/apps/ovn.py
gabrielcocenza: I've changed the description a little bit, but I preferred to not include `{self.status.workload_version}` because that represents just for the subordinates. For principal charms we would need to identify the unit with the version lower than 22.03 and I think we can be a little bit more generic to fit both cases.

cou/apps/ovn.py
gabrielcocenza: I couldn't use static method because we need `self` to access class attributes. However, I liked the idea and I've used class property and reused the function for both principal and subordinate.

cou/steps/plan.py
gabrielcocenza: It was not necessary, but because the way that the decorator works, sometimes we need to import and not use it. We are already not using it for the principal, so I agree that we can remove it.

cou/apps/ovn.py
aieri: but you don't necessarily have to use `self`. For example:
```
@staticmethod
def validate_support(version):
    if Version(version) < Version("22.03.0"):
        raise ApplicationError(
```

Then in `OvnPrincipalApplication.pre_upgrade_plan()`
```
for unit in units:
    self.validate_support(unit.workload_version)

```

And in `OvnSubordinateApplication.pre_upgrade_plan()`:
```
self.validate_support(self.status.workload_version)
```

tests/unit/apps/test_ovn.py
```diff
@@ -71,6 +71,20 @@ def test_ovn_workload_lesser_22(status, config, model):
         "ovn-central",
     )
 
+    with pytest.raises(ApplicationError, match=exp_error_msg_ovn_upgrade):
+        app_ovn_central.generate_upgrade_plan(target)
+
+
+def test_ovn_workload_lesser_22_subordinate(status, config, model):
```
aieri, line 78:
```suggestion
def test_ovn_workload_ver_lower_than_22_subordinate(status, config, model):
```

aieri (changes requested):

cou/apps/ovn.py
gabrielcocenza: Gotcha! This made the code cleaner indeed. Thanks for the suggestion.

I didn't use `staticmethod` because there is no real need to `OvnSubordinateApplication` inherit from `OvnPrincipalApplication`. I wrote the function in the same module to be shared on those classes.

aieri (dismissed):
Thank you, it looks simpler now

cou/apps/ovn.py
```diff
@@ -0,0 +1,77 @@
...
+from cou.apps.auxiliary import OpenStackAuxiliaryApplication
+from cou.apps.auxiliary_subordinate import OpenStackAuxiliarySubordinateApplication
+from cou.exceptions import ApplicationError
+from cou.steps import UpgradeStep
+from cou.utils.openstack import OpenStackRelease
+
+
+@AppFactory.register_application(["ovn-central", "ovn-dedicated-chassis"])
+class OvnPrincipalApplication(OpenStackAuxiliaryApplication):
+    """Ovn central application class."""
```
agileshaw, line 29:
```suggestion
    """Ovn principal application class."""
```

tests/unit/apps/test_ovn.py
```diff
@@ -0,0 +1,208 @@
...
+        "ovn-chassis",
+    )
+
+    with pytest.raises(ApplicationError, match=exp_error_msg_ovn_upgrade):
+        app_ovn_chassis.generate_upgrade_plan(target)
+
+
+def test_ovn_no_compatible_os_release(status, config, model):
+    ovn_central_status = status["ovn_central_ussuri_22"]
+    ovn_central_status.charm_channel = "55.7"
```
agileshaw, line 102: nit: since we are extending the range to include lower versions, can we test that also? So parametrize the `charm_channel` value to one lower than the lower bound (e.g. 19.03) and one higher than the upper bound (e.g. 55.7).

agileshaw (changes requested):
Two very small suggestions in-line. Otherwise LGTM,

agileshaw (approved):
LGTM, thanks!

aieri (approved):

PR #95, opened 2023-10-02
fix designate-bind, glance-simplestreams-sync and gnocchi plan generation

### LINKED ISSUE #45: Move channel getters/setters to OpenStackApplication superclass

### LINKED ISSUE #94: designate-bind is failing to generate upgrade plan
designate-bind uses the package `bind-9` to set the workload version and currently is failing with the following output:

```
Analyzing cloud... ✔
2023-09-29 23:11:58 [ERROR] Cannot generate upgrade plan for 'designate-bind': 'designate-bind' has unexpected channel: 'ussuri/stable' for the current workload version and OpenStack release: 'yoga'. Possible channels are: yoga/stable
Generating upgrade plan... ✖
2023-09-29 23:11:58 [ERROR] 'designate-bind' has unexpected channel: 'ussuri/stable' for the current workload version and OpenStack release: 'yoga'. Possible channels are: yoga/stable
```

gabrielcocenza:
- removed subordinate classes
- support for versionless applications
- added properties is_from_charm_store, is_subordinate, is_channel_based, is_versionless
- just generate workload upgrade plan for charms that have OpenStack origin setting: “source” or “openstack-origin”.

Closes #45,  #94, #104 and #106


gabrielcocenza:
Running cou with this branch on pc6b I discovered that the `OpenStackAlternativeApplication` used for designate-bind keeps generating upgrade plan even when the channel already changed to the OpenStack release target. This is probably because there is no `openstack-origin` or `source` to designate-bind charm and  apt_source_codename return None. I need to tweak a little bit the class and add unit tests for this particular case.

cou/apps/app.py
```diff
@@ -640,3 +640,46 @@ def _get_reached_expected_target_plan(
             function=self._check_upgrade,
             target=target,
         )
+
+
+@AppFactory.register_application(["designate-bind", "gnocchi"])
+class OpenStackChannelBasedApplication(OpenStackApplication):
```
aieri, line 646: would it be possible to modify the base openstack app to handle these special cases? For example by falling back to `self.channel_codename` if workload version is not available.

cou/apps/app.py
```diff
@@ -335,19 +418,17 @@ def apt_source_codename(self) -> Optional[OpenStackRelease]:
...
-            # get the OpenStack release from the channel track of the application.
-            os_track_release_channel = OpenStackRelease(self.channel.split("/", maxsplit=1)[0])
-        except ValueError:
-            logger.debug(
-                "The current channel of '%s' does not exist or is unexpectedly formatted",
-                self.name,
-            )
-            os_track_release_channel = self.current_os_release
-        return os_track_release_channel
+        self.channel = self.status.charm_channel
```
rgildein, line 426: Why we are configuring `channel` here? I think it should be in `post_init`. What do you say?

cou/apps/auxiliary.py
```diff
@@ -42,17 +44,9 @@ def possible_current_channels(self) -> list[str]:
...
-        )
-        if tracks:
-            return [f"{track}/stable" for track in tracks]
-
-        raise ApplicationError(
-            (
-                f"Cannot find a suitable '{self.charm}' charm channel for "
-                f"{self.current_os_release.codename}"
-            )
+            (self.charm, self.series, self.current_os_release.codename), []
```
rgildein, line 47: I think we should remove `:raises ApplicationError: When cannot find tracks.` since this will no longer raise any exception.

cou/apps/auxiliary.py
```diff
@@ -93,3 +87,41 @@ def channel_codename(self) -> OpenStackRelease:
...
+            track_from_channel = charm_channel.split("/")[0]
+            if track_from_channel not in tracks:
+                raise ValueError(
+                    (
+                        f"'{self.name}' cannot find a channel track for charm: '{charm_channel}' "
+                        f"series: {self.series} and OpenStack "
+                        f"release: {self.current_os_release.codename}"
+                    )
+                )
+            self._channel = charm_channel
```
rgildein, line 127: not-blocking: same as in previous one, I think it's ok to use return here
```suggestion
        if self.is_from_charm_store or self.is_subordinate:
            self._channel = charm_channel
            return

        tracks = OPENSTACK_TO_TRACK_MAPPING.get(
            (self.charm, self.series, self.current_os_release.codename), []
        )
        track_from_channel = charm_channel.split("/")[0]
        if track_from_channel not in tracks:
            raise ValueError(
                (
                    f"'{self.name}' cannot find a channel track for charm: '{charm_channel}' "
                    f"series: {self.series} and OpenStack "
                    f"release: {self.current_os_release.codename}"
                )
            )
        self._channel = charm_channel
```

rgildein (changes requested):

cou/apps/app.py
gabrielcocenza: Problem is that for versionless applications we end up calling `channel_codename` without setting the channel before. My next commit tries to have better comment of this issue. See what you think about it.

cou/apps/app.py
```diff
@@ -245,6 +224,73 @@ def __str__(self) -> str:
...
+            # some charms like glance-simplestreams-sync does not have workload_version.
+            elif not unit.workload_version:
+                self.units.append(
+                    ApplicationUnit(
+                        name=name,
+                        workload_version=unit.workload_version,
+                        os_version=self.channel_codename,
+                        machine=unit.machine,
+                    )
+                )
```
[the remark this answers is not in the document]
gabrielcocenza: I think this solution it's not explicit on what we decide and why we decided to use `self.channel_codename` for versionless applications. I've refactored, added a new property `is_versionless` and I think it's better now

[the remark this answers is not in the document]
rgildein: Let me check. Adding lot of properties will make code less readable. For me those two if conditions were just confusing and we can have two, but with proper description or more readable condition.

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
...
+            return self.channel_codename
+
+        try:
+            return max(
+                OpenStackCodenameLookup.find_compatible_versions(self.charm, unit.workload_version)
+            )
+        except ValueError as exc:
+            raise ApplicationError(
+                f"'{self.name}' with workload version {unit.workload_version} has no "
+                f"compatible OpenStack release."
```
rgildein, line 267: not-blocking: 
```suggestion
                "compatible OpenStack release."
```

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
             yaml.dump(summary, stream)
             return stream.getvalue()
 
+    def _populate_units(self) -> None:
+        """Populate application units."""
+        if not self.is_subordinate:
+            for name, unit in self.status.units.items():
+                compatible_os_version = (
+                    self._get_latest_compatible_os_version_by_unit_workload_version(unit)
```
rgildein, line 232: not-blocking: Is there chance to make name of this function shorter?

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
...
+            channel.
+        """
+        if (
+            self.is_from_charm_store
+            or (not self.is_os_channel_based and self.is_subordinate)
+            or self.is_valid_track
+        ):
+            self._channel = charm_channel
+            return
+        raise ValueError(self.channel_err_msg(charm_channel))
```
rgildein, line 296: not-blocking: 
```suggestion
        if not (
            self.is_from_charm_store
            or (not self.is_os_channel_based and self.is_subordinate)
            or self.is_valid_track
        ):
            raise ValueError(self.channel_err_msg(charm_channel))

        self._channel = charm_channel
```

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
...
+
+    def channel_err_msg(self, charm_channel: str) -> str:
+        """Error message when channel is not valid.
+
+        :param charm_channel: Charm channel.
+        :type charm_channel: str
+        :return: error message
+        :rtype: str
+        """
+        return f"'{self.name}' has invalid channel: '{charm_channel}'"
```
rgildein, line 306: To me this is unnecessary function used once, so instead of one line we add 10+unit test.

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
...
+
+    def _get_channels_based_on_os(self, os_release: OpenStackRelease) -> list[str]:
+        """Get the channels based on the OpenStack release.
+
+        :param os_release: OpenStack release.
+        :type os_release: OpenStackRelease
+        :return: List of possible channels compatible with a OpenStack release.
+        :rtype: list[str]
+        """
+        return [f"{os_release.codename}/stable"]
```
rgildein, line 341: Again this function makes just more room to error and does not same any time to use it, check my previous comment.

cou/apps/app.py
```diff
@@ -245,24 +224,121 @@ def __str__(self) -> str:
...
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
+        :raises ApplicationError: When cannot find a suitable target channel.
         :return: The next channel for the application. E.g: victoria/stable
         :rtype: str
         """
-        return f"{target.codename}/stable"
+        targets_channels = self._get_channels_based_on_os(target)
+        if targets_channels:
+            return targets_channels[-1]
```
rgildein, line 328: To my this part does not make since default `_get_channels_based_on_os` return list with one element, I know that in `OpenStackAuxiliaryApplication` you overwritten it, but why not override this function directly.

cou/apps/app.py
```diff
@@ -293,6 +379,67 @@ def current_os_release(self) -> OpenStackRelease:
...
+        :rtype: bool
+        """
+        if (
+            not self.origin_setting
+            or self.is_versionless
+            or self.is_subordinate
+            or self.charm == "gnocchi"
+        ):
+            return True
+        return False
```
rgildein, line 414:
```suggestion
        return (
            not self.origin_setting
            or self.is_versionless
            or self.is_subordinate
            or self.charm == "gnocchi"
        )
```

cou/apps/app.py
```diff
@@ -293,6 +379,67 @@ def current_os_release(self) -> OpenStackRelease:
...
+    @property
+    def is_versionless(self) -> bool:
+        """Check if the application is versionless.
+
+        Versionless applications are those that does not set a workload version.
+        E.g: glance-simplestreams-sync
+        :return: True if is versionless, False otherwise.
+        :rtype: bool
+        """
+        return not all((unit.workload_version for unit in self.status.units.values()))
```
rgildein, line 441:
```suggestion
        return not all(unit.workload_version for unit in self.status.units.values())
```

cou/apps/app.py
```diff
@@ -293,6 +379,67 @@ def current_os_release(self) -> OpenStackRelease:
...
+    def is_valid_track(self) -> bool:
+        """Check if the channel track is valid.
+
+        :return: True if valid, False otherwise.
+        :rtype: bool
+        """
+        if self.is_from_charm_store:
+            return True
+        try:
+            track = self.status.charm_channel.split("/", maxsplit=1)[0]
```
rgildein, line 426: Why we are using `self.status.charm_channel.split("/", maxsplit=1)[0]` and not `self.channel` as we did above? Also since we are adding so many properties, why not add also track property?

cou/apps/app.py
```diff
@@ -335,19 +482,29 @@ def apt_source_codename(self) -> Optional[OpenStackRelease]:
...
-            )
-            os_track_release_channel = self.current_os_release
-        return os_track_release_channel
+        if self.is_from_charm_store:
+            return OpenStackRelease("ussuri")
+
+        # applications with no workload version will call channel_codename without setting
+        # the channel before
+        if not hasattr(self, "channel"):
+            self.channel = self.status.charm_channel
```
rgildein, line 496: Why dis is not define at property, it really strange to set property value in getter for other property.

cou/apps/app.py
```diff
@@ -335,19 +482,29 @@ def apt_source_codename(self) -> Optional[OpenStackRelease]:
...
+        # get the OpenStack release from the channel track of the application.
+        return self._get_channel_codename
```
rgildein, line 498: One property calling another, why?

cou/apps/app.py
gabrielcocenza: Any suggestion?
maybe `_get_latest_os_version_by_workload_version`

cou/apps/app.py
gabrielcocenza: IMO using `not`make it less readable

cou/apps/app.py
gabrielcocenza: All the logic of the `@channel.setter` and channel property is the same for `OpenStackApplication` and `OpenStackAuxiliaryApplication` with the exception of this method. 
I just want to have a more clear error msg for the `AuxiliaryApplication` that depends on the `charm`, `series` and `os_release`.

In the end of the day I think it's a good deal to share the same logic on both classes and having this method to be more explicit in the error message.

cou/apps/app.py
gabrielcocenza: I don't agree. This function we are able to share `target_channel` and `possible_current_channels` on `OpenStackAuxiliaryApplication` and `OpenStackApplication`. It's one simple function that save us overwriting two other functions

cou/apps/app.py
gabrielcocenza: `is_valid_track` is used as a validation on `@channel.setter`, so we can't call self.channel because it won't exist. However, we can pass the `charm_channel` and leave as method instead of a property.

I thought about adding a track property. I think it's a good idea

cou/apps/app.py
rgildein: To me it's always more readable to do break, raise or return None/default value at the beginning of function. To separate actual logic of function. The `self._channel = charm_channel` is main purpose of this function and raising exception is corner case.
But it's only my point of view and it's not blocking. Resolving this comment.

cou/apps/app.py
rgildein: Still long I think, but better :+1:.

cou/apps/app.py
rgildein: To me this with all functions like this, it's just harder to understand what is happening in one class. It's harder to debug if something is not working, because logic is across multiple classes. It's good practice until it become unreadable.
I would rather change it in two places and be explicit, than jump from one class to another to get idea what is actually happening and especially when function is one line of code.

cou/apps/app.py
rgildein: calling `f"{os_release.codename}/stable"` is easier than calling `_get_channels_based_on_os`

cou/apps/app.py
gabrielcocenza: `channel_codename` can be called before setting `channel` during `_populate_units`. If we set the channel before populating, auxiliary charms will try to call `current_os_release` without populating previously. I don't have a better solution than this. It's a chicken and egg problem.

rgildein: How that is possible? The `channel` property will be there before any post_innit is executed or not ?

gabrielcocenza: 1. `_populate_units` is above channel setter
2. `_get_latest_os_version_by_workload_version` calls `channel_codename` if the charm is openstack channel based
3. `channel_codename` calls self.channel that it's not declared yet.

A possible  option is leaving units without an OpenStack version when is channel based

rgildein: We can ensure that `_populate_units` will set channel before calling `_get_latest_os_version_by_workload_version`. However you mention that `_populate_units` is above channel setter, but here you are checking if `not hasattr(self, "channel")` what means that you are checking if self has property channel, which should be true always or not?

cou/apps/app.py
```diff
@@ -293,6 +375,76 @@ def current_os_release(self) -> OpenStackRelease:
...
+        if self.is_from_charm_store:
+            return True
+        try:
+            OpenStackRelease(self._get_track_from_channel(charm_channel))
+            return True
+        except ValueError:
+            return False
+
+    @staticmethod
+    def _get_track_from_channel(charm_channel: str) -> str:
```
agileshaw, line 427: `CephMonApplication` also needs to get the track from channel for `_get_change_require_osd_release_plan` method. This is probably an indicator that we need to split the PR into multiple more-focused ones.

cou/apps/app.py
```diff
@@ -293,6 +375,76 @@ def current_os_release(self) -> OpenStackRelease:
...
+        :return: True if subordinate, False otherwise.
+        :rtype: bool
+        """
+        return bool(self.status.subordinate_to)
+
+    @property
+    def is_os_channel_based(self) -> bool:
+        """Check if application is channel based.
+
+        :return: True if does have origin setting, False otherwise.
```
agileshaw, line 400: This description doesn't match what we are checking. Could you update it?

cou/apps/auxiliary.py
```diff
@@ -26,70 +26,77 @@
...
-        :return: OpenStackRelease object
+        :return: The OpenStack release codename based on the channel.
         :rtype: OpenStackRelease
         """
-        track: str = self.channel.split("/", maxsplit=1)[0]
+        track: str = self._get_track_from_channel(self.channel)
         compatible_os_releases = TRACK_TO_OPENSTACK_MAPPING.get((self.charm, self.series, track))
-        if compatible_os_releases:
-            return max(compatible_os_releases)
+        # channel setter already validate if it is a valid channel.
```
agileshaw, line 61:
```suggestion
        # channel setter already validated if it is a valid channel.
```

agileshaw (changes requested):
Aside from fixing bugs for designate-bind, glance-simplestreams-sync and gnocchi, this PR refactors applications classes in a great degree. It is easy to miss something in implementation (as mentioned in my in-line comment about "CephMonApplication" class) when trying to do many different things in a single PR, so I agree with Robert, we should split this.

cou/apps/app.py
rgildein: Completely agree on this.

cou/apps/auxiliary.py
```diff
@@ -26,70 +26,77 @@
 logger = logging.getLogger(__name__)
 
 
+# NOTE (gabrielcocenza) ceph-osd might need to be removed if data plane has a specific class
 @AppFactory.register_application(
     ["rabbitmq-server", "vault", "mysql-innodb-cluster", "ceph-fs", "ceph-radosgw"]
     + CHARM_FAMILIES["ovn"]
+    + AUXILIARY_SUBORDINATES
+    + ["ceph-osd"]
```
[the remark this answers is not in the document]
gabrielcocenza: The problem that I found in partner cloud is with the channel setters, `ceph-osd` will be instantiated as normal OpenStackApplication and the channel will be rejected and won't be possible to generate a plan. I can try to think in a alternative solution

PR #98, opened 2023-10-04
fix broken lint because of docstrings.

gabrielcocenza:
Currently lint is broken because merges happened without rebasing #85


agileshaw (approved):

rgildein (approved):
LGTM

Issue #110, opened 2023-10-09
Running `cou plan --parallel` do not make sense


rgildein:
The `--parallel` option is available for the `plan' command, which makes no sense.

```bash
$ cou plan --help            
Usage: cou plan [options]

Show the steps COU will take to upgrade the cloud to the next release.

Options:
  -h, --help          Show this help message and exit.
  --model MODEL_NAME  Set the model to operate on.
                      If unset, the model name will be determined by inspecting the environment as follows:
                        1 - Environment variable JUJU_MODEL,  2 - Environment variable MODEL_NAME,  3 - Current active juju model
  --parallel          Run upgrade steps in parallel where possible.
  --verbose, -v       Increase logging verbosity in STDOUT. Multiple 'v's yield progressively more detail (up to 4).
                      Note that this doesn't affect the verbosity in logfile, which will always have the maximum verbosity.
  --quiet, -q         Disable output in STDOUT. 
```

Even the description is suggesting that this will be used while running upgrade steps.

agileshaw:
I don't agree that `cou plan --parallel` doesn't make sense because in data-plane stage the plan COU generates should contain the information on how the parallelization groups are formed when passing the flag. But, as we concluded in earlier discussion today, we are going to remove `--parallel` option all together and make parallel execution a default, so plan should contain that grouping information by default as well.

rgildein:
I think that even if we left the parameter `--parallel` that should not change any results from Analyze or plan, since it's relevant only for running upgrade.

agileshaw:
It is relevant to the plan phase because we can either plan for serial upgrade or parallel upgrade (with different upgrade step layout), especially with data-plane components. But since we are going to make parallel the default, this discussion is out-of-scope here. We can talk about UI design for plan result when we are ready for data-plane upgrades.

PR #111, opened 2023-10-09
drop parallel for `cou plan`

rgildein:
Dropping option `parallel` for `plan` command.

Fixes: #110

### DIFF
--- a/completions/bash/cou_bash_completion
+++ b/completions/bash/cou_bash_completion
@@ -2,7 +2,7 @@ _subcommand_args() {
     local cmd=$1
     case "${cmd}" in
         plan)
-            opts="--help --quiet --verbose --parallel --model"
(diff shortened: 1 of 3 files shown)


agileshaw (changes requested):
As we discussed today, since we should consider unit level parallel execution a default (automatically for control-plane components and grouped by AZ for data-plane components), we can drop `--parallel` option all together.

rgildein:
Right now we are not ready for this, since it will required actually changing how we run UpgradeSteps. I think that we can merge this first and open another issue to drop it also from `run`.

agileshaw (approved):
For the moment, I'm ok with removing parallel feature for plan only. Please open an issue to track dropping it everywhere.

gabrielcocenza (approved):
LGTM

Issue #100, opened 2023-10-04
COU doesn't wait for the completion of a step before allowing the user to continue


gabrielcocenza:
Currently COU it's not properly checking if the application and/or the model is active idle before running the next step.

Issue #78, opened 2023-09-21
COU should halt when the model it's not active idle 


gabrielcocenza:
Currently COU doesn't check if the model is blocked or on error state. We should check it first before analyzing or generating the upgrade plan

rgildein:
We do not need to check only relevant (data and control plane) applications and not whole model.

PR #113, opened 2023-10-11
waiting for model or application to be idle

rgildein:
- add waiting to model to become idle for RabbitMQ server, ceph-mon and Keystone
- add waiting to app itself for other apps

fixes: #100


cou/apps/auxiliary.py
```diff
@@ -93,3 +94,29 @@ def channel_codename(self) -> OpenStackRelease:
...
+    def post_upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
+        """Adding wait to post upgrade step.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: Plan that will add pre upgrade as sub steps.
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        wait = UpgradeStep(
+            description=f"Check if the workload of '{self.name}' has been upgraded",
```
aieri, line 116: this needs to be updated, it's the same output of `_get_reached_expected_target_plan()`

rgildein: Yes, I copy pasted it and forgot to change it. Thanks

cou/apps/auxiliary.py
```diff
@@ -93,3 +94,49 @@ def channel_codename(self) -> OpenStackRelease:
...
+        :rtype: list[Optional[UpgradeStep]]
+        """
+        return [
+            self._get_wait_step(300, wait_for_itself=False),
+            self._get_reached_expected_target_plan(target),
+        ]
+
+
+@AppFactory.register_application(["keystone"])
+class Keystone(OpenStackApplication):
```
gabrielcocenza, line 123: Doesn't make sense having Keystone on the auxiliary module

gabrielcocenza (changes requested):

cou/apps/app.py
```diff
@@ -640,3 +645,28 @@ def _get_reached_expected_target_plan(
             function=self._check_upgrade,
             target=target,
         )
+
+    def _get_wait_step(self, timeout: int, wait_for_itself: bool = True) -> UpgradeStep:
```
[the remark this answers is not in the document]
rgildein: Agree, I'll change it.

cou/apps/app.py
```diff
@@ -437,12 +437,17 @@ def upgrade_plan(self, target: OpenStackRelease) -> list[Optional[UpgradeStep]]:
...
+        Wait until the application reaches the idle state and then check the target workload.
+
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
         :return: Plan that will add post upgrade as sub steps.
         :rtype: list[Optional[UpgradeStep]]
         """
-        return [self._get_reached_expected_target_plan(target)]
+        return [
+            self._get_wait_step(120),
```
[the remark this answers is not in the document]
rgildein: This should be defined as constant, but I was not sure about the right number. Maybe we can use same value for both model or only app.
I will not crate function, but rather constant only. Plus I do not see any benefit to hind logic for keystone, ceph-mon or rebitmq to `OpenStackApplication`, since we are defining new class for it.

cou/apps/auxiliary.py
rgildein: I do not understand what is auxiliary and why we have ceph.py where we are using OpenStackAuxiliaryApplication. Do you have any suggestion? To my we should have base.py, auxiliary.py, principal.py and subordinate.py. What do you say? Maybe some other suggestion?

gabrielcocenza: Auxiliary charms are those that makes part of OpenStack project, but are not directly connected with the o7k releases like rabbitmq, ceph, mysql and etc. On those charms we have 1:N mapping where one workload version can match with multiple OpenStack releases. On the other hand, OpenStack applications have a 1:1 mapping where a workload version matches with just one OpenStack release.

#95 tries to destroy the need of new classes for subordinates and I think it's more maintainable the project if we have fewer classes.

rgildein:
As we discussed I will propose small refactor with `core.py` and `facotry.py`, and I will move application to proper Python files.

tests/unit/steps/test_steps_plan.py
```diff
@@ -34,6 +34,24 @@ def generate_expected_upgrade_plan_principal(app, target, model):
         parallel=False,
         function=None,
     )
+    if app.charm in ["rabbitmq-server", "ceph-mon", "keystone"]:
```
gabrielcocenza, line 37: maybe adding a global variable to this list is going to be more clear

cou/apps/auxiliary.py
```diff
@@ -93,3 +92,27 @@ def channel_codename(self) -> OpenStackRelease:
...
+    RabbitMQ had to wait until the entire model will be idle after the upgrade to ensure other
+    upgrade can continue.
+    """
+
+    wait_timeout = 300
+    wait_for_model = True
+
+
+@AppFactory.register_application(["keystone"])
+class Keystone(OpenStackApplication):
```
gabrielcocenza, line 110: Please move `Keystone` to `app.py`. This is not an auxiliary charm

gabrielcocenza (changes requested):

agileshaw (commented):
Overall looks good. Please merge the upstream main branch and sort the custom application classes as appropriate

cou/apps/app.py
```diff
@@ -640,3 +650,28 @@ def _get_reached_expected_target_plan(
...
+        if self.wait_for_model:
+            description = (
+                f"Wait {self.wait_timeout} s for model {self.model.name} to reach the idle state."
+            )
+            apps = None
+        else:
+            description = (
+                f"Wait {self.wait_timeout} s for app {self.name} to reach the idle state."
+            )
+            apps = [self.name]
```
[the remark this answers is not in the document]
rgildein: I was doing it like that, but I hit
```bash
cou/apps/app.py:669: error: Incompatible types in assignment (expression has type "None", variable has type "list[str]")  [assignment]
```
so made `else`. But what I can do is `apps: Optional[list[str]] = [self.name]` :+1:

tests/unit/steps/test_steps_plan.py
rgildein: These are only tests and I want to refactor it as I mentioned before to have proper fixtures. But I agree and I will create constant for it.

cou/apps/auxiliary.py
rgildein: after #115 is merged.

rgildein: I must move core.py to base.py to be able to move Keystone to coro.py, otherwise there will be circular import.

rgildein:
Right now this PR is already too big, so I did not introduced the waiting for all relevant applications instead of whole model. This will be handled by fixing the #78.

cou/apps/auxiliary.py
```diff
@@ -95,10 +95,25 @@ def channel_codename(self) -> OpenStackRelease:
         )
 
 
+@AppFactory.register_application(["rabbitmq-server"])
+class RabbitMQServer(OpenStackAuxiliaryApplication):
+    """RabbitMQ application.
+
+    RabbitMQ had to wait until the entire model will be idle after the upgrade to ensure other
+    upgrade can continue.
```
aieri, line 103: Maybe
```suggestion
    RabbitMQ must wait for the entire model to be idle before declaring the upgrade complete.
```

aieri (commented):
Overall lgtm, I have a question though

cou/apps/core.py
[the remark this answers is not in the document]
rgildein: It can includes `OpenStackAppplication`, if we refactor it to same real base class or mixins. I was voting to have `principal.py`, but @gabrielcocenza did not agreed and provide goos reason why we should go with core.

aieri (dismissed):

gabrielcocenza (approved):

aieri (approved):

gabrielcocenza (approved):

Issue #103, opened 2023-10-05
Catching Timeout error and KeyboardInterrupt


rgildein:
If control+c is pressed the result should not be showing any traceback.  If it's not possible to connect to model, the error should provide more user friendly output. 

Example with `TimeoutError`:
```bash
x1:➜  charmed-openstack-upgrader git:(tmp/testing) ✗ cou plan --model openstack
Configuring logging... ✖
2023-10-05 22:05:48 [ERROR] function _connect timed out after 60s                                                                  [1m5.76s]
```

Example with `KeyboardInterrupt`:
```bash
x1:➜  charmed-openstack-upgrader git:(tmp/testing) ✗ cou plan --model openstack 
^CTraceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 1022, in _find_and_load
KeyboardInterrupt
thread '<unnamed>' panicked at 'Python API call failed', /github/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pyo3-0.18.3/src/err/mod.rs:790:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/snap/charmed-openstack-upgrader/x1/bin/cou", line 5, in <module>
    from cou import main
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/__init__.py", line 15, in <module>
    from cou.__main__ import main
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/__main__.py", line 18, in <module>
    from cou import cli
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/cli.py", line 25, in <module>
    from cou.exceptions import COUException
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/exceptions.py", line 17, in <module>
    from juju.action import Action
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/juju/action.py", line 4, in <module>
    from . import model
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/juju/model.py", line 25, in <module>
    from . import provisioner, tag, utils, jasyncio
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/juju/provisioner.py", line 11, in <module>
    import paramiko
[...]

PR #114, opened 2023-10-16
introduce cancel to `UpgradeStep`

rgildein:
Introduce cancel to `UpgradeStep`, which will cancel all sub-steps. The `cancel` can by run with `safe=True` (default value), which will keep all pending task running and prevent new one to be triggered. The `safe=False` will cancel also all pending `asyncio.Task` (running), which can cause stopping in middle of step and break cloud.

With these changes I also used coroutine as input of `UpgradeStep` instead of function and parameters. This helps with catching calling some coroutine with wrong arguments, e.k. linting cat catch it. However this change required to add warning ignore, since coroutines can be un-awaited in case of only planning.

FYI: We could not define `UpgradeStep` as `asyncio.Task` , since `asyncio.Task` is scheduled to be run as soon as possible if it's run in loop already. We can use it only if whole planning will be not asynchronous. 

related: #103 (preparation for proper handling of `KeyboardInterrupt`


tests/unit/steps/test_steps.py
```diff
@@ -13,104 +13,246 @@
...
+    """Test UpgradeStep adding sub steps."""
+    exp_sub_steps = 3
+    plan = UpgradeStep(description="plan")
+    for i in range(exp_sub_steps):
+        plan.add_step(UpgradeStep(description=f"sub-step-{i}"))
+
+    assert len(plan.sub_steps) == exp_sub_steps
+
+
+def test_step_cancel():
```
gabrielcocenza, line 134:
```suggestion
def test_step_cancel_safe():
```

tests/unit/steps/test_steps.py
```diff
@@ -13,104 +13,246 @@
...
+    sub_steps[0].sub_steps = [UpgradeStep(description=f"sub-0.{i}") for i in range(3)]
+
+    plan.cancel()
+
+    assert plan.canceled is True
+    assert all(step.canceled is True for step in sub_steps)
+    assert all(step.canceled for step in sub_steps[0].sub_steps)
+
+
+def test_step_unsafe_cancel():
```
gabrielcocenza, line 148:
```suggestion
def test_step_cancel_unsafe():
```

tests/unit/steps/test_steps.py
```diff
@@ -13,104 +13,246 @@
...
+        ),
     ],
 )
-def test_upgrade_step__eq__(description, function, parallel, expected_result):
-    upgrade_step = UpgradeStep(description="test", function=None, parallel=False)
-    other_upgrade_step = UpgradeStep(description=description, function=function, parallel=parallel)
-    assert (upgrade_step == other_upgrade_step) is expected_result
+async def test_step_full_run(sub_steps, exp_order, parallel):
+    """Test to simulate running full plan with steps."""
+    # warnings.filterwarnings("ignore", message="coroutine '.*' was never awaited")
```
gabrielcocenza, line 232: is this comment missing to remove?

tests/unit/steps/test_steps.py
```diff
@@ -13,104 +13,246 @@
...
+    assert hash(("test hash", False, coro)) == hash(step)
+
+
+@pytest.mark.parametrize(
+    "description, parallel, args",
+    [("test", False, ()), ("test description", True, ("name", 1, 2))],
+)
+def test_step_eq(description, parallel, args):
+    """Test UpgradeStep comparison."""
+    # coro = mock_coro(*args)
```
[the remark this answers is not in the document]
rgildein: Thanks, I was debugging the warning errors.

tests/unit/steps/test_steps.py
```diff
@@ -13,104 +13,246 @@
...
+    step_1 = UpgradeStep(description, parallel, mock_coro(*args))
+    step_2 = UpgradeStep(description, parallel, mock_coro(*args))
+    # define step with different coro
+    step_3 = UpgradeStep(description, parallel, mock_coro(unique_arg=True))
+
+    assert step_1 == step_2
+    assert step_1 != step_3
+    assert step_1 != 1
```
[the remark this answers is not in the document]
rgildein: Trying only if `return NotImplemented` is working fine if `not isinstance(other, UpgradeStep)`.

tests/unit/steps/test_steps.py
rgildein: Yes again, thanks.

rgildein:
> I personally had a bad time trying to understand what is the `_task` introduced in the UpgradeStep and I wish it had more explanation for future maintainability.

Do you have any suggestion?
The main reason is that instead awaiting the `core` we create task which can be canceled later + we are defining name of for task so we can have watcher later checking if task is still running or something like that.

cou/steps/__init__.py
```diff
@@ -112,3 +125,64 @@ def __repr__(self) -> str:
...
+
+        self._canceled = True
+        logger.debug("canceled: %s", self)
+
+    async def run(self) -> Any:
+        """Run the UpgradeStep coroutine.
+
+        :return: Result of the coroutine.
+        :rtype: Any
+        :raises CanceledUpgradeStep: If step has already been cancelled.
```
agileshaw, line 174:
```suggestion
        :raises CanceledUpgradeStep: If step has already been canceled.
```

agileshaw (changes requested):
Just two small suggestions inline. Overall I like the new approach and am looking forward to seeing everything comes together.

gabrielcocenza (approved):

agileshaw (approved):
LGTM

agileshaw (approved):

gabrielcocenza (approved):

Issue #105, opened 2023-10-06
COU should warn that Data Plane apps need to upgrade manually


gabrielcocenza:
After upgrading all control plane apps, COU need to have a way saying to the user that the data plane needs to be upgraded manually, otherwise it's hard to understand why it doesn't try to upgrade to the next OpenStack release.

PR #117, opened 2023-10-16
remove data plane charm support.

gabrielcocenza:
- with this removal, data-plane charms won't be instantiated as OpenStack applications.

### DIFF
--- a/cou/utils/openstack.py
+++ b/cou/utils/openstack.py
@@ -417,13 +417,15 @@ class OpenStackCodenameLookup:
 def is_charm_supported(charm: str) -> bool:
     """Check if a charm upgrade is supported.
 
+    Currently data plane apps are not supported.
     :param charm: Name of the charm.
     :type charm: str
     :return: True if supported, else False
     :rtype: bool
     """
     return (
-        bool(OpenStackCodenameLookup.lookup(charm))
+        charm not in DATA_PLANE_CHARMS
+        and bool(OpenStackCodenameLookup.lookup(charm))
         or charm in SUBORDINATES + AUXILIARY_SUBORDINATES
     )
 
--- a/tests/unit/steps/test_steps_analyze.py
+++ b/tests/unit/steps/test_steps_analyze.py
@@ -87,9 +87,10 @@ async def test_populate_model(full_status, config, model):
     model.get_status = AsyncMock(return_value=full_status)
     model.get_application_config = AsyncMock(return_value=config["openstack_ussuri"])
 
-    # Initially, 6 applications are in the status: keystone, cinder, rabbitmq-serve, my-app,
+    # Initially, 6 applications are in the status: keystone, cinder, rabbitmq-server, my-app,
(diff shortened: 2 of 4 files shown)


cou/utils/openstack.py
```diff
@@ -417,13 +417,15 @@ def lookup(cls, charm: str) -> dict:
...
 
+    Currently data plane apps are not supported.
     :param charm: Name of the charm.
     :type charm: str
     :return: True if supported, else False
     :rtype: bool
     """
     return (
-        bool(OpenStackCodenameLookup.lookup(charm))
+        charm not in DATA_PLANE_CHARMS
```
rgildein, line 427: Caw we add warning logging that charm is not supported?

rgildein: Like this we will raise and exception if charm `nova-compute` is not supported, but can we provide list of unsupported charms instead?

rgildein: Ah ok so there will no exception, in that is ok. But It would be nice to add warning logs for data plane charms. So user will see that these are not supported, not for each unsupported charm but only for data plane charms. I think it would be as good feedback to user.

gabrielcocenza: I think this work should be scope of another PR that address the issue of #105

rgildein: ACK

rgildein (dismissed):
LGTM

tests/unit/steps/test_steps_analyze.py
```diff
@@ -87,9 +87,10 @@ async def test_populate_model(full_status, config, model):
     model.get_status = AsyncMock(return_value=full_status)
     model.get_application_config = AsyncMock(return_value=config["openstack_ussuri"])
 
-    # Initially, 4 applications are in the status: keystone, cinder, rabbitmq-server and my-app
-    # my-app it's not on the lookup and won't be instantiated.
-    assert len(full_status.applications) == 4
+    # Initially, 6 applications are in the status: keystone, cinder, rabbitmq-serve, my-app,
```
agileshaw, line 90:
```suggestion
    # Initially, 6 applications are in the status: keystone, cinder, rabbitmq-server, my-app,
```

tests/unit/steps/test_steps_analyze.py
```diff
@@ -87,9 +87,10 @@ async def test_populate_model(full_status, config, model):
...
+    # ceph-osd and nova-compute. my-app it's not on the lookup table, ceph-osd and nova-compute
+    # are data plane (not supported yet) and because of that they won't be instantiated.
```
agileshaw, line 92:
```suggestion
    # are data plane applications (not supported yet) and because of that they won't be instantiated.
```

agileshaw (changes requested):
Overall LGTM. Just some re-wording suggestions.

agileshaw (approved):
LGTM

rgildein (approved):
LGTM

PR #119, opened 2023-10-17
Add additional compatible  OVN version in openstack_to_track.csv

agileshaw:
Track version will be accessed in `current_os_release` in Analyze stage, so we need to recognize 20.x and 21.x as compatible (but not supporting) OVN versions.

This PR also removes the incorrect OVN charm family registration for `OpenStackAuxiliaryApplication` class since they have their own subclasses.


gabrielcocenza (approved):
LGTM. Thanks

aieri (approved):

PR #120, opened 2023-10-18
Refactor cloud os release analysis

agileshaw:
- Delegate the responsibility of evaluating target cloud os release to the Analyze stage, alongside the existing current_cloud_os_release method
- Change the return type of `OpenStackRelease.next_release` and `OpenStackRelease.previous_release` from `str` to an `OpenStackRelease` object. This unifies the target release variable type across the project.

### DIFF
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -156,8 +156,6 @@ class Analysis:
     def current_cloud_os_release(self) -> Optional[OpenStackRelease]:
         """Shows the current OpenStack release codename.
 
-        This property just consider OpenStack charms as those that have
-        openstack-origin or source on the charm configuration (app.os_origin).
         :return: OpenStack release codename
         :rtype: Optional[OpenStackRelease]
         """
@@ -170,8 +168,6 @@ class Analysis:
     def next_cloud_os_release(self) -> Optional[OpenStackRelease]:
         """Shows the next OpenStack release codename.
 
-        This property just consider OpenStack charms as those that have
-        openstack-origin or source on the charm configuration (app.os_origin).
         :return: OpenStack release codename
         :rtype: Optional[OpenStackRelease]
         """
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -148,6 +148,8 @@ class Analysis:
             + "\n".join([str(app) for app in self.apps_control_plane])
             + "Data Plane:\n"
             + "\n".join([str(app) for app in self.apps_data_plane])
+            + f"\nCurrent Cloud OpenStack Release: {self.current_cloud_os_release}\n"
+            + f"Next Cloud OpenStack Release: {self.next_cloud_os_release}\n"
         )
 
     @property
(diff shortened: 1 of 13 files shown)


tests/unit/utils/test_openstack.py
```diff
@@ -269,7 +272,10 @@ def test_determine_next_openstack_release(os_release, release_year, next_os_rele
 )
 def test_determine_previous_openstack_release(os_release, previous_os_release):
     release = OpenStackRelease(os_release)
-    assert release.previous_release == previous_os_release
+    if previous_os_release is None:
```
gabrielcocenza, line 275: I prefer like the way I described above, but it's not important

cou/steps/analyze.py
```diff
@@ -157,9 +159,22 @@ def current_cloud_os_release(self) -> Optional[OpenStackRelease]:
...
             (app.current_os_release for app in self.apps_control_plane + self.apps_data_plane),
             default=None,
         )
+
+    @property
+    def next_cloud_os_release(self) -> Optional[OpenStackRelease]:
+        """Shows the next OpenStack release codename.
+
+        This property just consider OpenStack charms as those that have
+        openstack-origin or source on the charm configuration (app.os_origin).
```
gabrielcocenza, line 174: This description it's not true anymore. We are checking openstack releases by channel on subordinates charms and by workload version on auxiliary charms. Please remove on `current_cloud_os_release` and `next_cloud_os_release`

gabrielcocenza (changes requested):

cou/utils/openstack.py
```diff
@@ -240,20 +240,20 @@ def codename(self, value: str) -> None:
             raise ValueError(f"OpenStack '{value}' is not in '{self.openstack_codenames}'")
 
     @property
-    def next_release(self) -> Optional[str]:
+    def next_release(self) -> Optional["OpenStackRelease"]:
```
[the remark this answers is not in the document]
gabrielcocenza: It seems that in your next patch you didn't update the return type. I hope the lint is able to get it.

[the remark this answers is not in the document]
agileshaw: Hmm no, the lint didn't catch it when I was running it locally, not even when I was trying to change the return type in docstring to something random. Strange. But I just updated the docstring.

cou/utils/openstack.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 """Lookup utils to determine compatible OpenStack codenames for a given component."""
+from __future__ import annotations
```
rgildein, line 16: Why is this needed?

tests/unit/steps/test_steps_analyze.py
```diff
@@ -133,6 +136,19 @@ async def test_analysis_detect_current_cloud_os_release_same_release(apps, model
 
     # current_cloud_os_release takes the minimum OpenStack version
     assert result.current_cloud_os_release == "ussuri"
+    assert result.next_cloud_os_release == "victoria"
+
+
+@pytest.mark.asyncio
+async def test_analysis_detect_current_and_next_cloud_os_release_empty_cloud(model):
```
rgildein, line 143: not-blocking: I think that using docstring in test function make are more readable than long name of function.

rgildein (approved):
LGTM, I have one not-blocking question

cou/utils/openstack.py
agileshaw: As suggested by Gabriel in [this comment](https://github.com/canonical/charmed-openstack-upgrader/pull/120#discussion_r1363833076), this is needed for referring the current class in type annotation.

cou/utils/openstack.py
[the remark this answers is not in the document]
rgildein: What if benefit to use `from __future__ import annotations` instead of `"OpenStackRelease"`? I always using it with quotes, e.g. ``"OpenStackRelease"``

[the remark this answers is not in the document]
gabrielcocenza: using `OpenStackRelease` lint was able to find that the docstring was not updated.

agileshaw:
As discussed elsewhere, we decide to keep the evaluation of next release in the plan stage. 

Other changes of this PR (e.g. changing the return type of OpenStackRelease.next_release and OpenStackRelease.previous_release from str to an OpenStackRelease object) will be proposed in a different PR.

Issue #108, opened 2023-10-07
Pre-upgrade package refresh causes mysql outage


aieri:
The pre-upgrade package refresh runs in parallel on all units of a given application. If mysql-innodb-cluster has updates available for package `mysql-server-core-8.0`, all `mysqld` processes may be restarted at roughly the same time, making the cluster lose quorum.

The ideal solution would be to carefully orchestrate the upgrade one unit at a time, only proceeding with the next unit when full cluster redundancy has been re-established.

As an interim solution to avoid this issue we could simply skip upgrading that package, possibly by applying (and then removing) a hold.

PR #121, opened 2023-10-18
Implement package on hold during software upgrades

agileshaw:
Allow specifying a list of packages to put on hold during software upgrades. This is useful for charms like mysql-innodb-cluster to prevent undesired service outage caused by package refreshes.

fixes: #108


cou/apps/auxiliary.py
```diff
@@ -91,3 +91,33 @@ def channel_codename(self) -> OpenStackRelease:
...
+        :return: Plan for upgrading software packages to the latest of the current release.
+        :rtype: UpgradeStep
+        """
+        description = (
+            f"Upgrade software packages of '{self.name}' from the current APT repositories"
+        )
+
+        # NOTE(agileshaw): holding 'mysql-server-core-8.0' package prevents undesired
+        # mysqld processes from restarting, which lead to outages
+        packages_to_hold = ["mysql-server-core-8.0"]
```
gabrielcocenza, line 114: I'm wondering if this will happen on other applications. If this is likely to happen, I would prefer having `packages_to_hold` as a class attribute and not overriding the `_get_upgrade_current_release_packages_plan`.

gabrielcocenza (commented):
Overall LGTM, but I left one point to reflect.

cou/apps/auxiliary.py
agileshaw: So far this is the only application we need to do packages holding. However, I do want to make it as generic as possible for future proofing. Thanks for the suggestion, I've updated the code.

cou/apps/app.py
```diff
@@ -495,6 +496,7 @@ def _get_upgrade_current_release_packages_plan(self, parallel: bool = False) ->
             function=upgrade_packages,
             units=self.status.units.keys(),
             model=self.model,
+            packages_to_hold=self.packages_to_hold,
```
gabrielcocenza, line 499: nit: Do we really need this line?

agileshaw: I don't understand. How else can we share the same `_get_upgrade_current_release_packages_plan` in subclasses when `self.packages_to_hold` actually contains something?

gabrielcocenza (dismissed):
LGTM

cou/apps/app.py
agileshaw: Oh haha :laughing: The code collapsing in diff can be a curse sometimes.

rgildein: Same here, I was confused why it's in post_init.

rgildein (approved):
LGTM

cou/utils/app_utils.py
```diff
@@ -24,17 +25,24 @@
...
     :type model: COUModel
+    :param packages_to_hold: A list of packages to put on hold during package upgrade.
+    :type packages_to_hold: Optional[list]
     :raises RunUpgradeError: When an upgrade fails.
     """
     dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
     command = f"apt-get update && apt-get dist-upgrade {dpkg_opts} -y && apt-get autoremove -y"
+    if packages_to_hold:
+        packages = " ".join(packages_to_hold)
+        command = f"sudo apt-mark hold {packages} && {command} && sudo apt-mark unhold {packages}"
```
aieri, line 45: What if the upgrade fails, but the hold does not? You could consider splitting the three actions (hold/upgrade/unhold) to more easily respond to failures, or at least turn the last `&&` into an or to let the unhold command run in (almost) any case

agileshaw: Thanks for the suggestion.

Running commands 3 times would not be ideal because we would need to establish connection to a unit multiple times. Since running `unhold` on non-hold package is safe, I chose to implement the second option.

rgildein (approved):
LGTM

cou/utils/app_utils.py
rgildein: This is really good catch and I think that second approach is better, since connecting to unit multiple times can provides simply more room to failure.

gabrielcocenza (approved):

cou/utils/app_utils.py
```diff
@@ -25,17 +26,24 @@
...
     :type model: COUModel
+    :param packages_to_hold: A list of packages to put on hold during package upgrade.
+    :type packages_to_hold: Optional[list]
     :raises RunUpgradeError: When an upgrade fails.
     """
     dpkg_opts = "-o Dpkg::Options::=--force-confnew -o Dpkg::Options::=--force-confdef"
     command = f"apt-get update && apt-get dist-upgrade {dpkg_opts} -y && apt-get autoremove -y"
+    if packages_to_hold:
+        packages = " ".join(packages_to_hold)
+        command = f"sudo apt-mark hold {packages} && {command} ; sudo apt-mark unhold {packages}"
```
gabrielcocenza: If we run this command as root, I think we should remove `sudo`

agileshaw: Done

gabrielcocenza (dismissed):

aieri (approved):

cou/utils/app_utils.py
```diff
@@ -25,17 +26,24 @@
 logger = logging.getLogger(__name__)
 
 
-async def upgrade_packages(units: Iterable[str], model: COUModel) -> None:
+async def upgrade_packages(
+    units: Iterable[str], model: COUModel, packages_to_hold: Optional[list] = None
```
gabrielcocenza, line 30: maybe it's safer if we remove the default value to `None` so we force to always pass `packages_to_hold`?

agileshaw: As discussed elsewhere, since we always need to specify `packages_to_hold` when using this function in Application classes for inheritance purposes, I removed its default value here.

gabrielcocenza (approved):

aieri (approved):

Issue #122, opened 2023-10-19
Require OSD release should happen as a post upgrade of ceph-osd and not ceph-mon


gabrielcocenza:
Currently, the COU is running `_get_change_require_osd_release_plan` as a post upgrade of ceph-mon, but this step should happen after upgrading ceph-osd (data plane).

Running before osd upgrade this issue happen:
```shell
sudo ceph osd require-osd-release pacific
Error EPERM: not all up OSDs have CEPH_FEATURE_SERVER_PACIFIC feature
```

agileshaw:
After discussion will remove the `_get_change_require_osd_release_plan` from post_upgrade (and keep it in ceph-mon's pre_upgrade), and check osd version when setting the release

PR #130, opened 2023-10-20
Remove setting require_osd_release in post_upgrade step of ceph-mon

agileshaw:
- Require OSD release should happen as a post upgrade of ceph-osd and not ceph-mon. Therefore removed this step from the post-upgrade plan of CephMonApplication class (but keeping the one in pre-upgrade plan)
- Instead of comparing require_osd_release value with ceph-mon release, we compare it with the actual release of OSDs. This information is parsed from `ceph versions` command output running on a ceph-mon unit.

fixes: https://github.com/canonical/charmed-openstack-upgrader/issues/122


cou/utils/app_utils.py
```diff
@@ -126,3 +109,101 @@ def validate_ovn_support(version: str) -> None:
                 "ovn-upgrade-2203.html"
             )
         )
+
+
+# Private functions
+async def _get_required_osd_release(unit: str, model: COUModel) -> str:
```
gabrielcocenza, line 115: Both `_get_required_osd_release` and ` _get_current_osd_release` they share the same structure:

1. run a command
2. check if the command returned 0
3. raise RunUpgradeError if not 0
4. raise  RunUpgradeError if JujuError happens.

Can't we share a single generic function that does this job?

cou/utils/app_utils.py
```diff
@@ -126,3 +109,101 @@ def validate_ovn_support(version: str) -> None:
...
+
+    The release of OSDs is parsed from the output of running `ceph versions` command
+    on a ceph-mon unit.
+    :param unit: The ceph-mon unit name where the check command runs on.
+    :type unit: str
+    :param model: COUModel object
+    :type model: COUModel
+    :return: True if the value of require-osd-release is the same as the release of OSDs;
+    False otherwise.
+    :rtype: bool
```
gabrielcocenza, line 168: update docstrings

gabrielcocenza (changes requested):

cou/utils/app_utils.py
```diff
@@ -126,3 +109,101 @@ def validate_ovn_support(version: str) -> None:
...
+    :raises RunUpgradeError: When an upgrade fails.
+    """
+    current_require_osd_release = ""
+    check_command = "ceph osd dump"
```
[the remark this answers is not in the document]
agileshaw: I actually didn't know it supports formatting to json. Thanks! Now parsing the output is much more robust.

cou/utils/app_utils.py
agileshaw: They do have very similar structures, and I would love to merge them. However, their error messages are tailored, so it is hard to do so. Do you have any suggestions?

cou/utils/app_utils.py
```diff
@@ -126,3 +109,101 @@ def validate_ovn_support(version: str) -> None:
...
+            if len(osd_release_output) > 1:
+                raise RunUpgradeError(
+                    f"OSDs are on mismatched releases:\n{osd_release_output}."
+                    "Please manually upgrade them to be on the same release before proceeding."
+                )
+
+            # get release name from "osd_release_output". Example value of "osd_release_output":
+            # {'ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632) octopus (stable)': 1}
+            osd_release_key = list(osd_release_output.keys())[0]
+            current_osd_release = osd_release_key.split(")", maxsplit=1)[-1].split("(")[0].strip()
```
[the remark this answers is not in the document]
agileshaw: I thought about this, but doing regex matching seems to be a bit too much. In the latest patch, I changed it to use the location where the ceph release is, but I will think about the regex solution.

cou/utils/app_utils.py
gabrielcocenza: taking a closer look, set_require_osd_release_option also has a similar structure. I would try something like:

```python
async def _run_ceph_mon_command(command, unit, model: COUModel, run_upgrade_err_msg):
    logger.debug("Running '%s' on '%s'", command, unit)
    try:
        result = await model.run_on_unit(unit_name=unit, command=command, timeout=600)
        if str(result["Code"]) != "0":
            raise RunUpgradeError(run_upgrade_err_msg) from CommandRunFailed(cmd=command, result=result)
        return result["Stdout"]

    except JujuError as exc:
        raise RunUpgradeError(run_upgrade_err_msg) from exc

async def _get_required_osd_release(unit: str, model: COUModel) -> str:
    cmd = "ceph versions -f json"
    err_msg = f"Cannot determine the current value of require_osd_release on ceph-mon unit '{unit}'."
    return json.loads(wait _run_ceph_mon_command(cmd, unit, err_msg)).get("require_osd_release", "")
```

This can increase substantially the readability and apply DRY

cou/utils/app_utils.py
```diff
@@ -126,3 +109,98 @@ def validate_ovn_support(version: str) -> None:
...
+            if len(osd_release_output) > 1:
+                raise RunUpgradeError(
+                    f"OSDs are on mismatched releases:\n{osd_release_output}."
+                    "Please manually upgrade them to be on the same release before proceeding."
+                )
+
+            # get release name from "osd_release_output". Example value of "osd_release_output":
+            # {'ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632) octopus (stable)': 1}
+            osd_release_key, *_ = osd_release_output.keys()
+            current_osd_release = osd_release_key.split(" ")[4].strip()
```
gabrielcocenza, line 194: I still think that a dictionary on openstack.py having the ceph  could be beneficial here.

```python
CEPH_RELEASES = OrderedDict(
    [
        ....
        ("pacific", VersionRange("16.2.0", "17.2.0" )),
        .....
    ]
)

....
if current_osd_release not in CEPH_RELEASES.keys():
     raise RunUpgradeError(f"Invalid ceph-osd release '{current_osd_release}'")
```

gabrielcocenza (changes requested):

cou/utils/app_utils.py
agileshaw: Updated

cou/utils/juju_utils.py
```diff
@@ -326,12 +327,20 @@ async def run_on_unit(
...
         """
+        logger.debug("Running '%s' on '%s'", command, unit_name)
+
         unit = await self._get_unit(unit_name)
         action = await unit.run(command, timeout=timeout)
         results = action.data.get("results")
-        return _normalize_action_results(results)
+        normalize_results = _normalize_action_results(results)
+
+        if str(normalize_results["Code"]) != "0":
```
rgildein, line 339: Should we use same approach as we are doing in `run_action`? I mean to use `raise_on_failure`.

tests/unit/apps/test_auxiliary.py
```diff
@@ -413,11 +413,9 @@ def test_ceph_mon_upgrade_plan_xena_to_yoga(
             coro=model.upgrade_charm(app.name, "pacific/stable", switch=None),
         ),
         UpgradeStep(
-            description=(
-                "Ensure require-osd-release option on ceph-mon units correctly set to 'pacific'"
-            ),
+            description=("Ensure require-osd-release option matches with ceph-osd version"),
```
rgildein, line 416: not-blocking:
```suggestion
            description="Ensure require-osd-release option matches with ceph-osd version",
```

rgildein (commented):
Mostly it looks good to me, but there is one comment about raising an exception.

cou/utils/juju_utils.py
agileshaw: I don't think so. The point of having `raise_on_failure` for actions is because we are retrying for getting action result. But for run on command, getting the result is suppose to be static and no need for retrying.

cou/utils/openstack.py
```diff
@@ -147,6 +147,12 @@
     "lunar": "antelope",
 }
 
+CEPH_RELEASES = [
+    "octopus",
```
gabrielcocenza, line 151: I think it's possible to encounter clouds using `natilus` if they are coming from bionic, so let's add to this list

tests/unit/utils/test_app_utils.py
```diff
@@ -79,235 +78,162 @@ async def test_application_upgrade_packages_with_hold(model):
...
-        timeout=600,
-    )
-
-
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    "current_release_value, target_release_value",
+    "current_required_osd_release, current_osd_release",
     [
         ("octopus", "pacific"),
```
gabrielcocenza, line 85: add ("natilus", "octopus")

tests/unit/utils/test_app_utils.py
```diff
@@ -79,235 +78,162 @@ async def test_application_upgrade_packages_with_hold(model):
...
+    check_output = """
+    {
+        "mon": {
+            "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632) octopus (stable)": 1
+        },
+        "mgr": {
+            "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632) octopus (stable)": 1
+        },
+        "osd": {
+            "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632) %s (stable)": 3
```
gabrielcocenza, line 167: If we are not parameterizing, why not use `octopus` directly here?

cou/utils/openstack.py
agileshaw: We are not letting people upgrade from bionic, so I don't think adding `nautilus` is necessary nor desirable

tests/unit/utils/test_app_utils.py
agileshaw: Because I think it it is clearer to see what we are expecting with `expected_osd_release`.

cou/utils/openstack.py
gabrielcocenza: I think it's easy to someone forget to do this post upgrade step of `require-osd-release` and have a focal cloud on natilus

agileshaw: We are only using this CEPH_RELEASES list for checking whether the osd versions are in these range,  and IMO if their cloud is correct, they should never have OSD running `nautilus` on focal. If the operator starts an upgrade with `nautilus` value of `required_osd_release` but osd is at `octopus`, that value will be overwrite to `octopus` in the pre-upgrade step of ceph-mon.

cou/utils/juju_utils.py
rgildein: But what if I want to run command and I expecting that it could failed?

agileshaw: Opened #148 to track

cou/utils/openstack.py
gabrielcocenza: I see your point and will resolve this conversation, but as you said `required_osd_release` will overwrite to octopus in the pre-upgrade step of ceph-mon, so I think it's good to have it on `test_set_require_osd_release_option_different_releases`

rgildein (approved):
LGTM

cou/utils/juju_utils.py
rgildein: Thanks

gabrielcocenza (approved):
LGTM. Thanks

rgildein (approved):
LGTM

agileshaw:
Will wait for the CI to trigger again before merging

PR #132, opened 2023-10-20
Warn if data plane needs to upgrade manually

### LINKED ISSUE #105: COU should warn that Data Plane apps need to upgrade manually

gabrielcocenza:
- back to instantiate data-plane charms
- ceph-osd class
- separate min OpenStack release from data and control plane apps

Closes #105

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -179,6 +179,10 @@ class MysqlInnodbClusterApplication(OpenStackAuxiliaryApplication):
     packages_to_hold: Optional[list] = ["mysql-server-core-8.0"]
 
 
+# NOTE (gabrielcocenza): Although CephOSD class is empty now, it will be
+# necessary to add post upgrade plan to set require-osd-release. Registering on
+# OpenStackAuxiliaryApplication can be easily forgot and ceph-osd can't be instantiated
+# as a normal OpenStackApplication.
 @AppFactory.register_application(["ceph-osd"])
 class CephOSD(OpenStackAuxiliaryApplication):
     """Application for ceph-osd."""
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -177,3 +177,8 @@ class MysqlInnodbClusterApplication(OpenStackAuxiliaryApplication):
     # NOTE(agileshaw): holding 'mysql-server-core-8.0' package prevents undesired
     # mysqld processes from restarting, which lead to outages
     packages_to_hold: Optional[list] = ["mysql-server-core-8.0"]
+
+
+@AppFactory.register_application(["ceph-osd"])
+class CephOSD(OpenStackAuxiliaryApplication):
+    """Application for ceph-osd."""
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -83,13 +83,13 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
     return VerbosityLevel(verbosity).name
 
 
-async def analyze_and_plan(model_name: Optional[str] = None) -> UpgradeStep:
+async def analyze_and_plan(model_name: Optional[str] = None) -> tuple[Analysis, UpgradeStep]:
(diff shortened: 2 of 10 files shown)


cou/apps/auxiliary.py
```diff
@@ -177,3 +177,8 @@ class MysqlInnodbClusterApplication(OpenStackAuxiliaryApplication):
     # NOTE(agileshaw): holding 'mysql-server-core-8.0' package prevents undesired
     # mysqld processes from restarting, which lead to outages
     packages_to_hold: Optional[list] = ["mysql-server-core-8.0"]
+
+
+@AppFactory.register_application(["ceph-osd"])
+class CephOSD(OpenStackAuxiliaryApplication):
+    """Application for ceph-osd."""
```
rgildein, line 184: It's not enough to add `ceph-osd` to `register_application` for `OpenStackAuxiliaryApplication`?

cou/steps/analyze.py
```diff
@@ -160,7 +181,15 @@ def current_cloud_os_release(self) -> Optional[OpenStackRelease]:
...
         )
+        data_plane = [self.min_os_version_data_plane] if self.min_os_version_data_plane else []
+        return min(control_plane + data_plane, default=None)
+
+    def manually_upgrade_data_plane(self) -> None:
+        """Warning message to upgrade data plane charms if necessary."""
+        if self.min_os_version_control_plane and self.min_os_version_data_plane:
+            if self.min_os_version_control_plane > self.min_os_version_data_plane:
+                data_plane_apps = ", ".join([app.name for app in self.apps_data_plane])
+                print(f"WARNING: Please upgrade manually the data plane apps: {data_plane_apps}")
```
rgildein, line 195: To my this should not be part of analyze, it's saying upgrade plan so it's should be part of generate plan. What do you say?

rgildein (changes requested):

cou/apps/auxiliary.py
gabrielcocenza: Yes and no. Ceph-osd will need special upgrade steps for sure and my concern is that we forget to unregister on  `OpenStackAuxiliaryApplication`

aieri: ok then I think you could add a comment indicating that although it's empty for now, this is the right place for the new class as it will need overrides later on

cou/cli.py
```diff
@@ -115,7 +115,7 @@ async def get_upgrade_plan(model_name: Optional[str] = None) -> None:
     analysis_result, upgrade_plan = await analyze_and_plan(model_name)
     logger.info(upgrade_plan)
     print(upgrade_plan)  # print plan to console even in quiet mode
-    analysis_result.manually_upgrade_data_plane()
+    manually_upgrade_data_plane(analysis_result)
```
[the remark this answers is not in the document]
rgildein: I meant call this function there, so printing will happen after plan is generated, but before it's returns from functions.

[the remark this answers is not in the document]
gabrielcocenza: IMO it's better to have this message at the bottom because it's the last thing that the user will see.

cou/steps/analyze.py
```diff
@@ -151,6 +158,20 @@ def __str__(self) -> str:
...
+
+        :param apps: Applications.
+        :type apps: list[OpenStackApplication]
+        :return: OpenStack release.
+        :rtype: Optional[OpenStackRelease]
+        """
+        return min(
+            (app.current_os_release for app in apps),
+            default=None,
+        )
```
rgildein, line 173: not-blocking:
```suggestion
        return min((app.current_os_release for app in apps), default=None)
```

rgildein (approved):

rgildein (approved):

aieri (approved):

Issue #129, opened 2023-10-19
Be clear about which release we're upgrading from and to


aieri:
Although the plan is very clear about what's going to happen to the individual cloud components, we don't clarify what version the overall cloud is at, and which OpenStack release we're going to be upgrading to.

agileshaw:
I implemented something that displays current and next cloud OS release ([PR#120](https://github.com/canonical/charmed-openstack-upgrader/pull/120/files#diff-e5cc2183ba957040fcefe3159399dc8f89b61edcef0ba57300fdfca6c30dd291R151-R152)). Currently this is only a part of analysis result and will not show unless verbose mode is turned on. I can add it to the plan and always display it (except in quiet mode) before the actual steps

Issue #99, opened 2023-10-04
OpenStackRelease should be smarter about next_release


gabrielcocenza:
Currently, the `OpenStackRelease` class is not able to predict series upgrade and this can cause issues if user run to upgrade an application that is already considered as yoga. E.g: if you deploy just rabbitmq-server and run cou, the following error will show:

```
cou run 
Analyzing cloud... ✔
Generating upgrade plan... \2023-10-03 21:35:11 [ERROR] Cannot generate upgrade plan for 'rabbitmq-server': Cannot find a suitable 'rabbitmq-server' charm channel for zed
Generating upgrade plan... ✖
2023-10-03 21:35:11 [ERROR] Cannot find a suitable 'rabbitmq-server' charm channel for zed 
```

This happens because by the workload version, rabbitmq is considered as yoga and the next OpenStack release is zed. There is no track that maps rabbitmq, focal and zed

PR #134, opened 2023-10-21
Improve handling of Ubuntu series and OS release combinations

agileshaw:
- Get current minimum series of the cloud in the analysis part
- set current_cloud_os_release and current_cloud_series right away in the analysis part so that errors can be caught early
- Add LTS_TO_OS_RELEASE, which maps a LTS series to a range of supporting OS releases
- Add function in plan stage to determine the upgrade route based on series/release combination, including limiting the support range of COU to focal ussuri - yoga
- Add message about which release we're upgrading from and to
- Fix release codename typo: kinect -> kinetic

fixes: #99 
fixes: #129


tests/unit/steps/test_steps_plan.py
```diff
@@ -162,14 +162,74 @@ async def test_generate_plan(apps, model):
...
-async def test_generate_plan_raise_NoTargetError(mocker):
-    exp_error_msg = "Cannot find target to upgrade."
-    analysis_result = MagicMock(spec=Analysis)
-    analysis_result.current_cloud_os_release.next_release = None
-    # not possible to determine target
+@pytest.mark.parametrize(
+    "current_os_release, current_series, next_release",
+    [
+        (OpenStackRelease("victoria"), "focal", "wallaby"),
+        (OpenStackRelease("yoga"), "jammy", "zed"),
```
gabrielcocenza, line 169: This test should fail IMO. If a user manually upgrade the series to jammy and run COU I would fell more comfortable not letting go ahead and upgrade to zed.

gabrielcocenza (changes requested):

cou/steps/analyze.py
```diff
@@ -43,6 +43,11 @@ class Analysis:
     apps_control_plane: list[OpenStackApplication]
     apps_data_plane: list[OpenStackApplication]
 
+    def __post_init__(self) -> None:
+        """Initialize the Analysis dataclass."""
+        self.current_cloud_os_release = self._get_minimum_cloud_os_release()
+        self.current_cloud_series = self._get_minimum_cloud_series()
```
rgildein, line 49: Can we define these in dataclass itself and set value only here? e.g.
 ```python
>>> @dataclass
... class A:
...  a: str = field(init=False)
...  def __post_init__(self) -> None:
...   self.a = "test"
... 
>>> obj = A()
>>> obj.a
'test'
```

File: cou/steps/plan.py [hunk 3]
```diff
@@ -117,3 +119,55 @@ async def create_upgrade_group(
...
+    :return: The target OS release to upgrade the cloud to.
+    :rtype: str
+    """
+    if not current_os_release:
+        raise NoTargetError(
+            "Cannot determine the current OS release in the cloud. "
+            "Is this a valid OpenStack cloud?"
+        )
+
+    supporting_lts_series = ",".join(list(LTS_TO_OS_RELEASE.keys()))
```
rgildein, line 144: not-blocking: 
```suggestion
    supporting_lts_series = ",".join(LTS_TO_OS_RELEASE)
```

cou/steps/plan.py
```diff
@@ -117,3 +119,55 @@ async def create_upgrade_group(
...
+
+    try:
+        # Check if the release is the "last" supported by the series
+        if str(current_os_release) == LTS_TO_OS_RELEASE[current_series][-1]:
+            print(
+                f"No upgrade available. '{current_os_release}' (the minimum OS release found "
+                f"in the cloud) is the highest release '{current_series}' (the minimum Ubuntu "
+                "series in the cloud) supports."
+            )
+            sys.exit(0)
```
rgildein, line 159: To me this is not an good idea hiding sys exit in code. I think we should raise and exception here. Exception should take that string as message and in CLI we should catch it and print the message.

tests/unit/steps/test_steps_analyze.py
```diff
@@ -137,6 +139,30 @@ async def test_analysis_detect_current_cloud_os_release_same_release(apps, model
     assert result.current_cloud_os_release == "ussuri"
 
 
+@pytest.mark.asyncio
+async def test_analysis_detect_current_cloud_series_same_series(apps, model):
+    result = analyze.Analysis(
+        model=model,
+        apps_control_plane=[apps["rmq_ussuri"], apps["keystone_wallaby"], apps["cinder_ussuri"]],
+        apps_data_plane=[],
```
rgildein, line 147: I think that we should used mocked apps here, this is not related with app class at all and if that failed all tests like this will failed and that bad practice.

rgildein (changes requested):

cou/steps/plan.py
```diff
@@ -117,3 +119,55 @@ async def create_upgrade_group(
...
+            f"The supporting series are: {supporting_lts_series}"
+        )
+
+    try:
+        # Check if the release is the "last" supported by the series
+        if str(current_os_release) == LTS_TO_OS_RELEASE[current_series][-1]:
+            print(
+                f"No upgrade available. '{current_os_release}' (the minimum OS release found "
+                f"in the cloud) is the highest release '{current_series}' (the minimum Ubuntu "
+                "series in the cloud) supports."
```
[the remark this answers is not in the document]
agileshaw: I re-worded this message. Please let me know if it is clearer

cou/steps/plan.py
```diff
@@ -117,3 +119,55 @@ async def create_upgrade_group(
...
+            )
+            sys.exit(0)
+
+        # get the next release as the target from the current cloud os release
+        target = current_os_release.next_release
+        if not target:
```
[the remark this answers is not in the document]
agileshaw: The purpose of this PR was to just have a smart way of determine whether the next move should be series upgrade or release upgrade, so limiting the support range to ussuri<->yoga was not in scope. But I think we could expand the scope of this PR and include it here. Updated in the new patch

cou/steps/plan.py
```diff
@@ -117,3 +119,55 @@ async def create_upgrade_group(
...
+            raise NoTargetError(
+                "Cannot find target to upgrade. Current minimum OS release is "
+                f"'{str(current_os_release)}'. Current Ubuntu series is '{current_series}'."
```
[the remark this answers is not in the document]
agileshaw: I'm also letting the user know the starting point should be larger than ussuri

tests/unit/steps/test_steps_plan.py
agileshaw: As we are including the range limit, I revised the tests

cou/steps/analyze.py
agileshaw: Good suggestion. Thanks :+1:

cou/steps/plan.py
agileshaw: I was actually torn between this approach and what you are suggesting here. After re-consideration, yes, I think putting all `sys.exit` in `cli.py` is better.

tests/unit/steps/test_steps_analyze.py
agileshaw: We could but I don't think it is necessary because this is defined in `conftest.py`, which is meant to be shared across multiple test modules.

cou/steps/plan.py
```diff
@@ -119,3 +128,66 @@ async def create_upgrade_group(
...
+        if current_os_release < OpenStackRelease("ussuri") or OpenStackRelease(
+            target
+        ) > OpenStackRelease("yoga"):
+            raise ReleaseOutOfRange(
+                "At the moment COU only supports upgrade in the range of `ussuri` and `yoga` "
+                "(inclusive) for `focal` series. Current minimum OS release is "
+                f"'{str(current_os_release)}'. Current Ubuntu series is '{current_series}'."
+            )
+        return target
+    except KeyError:  # raise exception if the current series is not in LTS_TO_OS_RELEASE map
```
gabrielcocenza, line 189: IMO this try and except `KeyError` is catching too much code. Is  it possible to isolate where you are really checking that?

agileshaw:
Create issue #144 to remind us to expand supporting range when COU can handle more releases/series in the future

rgildein (approved):
LGTM

gabrielcocenza (approved):

Issue #123, opened 2023-10-19
cou should check the can_upgrade_to on status of application


gabrielcocenza:
principal and subordinate charms should check `can_upgrade_to` on the application status before trying to refresh to the current channel. With this kind of check we can avoid subordinates generating plan when there is nothing to do

PR #137, opened 2023-10-24
Check if application can upgrade to the current channel

gabrielcocenza:
- generate_upgrade_plan can return empty plan
- create_upgrade_group can return empty plan
Closes: #123


aieri (dismissed):

cou/apps/core.py
```diff
@@ -422,7 +431,7 @@ def generate_upgrade_plan(self, target: str) -> UpgradeStep:
         for step in all_steps:
             if step:
                 upgrade_steps.add_step(step)
-        return upgrade_steps
+        return upgrade_steps if upgrade_steps.sub_steps else None
```
rgildein, line 434: What if we provide `empty` property, and basically skip any empty step instead of returning the `None`?

cou/steps/plan.py
```diff
@@ -88,7 +90,7 @@ async def create_upgrade_group(
     target: str,
     description: str,
     filter_function: Callable[[OpenStackApplication], bool],
-) -> UpgradeStep:
+) -> Optional[UpgradeStep]:
```
rgildein, line 93: I do not see how this can returns `None`. I think it will always returns `group_upgrade_plan`.

cou/steps/plan.py
```diff
@@ -101,13 +103,14 @@ async def create_upgrade_group(
...
+    :rtype: Optional[UpgradeStep]
     """
     group_upgrade_plan = UpgradeStep(description=description, parallel=False)
     for app in filter(filter_function, apps):
         try:
-            app_upgrade_plan = app.generate_upgrade_plan(target)
-            group_upgrade_plan.add_step(app_upgrade_plan)
+            app_upgrade_plan: Optional[UpgradeStep] = app.generate_upgrade_plan(target)
+            if app_upgrade_plan:
+                group_upgrade_plan.add_step(app_upgrade_plan)
```
rgildein, line 113: If we define empty property step we can do `if.app_upgrade_plan.empty`, which will be similar to what we have here. But what if we simply not show empty step in string representation and in execution we can skip any empty step. Like this we will have information about step, but we will know that's empty.

rgildein (commented):
I do not like that `generate_upgrade_plan` can return `None` if there are no sub-steps, to me the plan should be simply empty and we can define `empty` property or batter will be to just define class representation better, so we can do `if step: ...`.

cou/apps/subordinate.py
```diff
@@ -26,36 +25,36 @@
...
         """Post Upgrade planning.
 
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
         :return: Plan that will add post upgrade as sub steps.
-        :rtype: list[Optional[UpgradeStep]]
+        :rtype: list[UpgradeStep]
         """
-        return [None]
+        return [UpgradeStep()]
```
rgildein, line 57: Why this cannot be simply empty list?
```suggestion
        return []
```

cou/steps/__init__.py
```diff
@@ -134,6 +134,17 @@ def __repr__(self) -> str:
...
+    def __bool__(self) -> bool:
+        """Boolean magic method for UpgradeStep.
+
+        :return: True if there is at least one coroutine in a UpgradeStep
+        or in its sub steps.
+        :rtype: bool
+        """
+        if self._coro or any(step._coro for step in self.sub_steps):
+            return True
+        return any(bool(step) for step in self.sub_steps)
```
rgildein, line 146: I think that logic will be same if you use `and`, or not?
```suggestion
        return self._coro and any(bool(step) for step in self.sub_steps):
```

cou/steps/plan.py
```diff
@@ -107,7 +107,8 @@ async def create_upgrade_group(
     for app in filter(filter_function, apps):
         try:
             app_upgrade_plan = app.generate_upgrade_plan(target)
-            group_upgrade_plan.add_step(app_upgrade_plan)
+            if app_upgrade_plan:
```
rgildein, line 110: Is this if even needed? Empty steps will not be show to user and their `run` will do nothing.

tests/unit/steps/test_steps.py
```diff
@@ -80,17 +80,65 @@ def test_step_eq(description, parallel, args):
...
     plan.sub_steps = [sub_step, UpgradeStep(description="a.b")]
 
+    assert str(plan) == ""
+
+
+def test_step_str_partially_show():
+    """Test UpgradeStep string representation when print UpgradeSteps that have coro."""
+    expected = "a\n\ta.a\n\t\ta.a.a\n"
+    plan = UpgradeStep(description="a")
+    sub_step = UpgradeStep(description="a.a")
```
rgildein, line 135: Can you also test to add empty step without even the description? I think we should skip such step in `__str__`, because right now, if I'm right, we will get `expected = "a\n\t\n\ta.a\n\t\ta.a.a\n"`

cou/steps/__init__.py
gabrielcocenza: No. The Top plan doesn't have a coroutine (self._coro) but their sub steps will have.

rgildein: That is true, but `return self._coro is not None or any(bool(step) for step in self.sub_steps)` will work.

gabrielcocenza: That is what I did in the next commit

cou/apps/subordinate.py
```diff
@@ -26,36 +25,36 @@
...
         """Post Upgrade planning.
 
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
         :return: Plan that will add post upgrade as sub steps.
-        :rtype: list[Optional[UpgradeStep]]
+        :rtype: list[UpgradeStep]
         """
-        return [None]
+        return []
```
aieri, line 57: is this correct? If we claim to be returning an UpgradeStep should it not be `return [UpgradeStep()]`?

aieri (commented):
Looks good overall, just a question

cou/apps/subordinate.py
gabrielcocenza: I was doing that one commit before and @rgildein suggested to change to an empty list. The type `list[UpgradeStep]` includes the empty list, So I think either way is fine.

aieri (approved):

rgildein (approved):
LGTM

Issue #145, opened 2023-10-30
wait_for_idle should not have retry


rgildein:
The `wait_for_idle` should not use retry or that retry needs to have high timeout than the timeout used in `wait_for_idle` function.

Output example:
```bash
...
Wait 300 s for model openstack to reach the idle state. (c)ontinue/(a)bort/(s)kip:c
Generating upgrade plan... ✖
The connection was lost. Check your connection or increase the timeout.
```

PR #109, opened 2023-10-09
catch KeyboardInterrupt error

rgildein:
We did not catch `KeyboardInterrupt` error before, what cause showing lot of tracebacks if crt+c was pressed.

- [x] add proper `KeyboardInterrupt` handling
- [x] ~~add prompt for confirmation before unsafe cancel~~
- [x] fix all unit tests

fixes: #103

---
Results from my testing
```bash
$ cou run --model zaza-413ca955bd34
Connecting to 'zaza-413ca955bd34' model... ✖
charmed-openstack-upgrader has been terminated
$ cou run --model zaza-413ca955bd34
Connected to 'zaza-413ca955bd34' ✔
Analyzing cloud... ✖
charmed-openstack-upgrader has been terminated
$ cou run --model zaza-413ca955bd34
Connected to 'zaza-413ca955bd34' ✔
Analyzing cloud... ✔
Generating upgrade plan... \Upgrading cloud from 'ussuri' to 'victoria'
.
Generating upgrade plan... ✔
Top level plan
        backup mysql databases
        Control Plane principal(s) upgrade plan
                Upgrade plan for 'rabbitmq-server' to victoria
                        Upgrade software packages of 'rabbitmq-server' from the current APT repositories
                        Upgrade 'rabbitmq-server' to the new channel: '3.9/stable'
                        Change charm config of 'rabbitmq-server' 'source' to 'cloud:focal-victoria'
                        Wait 1800s for model zaza-413ca955bd34 to reach the idle state.
                        Check if the workload of 'rabbitmq-server' has been upgraded
                Upgrade plan for 'keystone' to victoria
                        Upgrade software packages of 'keystone' from the current APT repositories
                        Upgrade 'keystone' to the new channel: 'victoria/stable'
                        Change charm config of 'keystone' 'openstack-origin' to 'cloud:focal-victoria'
                        Wait 1800s for model zaza-413ca955bd34 to reach the idle state.
                        Check if the workload of 'keystone' has been upgraded
                Upgrade plan for 'cinder' to victoria
                        Upgrade software packages of 'cinder' from the current APT repositories
                        Upgrade 'cinder' to the new channel: 'victoria/stable'
                        Change charm config of 'cinder' 'openstack-origin' to 'cloud:focal-victoria'
                        Wait 300s for app cinder to reach the idle state.
                        Check if the workload of 'cinder' has been upgraded
                Upgrade plan for 'glance' to victoria
                        Upgrade software packages of 'glance' from the current APT repositories
[...]

PR #146, opened 2023-10-30
remove retry from wait_for_idle

rgildein:
fixes: #145


agileshaw (approved):

rgildein:
After yesterday discussion I think that there could be better solution to solve #145. I will write all options down for further discussion.

1. drop retry for `wait_for_idle`, which is the simplest solution but if we lose connection during this call COU will fail. If COU fail and we are waiting for Keystone, user can easily run COU again and that can cause other problem since we will not wait for Keystone to finish. This may be fixed by fixing #78.
2. Increase the timeout to max timeout (25 min for Keystone), but this cause unintentionally waiting.
3. Add simple wrapper function, which will be decorated with retry and timeout provided from `wait_for_idle` itself.
4. Add unlimited timeout option to retry and define max number of attempts. Like this the `wait_for_idle` function can be retry max 3 times.

gabrielcocenza (approved):
LGTM

cou/utils/juju_utils.py
```diff
@@ -434,7 +433,22 @@ async def wait_for_idle(self, timeout: int, apps: Optional[list[str]] = None) ->
...
+            model = await self._get_model()
+            try:
+                await model.wait_for_idle(
+                    apps=apps, timeout=timeout, idle_period=DEFAULT_MODEL_IDLE_PERIOD
+                )
+            except asyncio.exceptions.TimeoutError as error:
+                if apps is None:
+                    msg = f"Model {self.name} did not reach its idle state in {timeout}s"
+                else:
+                    msg = f"Apps {','.join(apps)} did not reach their idle state in {timeout}s"
```
agileshaw, line 450: The error message would be incorrect if only a subset of the apps have not reached idle state. The ideal approach would be providing the apps that are actually not at idle state. If that's too complicated, we should at very least change the error message to the something like:
```suggestion
                    msg = f"Not all apps {','.join(apps)} reached their idle state in {timeout}s"
```

agileshaw (changes requested):

cou/utils/juju_utils.py
```diff
@@ -458,10 +457,22 @@ async def wait_for_idle(self, timeout: int, apps: Optional[list[str]] = None) ->
...
+        async def _wait_for_idle() -> None:
+            # NOTE(rgildein): Defining wrapper so we can use retry with proper timeout
+            model = await self._get_model()
+            try:
+                await model.wait_for_idle(
+                    apps=apps, timeout=timeout, idle_period=DEFAULT_MODEL_IDLE_PERIOD
+                )
+            except asyncio.exceptions.TimeoutError as error:
+                msg = str(error)
+                msg = msg.replace("\n", "\n  ")  # make error message more readable
```
gabrielcocenza, line 472: It's not clear to me what this message will look like. Can you provide at least one expected message on unit tests?

rgildein: After second though I really do not think that it will be helpful to add such example to unit tests, since we need to define exception and than catch it. I provided example in `Note`.

tests/unit/utils/test_juju_utils.py
```diff
@@ -459,4 +460,26 @@ async def test_coumodel_wait_for_idle(mock_get_supported_apps, mocked_model):
...
+    model = juju_utils.COUModel(None)
+
+    with pytest.raises(WaitForApplications):
+        await model.wait_for_idle(timeout, apps=apps)
+
+    if apps is None:
+        mock_get_supported_apps.assert_awaited_once_with()
+        exp_apps = mock_get_supported_apps.return_value
+    else:
+        exp_apps = apps
```
agileshaw, line 481: Non-blocking: Since in https://github.com/canonical/charmed-openstack-upgrader/pull/109 you separated test cases to avoid `if` statements in unit tests, I would prefer we do the same here and keep it consistent

agileshaw (changes requested):

agileshaw (approved):
LGTM

Issue #106, opened 2023-10-06
Charm Gnocchi can't be considered as a normal OpenStackApplication


gabrielcocenza:
charm-gnocchi has the same workload version from Victoria to Wallaby. With this behavior, after upgrading the cloud to Victoria and running `cou plan` the following error appears:

```
Analyzing cloud... ✔
2023-10-06 14:53:38 [ERROR] Cannot generate upgrade plan for 'gnocchi': 'gnocchi' has unexpected channel: 'victoria/stable' for the current workload version and OpenStack release: 'wallaby'. Possible channels are: wallaby/stable
Generating upgrade plan... ✖
2023-10-06 14:53:38 [ERROR] 'gnocchi' has unexpected channel: 'victoria/stable' for the current workload version and OpenStack release: 'wallaby'. Possible channels are: wallaby/stable
```

With this error and with the need of including `glance-simplestreams-sync` described on #104 , we probably need an OpenStack class that the identification of the current OpenStack release come from the charm channel instead of the workload version

gabrielcocenza:
Closed by #156

PR #147, opened 2023-10-30
Add support to designate-bind and gnocchi

gabrielcocenza:
- Those charms can be compatible with more than one OpenStack release for a workload version, so they should be channel based.

Closes: #106 #94

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -34,6 +34,7 @@ from cou.utils.app_utils import upgrade_packages
 from cou.utils.juju_utils import COUModel
 from cou.utils.openstack import (
     DISTRO_TO_OPENSTACK_MAPPING,
+    OPENSTACK_CHANNEL_BASED_CHARMS,
     OpenStackCodenameLookup,
     OpenStackRelease,
 )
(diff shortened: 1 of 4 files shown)


rgildein (dismissed):
LGTM

cou/apps/base.py
```diff
@@ -183,6 +184,11 @@ def channel(self) -> str:
         :return: Charm channel. E.g: ussuri/stable
         :rtype: str
         """
+        # NOTE(gabrielcocenza) Some applications current_os_release points
+        # to channel_codename without setting the channel before. On that
```
agileshaw, line 188:
```suggestion
        # to channel_codename without setting the channel before. In that
```

tests/unit/apps/test_base.py
```diff
@@ -0,0 +1,240 @@
...
+            coro=app._check_upgrade(target),
+        ),
+    ]
+
+    add_steps(expected_plan, upgrade_steps)
+
+    assert upgrade_plan == expected_plan
+
+
+def test_application_gnocchi_upgrade_plan_xena_to_yoga(status, config, model):
```
agileshaw, line 113: By looking at the test code, it's unclear why we would need to test upgrade plan for both ussuri->victoria and xena->yoga. It there are additional things covered by one of the plan, we can mention it in the docstring? Otherwise, I think we should remove one of them.

agileshaw (changes requested):

Issue #149, opened 2023-10-30
The `COU_DATA` is define in two places


rgildein:
Currently the `COU_DATA`  environment variable is defined in two places [1] and [2]. It should be defined only at one place.

[1]: https://github.com/canonical/charmed-openstack-upgrader/blob/374b3e0d99fd33dc6cc0b0d8c6251eecace6f08e/cou/logging.py#L22
[2]: https://github.com/canonical/charmed-openstack-upgrader/blob/374b3e0d99fd33dc6cc0b0d8c6251eecace6f08e/cou/steps/backup.py#L45

Issue #151, opened 2023-10-30
CLI is saying that `MODEL_NAME` can be used


rgildein:
The `MODEL_NAME` environment variable was dropped with zaza, but it's still present in CLI.  That's why it should be removed from CLI to not confuse any user.

gabrielcocenza:
Closed by #157

PR #157, opened 2023-10-31
clean up usage of environment variables

rgildein:
The COU_DATA was defined in two places, so I moved it only to single place. Remove mention of `MODEL_NAME` from CLI.

fixes: #149 #151

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -80,10 +80,8 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         default=None,
         dest="model_name",
         type=str,
-        help="Set the model to operate on.\nIf unset, the model name will be determined by "
-        "inspecting the environment as follows:\n"
-        "  1 - Environment variable JUJU_MODEL,"
-        "  2 - Current active juju model",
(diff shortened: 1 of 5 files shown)


agileshaw (approved):

cou/commands.py
```diff
@@ -83,8 +83,7 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         help="Set the model to operate on.\nIf unset, the model name will be determined by "
         "inspecting the environment as follows:\n"
         "  1 - Environment variable JUJU_MODEL,"
```
gabrielcocenza, line 85: `JUJU_MODEL` it's  not been used on [pythonlib-juju](https://github.com/juju/python-libjuju/blob/2.9/juju/client/jujudata.py#L35). I've manually tested and it does not work. I think we should remove it too and let just the `Current active juju model`

gabrielcocenza (changes requested):

cou/commands.py
rgildein: True, I did not know that it's not working since I'm using only `JUJU_DATA`.

gabrielcocenza (approved):

Issue #82, opened 2023-09-23
Drop the skip option


aieri:
Skipping steps can be dangerous, and iiuc doesn't really add any benefit when a user can stop the process, perform manual tweaks, and rerun COU from the top.

If we move taking a backup (or not) to a separate CLI option we could limit interactive choices to continue or stop/abort/exit.

aieri:
After thinking about this some more, I think it'd be worth simplifying even further by only prompting at a per-application level. Mock-up:
```
Upgrade of application keystone from Ussuri to Victoria:
    * software package upgrade to the latest revision in APT repository distro
    * charm refresh to the latest revision of channel ussuri/stable
    * charm upgrade to the latest revision of channel victoria/stable
    * software package upgrade to the latest revision in APT repository cloud:focal-victoria
 
Continue? (y/N): _
```

PR #159, opened 2023-11-01
Make backup step optional with CLI parameter

agileshaw:
As part of the effort of dropping skip option, we need to make backup step an optional CLI parameter. Operators can choose whether to backup cloud database before upgrade with `cou [plan|run] --backup` or `cou [plan|run] --no-backup`. If no related parameter is supplied, the default action of backing up database will be selected.

Also removes unnecessary default parameter values for functions in cli.py

relates-to: #82


rgildein (approved):
One not blocking comment, other that that +1

gabrielcocenza (dismissed):
There is a merge conflict, but LGTM

rgildein (approved):
LGTM

gabrielcocenza (approved):

Issue #155, opened 2023-10-30
COU should wait at least 20 mim for Keystone


rgildein:
Running keystone upgrade can took more than 20 minutes and right now COU is waiting only 5 min.

rgildein:
mysql-innodb-cluster required more than 120s, maybe 20-25 minutes

facundofc:
This issue isn't fixed. I can't really tell which timeouts were increased in #160, but there seems to be one that was left at 5 mins and it shouldn't.

When the channel for the Keystone charm is changed, `upgrade-charm` and `config-changed` are triggered in sequence. These hooks both iterate through all the endpoints and registers them (I have no idea what this operation is). This process takes more than 6 mins for each hook, so an adequate waiting time for Keystone after changing channel would be around 20 mins.

facundofc:
Hey Gabriel. Yes, Keystone times out _every time_ its charm's upgraded.

It's good to know about the env var, I can try that out. But the docs are confusing and vague:

>COU_LONG_IDLE_TIMEOUT - a longer version of COU_STANDARD_IDLE_TIMEOUT for applications that are known to need more time than usual to upgrade, such as Keystone and Octavia. The default value is 2400 seconds.

But as I mention in this issue's description, when Keystone's charm is upgraded, it waits for 300 secs (the default value for `COU_STANDARD_IDLE_TIMEOUT`, but might be a coincidence).

There being so many different timeouts for so many different phases, it'd be nice to have that well documented somewhere.

PR #160, opened 2023-11-02
increase waiting time

rgildein:
During testing of OpenStack cloud upgrade with COU, we found out that current waiting times are not enough. I increases all of times, since we raise an exception if any app and up in error state it's safe to have longer wait time, where most of that time will not be used.

fixes: #155

### DIFF
--- a/tests/unit/apps/test_auxiliary.py
+++ b/tests/unit/apps/test_auxiliary.py
@@ -130,7 +130,7 @@ def test_auxiliary_upgrade_plan_ussuri_to_victoria_change_channel(status, config
             ),
         ),
         UpgradeStep(
-            description=f"Wait 1800 s for model {model.name} to reach the idle state.",
+            description=f"Wait 1800s for model {model.name} to reach the idle state.",
             parallel=False,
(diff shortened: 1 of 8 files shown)


gabrielcocenza (approved):
LGTM

tests/unit/apps/test_auxiliary.py
```diff
@@ -130,9 +130,9 @@ def test_auxiliary_upgrade_plan_ussuri_to_victoria_change_channel(status, config
             ),
         ),
         UpgradeStep(
-            description=f"Wait 300 s for model {model.name} to reach the idle state.",
+            description=f"Wait 1800 s for model {model.name} to reach the idle state.",
```
agileshaw, line 133: Non-blocking: I find it a bit weird to look at when there's a space between `<num>` and `s`. I would suggest removing it in both code and tests:
```suggestion
            description=f"Wait 1800s for model {model.name} to reach the idle state.",
```

agileshaw (approved):

gabrielcocenza (approved):

agileshaw (approved):
LGTM. Thanks

Issue #153, opened 2023-10-30
UpgradeStep should have some hierarchy


rgildein:
Right now the UpgradeStep is used everywhere, even in place where cou refer to plan. That's why I think that cou should have UpgradePlan, UpgradeStep and potentially the BaseStep.

With the help of the Upgrade Plan, the COU can ensure that the functions required by the UpgradePlan will get it instead of UppgradePlan.

related with: #150

PR #163, opened 2023-11-05
Create separate classes for different types of steps

agileshaw:
- Rename UpgradeStep to BaseStep
- Add prompt and description_prefix attributes 
- Define UpgradePlan, ApplicationUpgradePlan, UpgradeStep, PreUpgradeStep, and PostUpgradeStep for different types of steps. This will also be used to improve execute stage.
- Change "Top level plan" description to the cloud upgrade 'from' and 'to' information

fixes: #153

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -25,7 +25,7 @@ from juju.errors import JujuError
 from cou.commands import parse_args
 from cou.exceptions import COUException, HighestReleaseAchieved, TimeoutException
 from cou.logging import setup_logging
-from cou.steps import BaseStep
+from cou.steps import UpgradePlan
 from cou.steps.analyze import Analysis
 from cou.steps.execute import apply_plan
 from cou.steps.plan import generate_plan, manually_upgrade_data_plane
@@ -86,7 +86,7 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
 
 async def analyze_and_plan(
     model_name: Optional[str], backup_database: bool
-) -> tuple[Analysis, BaseStep]:
+) -> tuple[Analysis, UpgradePlan]:
     """Analyze cloud and generate the upgrade plan with steps.
 
     :param model_name: Model name inputted by user.
@@ -94,7 +94,7 @@ async def analyze_and_plan(
     :param backup_database: Whether to create database backup before upgrade.
     :type backup_database: bool
     :return: Generated analyses and upgrade plan.
-    :rtype: tuple[Analysis, BaseStep]
+    :rtype: tuple[Analysis, UpgradePlan]
     """
     model = await COUModel.create(model_name)
     logger.info("Using model: %s", model.name)
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -25,7 +25,7 @@ from juju.errors import JujuError
 from cou.commands import parse_args
 from cou.exceptions import COUException, HighestReleaseAchieved, TimeoutException
 from cou.logging import setup_logging
-from cou.steps import UpgradeStep
+from cou.steps import BaseStep
 from cou.steps.analyze import Analysis
 from cou.steps.execute import apply_plan
 from cou.steps.plan import generate_plan, manually_upgrade_data_plane
@@ -86,7 +86,7 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
 
 async def analyze_and_plan(
     model_name: Optional[str], backup_database: bool
-) -> tuple[Analysis, UpgradeStep]:
+) -> tuple[Analysis, BaseStep]:
     """Analyze cloud and generate the upgrade plan with steps.
 
     :param model_name: Model name inputted by user.
@@ -94,7 +94,7 @@ async def analyze_and_plan(
     :param backup_database: Whether to create database backup before upgrade.
     :type backup_database: bool
     :return: Generated analyses and upgrade plan.
-    :rtype: tuple[Analysis, UpgradeStep]
+    :rtype: tuple[Analysis, BaseStep]
     """
     model = await COUModel.create(model_name)
     logger.info("Using model: %s", model.name)
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -240,7 +240,7 @@ class UpgradePlan(BaseStep):
     """Represents the upgrade plan.
 
     This class is intended to be used as a higher-level group for actual upgrade steps, therefore
-    doesn't accept corotine or parallel as inputs.
+    doesn't accept coroutine or parallel as inputs.
     """
 
     prompt: bool = False
@@ -260,17 +260,17 @@ class UpgradePlan(BaseStep):
     async def run(self) -> None:
         """Run UpgradePlan.
 
-        UpgradePlan should not have contain any corotine, so simply print a debug
+        UpgradePlan should not have contain any coroutine, so simply print a debug
         message to demonstrate the noop.
         """
-        logger.debug("No corotine to run for %s", repr(self))
+        logger.debug("No coroutine to run for %s", repr(self))
 
 
 class ApplicationUpgradePlan(UpgradePlan):
     """Represents the plan for application-level upgrade.
 
     This class is intended to be used as a group for application-level upgrade steps, therefore
-    doesn't accept corotine or parallel as inputs.
+    doesn't accept coroutine or parallel as inputs.
     """
 
     prompt: bool = True
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -71,7 +71,6 @@ class BaseStep:
         description: str = "",
         parallel: bool = False,
         coro: Optional[Coroutine] = None,
-        prompt: Optional[bool] = None,
     ):
         """Initialize BaseStep.
 
@@ -96,10 +95,6 @@ class BaseStep:
         self.sub_steps: List[BaseStep] = []
         self._canceled: bool = False
         self._task: Optional[asyncio.Task] = None
-
-        # Use default prompt value unless explicitly provided
-        if prompt is not None:
-            self.prompt = prompt
 
     def __hash__(self) -> int:
         """Get hash for BaseStep."""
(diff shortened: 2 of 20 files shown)


agileshaw:
Please refer to the `cou/steps/execute.py` of #164 to see how the changes in this PR is used.

cou/steps/__init__.py
```diff
@@ -171,14 +171,23 @@ def canceled(self) -> bool:
...
 
-    def add_step(self, step: UpgradeStep) -> None:
+    @property
+    def is_functionless(self) -> bool:
+        """Check if this is a functionless step.
+
+        :return: True if doesn't contain function. False otherwise.
+        :rtype: bool
+        """
+        return self._coro is None
```
gabrielcocenza, line 184: I didn't understand why we are adding this property if we already implemented `__bool__` that checks self._coro on itself and on the sub_steps. Is it not enough use what we already have?

cou/cli.py
```diff
@@ -86,15 +86,15 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
 
 async def analyze_and_plan(
     model_name: Optional[str], backup_database: bool
-) -> tuple[Analysis, UpgradeStep]:
+) -> tuple[Analysis, BaseStep]:
```
gabrielcocenza, line 89: If we are creating separate classes for different types of steps, doesn't make sense to have a new class to be the general plan? Something like `GeneralUpgradeStep`

gabrielcocenza (changes requested):

cou/cli.py
rgildein: We should have `UpgradePlan` and this function should refer to return that, because it does not make sense that it will return any other, e.g. Postupgrade, ...

cou/steps/__init__.py
```diff
@@ -127,25 +127,25 @@ def __str__(self) -> str:
...
 
     def __repr__(self) -> str:
-        """Representation of UpgradeStep.
+        """Representation of BaseStep.
 
-        :return: Representation of UpgradeStep.
+        :return: Representation of BaseStep.
         :rtype: str
         """
         return f"UpgradeStep({self.description})"
```
rgildein, line 135: I think that we should use `f"{self.__class__.__name__}({self.description})"`.

cou/steps/__init__.py
```diff
@@ -221,3 +230,61 @@ async def run(self) -> Any:
...
+    def description(self, description: str) -> None:
+        """Set the description of the PreUpgradeSubStep.
+
+        :param description: description
+        :type description: str
+        :raises ValueError: When a coroutine is passed without description.
+        """
+        if not description and self._coro:
+            raise ValueError("Every coroutine should have a description")
+        self._description = "[pre-upgrade] " + description
```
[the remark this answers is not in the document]
rgildein: if you check my draft PR #165, there is better solution to do overwrite the setter from base class or we can go with what @gabrielcocenza suggesting, only moving it to base class or defining `UpgradeStep` instead of `UpgradeSubStep`.

cou/steps/__init__.py
```diff
@@ -221,3 +230,61 @@ async def run(self) -> Any:
             return await self._task  # wait until task is completed
         except asyncio.CancelledError:  # ignoring asyncio.CancelledError
             logger.warning("Task %s was stopped unsafely.", repr(self))
+
+
+class ApplicationUpgradeStep(BaseStep):
```
rgildein, line 235: I think that `coro` should not be defined in BaseStep or we should by default set `None` for core in this class. So there will be no `ApplicationUpgradeStep` with `coro` defined. e,g. `ApplicationUpgradeStep(coro=my_coro())` will not be possible.

cou/steps/__init__.py
```diff
@@ -50,10 +50,10 @@ def compare_step_coroutines(coro1: Optional[Coroutine], coro2: Optional[Coroutin
     )
 
 
-class UpgradeStep:
+class BaseStep:
```
rgildein, line 53: We are missing `UpgradePlan`,  which should by similar to `ApplicationUpgradeStep`.

cou/steps/plan.py
```diff
@@ -64,18 +64,23 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
...
-    :rtype: UpgradeStep
+    :rtype: BaseStep
     """
     target = determine_upgrade_target(
         analysis_result.current_cloud_os_release, analysis_result.current_cloud_series
     )
-    print(f"Upgrading cloud from '{analysis_result.current_cloud_os_release}' to '{target}'\n.")
 
-    plan = UpgradeStep(description="Top level plan", parallel=False)
+    plan = BaseStep(
```
rgildein, line 73: I think that `BaseStep` should follow some abstraction, so it should be defined as abstract class and like that it should never be used directly.
This should be defined as `UpgradePlan(description=...)`, where `UpgradePlan` will set `parallel` to `False` and `coro` to `None`. I think that description should be only parameter for `UpgradePlan`, because it does not make sense to run in parallel or with any coroutine.

cou/steps/plan.py
```diff
@@ -64,18 +64,23 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
...
+            f"Upgrade cloud from '{analysis_result.current_cloud_os_release}' to '{target}'"
+        ),
+        parallel=False,
+    )
+
     if backup_database:
         plan.add_step(
-            UpgradeStep(
-                description="backup mysql databases",
+            BaseStep(
```
rgildein, line 82: Again `BaseStep` should never be used directly.

rgildein (changes requested):

cou/steps/__init__.py
```diff
@@ -51,29 +51,38 @@ def compare_step_coroutines(coro1: Optional[Coroutine], coro2: Optional[Coroutin
...
+    # pylint: disable=too-many-instance-attributes
+
+    prompt: bool = True  # whether to prompt for user input during execution
+
     def __init__(
         self,
         description: str = "",
         parallel: bool = False,
         coro: Optional[Coroutine] = None,
+        prompt: Optional[bool] = None,
```
rgildein, line 74: I do not understand what is benefit to have default value `None` and `True`  for class variable. Why using class variable and __init__ input variable? I think we should use only one and it should be class variable, since each type of step should behave same. e.g. `UpgradePlan` should not have prompt=False.
Maybe we will need to have some `UpgradeStep` which will have prompt=True, so in that class we may use prompt in as input variable for __init__ and change the default value, but this will be corner case situation which could be handled than.

cou/steps/__init__.py
```diff
@@ -232,45 +236,55 @@ async def run(self) -> Any:
...
+class UpgradeStep(BaseStep):
+    """Represents the upgrade step."""
+
+    sub_step_prefix = "[upgrade] "
+    prompt: bool = False
 
     @property
     def description(self) -> str:
-        """Get the description of the PostUpgradeSubStep.
+        """Get the description of the PreUpgradeStep.
```
rgildein, line 287: I think that you meant `UpgradeStep` and not `PreUpgradeStep`.

rgildein (changes requested):

cou/steps/__init__.py
agileshaw: As per our sync outcome, I remove prompt from init function and only let it being controlled as an class attribute. In case in the future we need an `UpgradeStep` with prompt, we can create a separate class, e.g. `CriticalUpgradeStep` and set its `prompt` value to True.

cou/steps/__init__.py
```diff
@@ -236,3 +244,82 @@ async def run(self) -> Any:
...
+    doesn't accept coroutine or parallel as inputs.
+    """
+
+    prompt: bool = True
+
+
+class UpgradeStep(BaseStep):
+    """Represents the upgrade step."""
+
+    sub_step_prefix = "[upgrade] "
```
rgildein, line 291: not-blocking: I think it would be cleaner to have `description_prefix`.

cou/steps/__init__.py
```diff
@@ -236,3 +244,82 @@ async def run(self) -> Any:
...
+    def description(self, description: str) -> None:
+        """Set the description of the UpgradeStep.
+
+        :param description: description
+        :type description: str
+        :raises ValueError: When a coroutine is passed without description.
+        """
+        if not description and self._coro:
+            raise ValueError("Every coroutine should have a description")
+        self._description = self.sub_step_prefix + description
```
rgildein, line 313: Can we use `description_prefix` (or what will be the name of class variable) in `BaseStep`? It means that no overwritten for setter is needed here.

cou/steps/__init__.py
```diff
@@ -236,3 +244,82 @@ async def run(self) -> Any:
...
+    This class is intended to be used as a higher-level group for actual upgrade steps, therefore
+    doesn't accept coroutine or parallel as inputs.
+    """
+
+    prompt: bool = False
+
+    def __init__(
+        self,
+        description: str = "",
+    ):
```
rgildein, line 261: not-blocking: Do we expect to have empty `UpgradePlan` or `ApplicationUpgradePlan`, without description?
```suggestion
    def __init__(self, description: str):
```

cou/steps/plan.py
```diff
@@ -64,18 +64,22 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
...
         analysis_result.current_cloud_os_release, analysis_result.current_cloud_series
     )
-    print(f"Upgrading cloud from '{analysis_result.current_cloud_os_release}' to '{target}'\n.")
 
-    plan = UpgradeStep(description="Top level plan", parallel=False)
+    plan = UpgradePlan(
+        description=(
+            f"Upgrade cloud from '{analysis_result.current_cloud_os_release}' to '{target}'"
+        ),
+    )
```
rgildein, line 77:
```suggestion
    plan = UpgradePlan(
        f"Upgrade cloud from '{analysis_result.current_cloud_os_release}' to '{target}'"
    )
```

rgildein (changes requested):
Small changes, but it looks good to me.

rgildein (approved):
LGTM

agileshaw:
After discussion with @rgildein, since this approach has been approved in a sync meeting and there's currently no availability for additional reviews, I'm merging it to unblock other tasks.

Issue #128, opened 2023-10-19
Don't prompt for the top-level plan


aieri:
Since the top level plan is just a container for the actual control/data plane upgrade paths, it doesn't make much sense to prompt the user about whether they want to continue or stop.

Issue #84, opened 2023-09-25
Progress indication for upgrade sub-steps in both interactive and non-interactive mode


agileshaw:
The progress indication is only implemented for the major stages, e.g. analyzing cloud, planning cloud, and running upgrades (in non-interactive mode). As an improvement, we could additionally display in progress status for upgrade sub-steps (e.g. refresh packages, switch channels) when running in both interactive and non-interactive mode.

aieri:
As part of this work, let's ensure that the output is resilient against users mashing the enter key. At the moment the output gets confusing:
```
Upgrade software packages of 'keystone' from the current APT repositories (c)ontinue/(a)bort/(s)kip:c




Refresh 'keystone' to the latest revision of 'ussuri/stable' (c)ontinue/(a)bort/(s)kip:Refresh 'keystone' to the latest revision of 'ussuri/stable' (c)ontinue/(a)bort/(s)kip:Refresh 'keystone' to the latest revision of 'ussuri/stable' (c)ontinue/(a)bort/(s)kip:Refresh 'keystone' to the latest revision of 'ussuri/stable' (c)ontinue/(a)bort/(s)kip:Refresh 'keystone' to the latest revision of 'ussuri/stable' (c)ontinue/(a)bort/(s)kip:c
```

Issue #76, opened 2023-09-21
Checking if upgrade to target is successful should be dropped from plan output and be a mandatory part of the upgrade step itself


agileshaw:
A feedback from demo: 

Checking if an application has been successfully upgraded to a target release should be mandatory validation of the upgrade process, not a skip-able step in the plan. Therefore, we should drop this step from the plan output and integrate it into the upgrade step.

aieri:
We could also consider dropping it from the plan output. Does showing it really add any value?

agileshaw:
Actually, I was implying that plan should no longer contain it as a step and make it as a mandatory part of the upgrade step itself. Will rephrase the description to make it more clear.

PR #164, opened 2023-11-05
Drop skip option and improve 'plan' and 'run' UI

agileshaw:
- Drop skip option and use "Continue(y/N)" instead
- For application level upgrade, group all of its substeps and prompt
only once.
- For functionless steps (except ApplicationUpgradeStep) and UpgradeSubSteps
(including its child classes), automatically proceed without prompting
- Add progress indicator for eligible steps

fixes: #76
fixes: #82
fixes: #84
fixes: #128


agileshaw:
Unit tests to be added after getting approval for on the approach.

gabrielcocenza (changes requested):
I don't know exactly where the bug is, but it's not working as expected:

```
Analyzing cloud... ✔
Generating upgrade plan... ✔
		Upgrade plan for 'my-designate-bind' to victoria
			[pre-upgrade] Upgrade software packages of 'my-designate-bind' from the current APT repositories
			Upgrade 'my-designate-bind' to the new channel: 'victoria/stable'
			[post-upgrade] Wait 120 s for app my-designate-bind to reach the idle state.
			[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded
		Upgrade plan for 'designate-bind' to victoria
			[pre-upgrade] Upgrade software packages of 'designate-bind' from the current APT repositories
			[post-upgrade] Wait 120 s for app designate-bind to reach the idle state.
			[post-upgrade] Check if the workload of 'designate-bind' has been upgraded

Upgrade plan for 'my-designate-bind' to victoria
	[pre-upgrade] Upgrade software packages of 'my-designate-bind' from the current APT repositories
	Upgrade 'my-designate-bind' to the new channel: 'victoria/stable'
	[post-upgrade] Wait 120 s for app my-designate-bind to reach the idle state.
	[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded

Continue(y/N):y
[pre-upgrade] Upgrade software packages of 'my-designate-bind' from the current APT repositories ✔
Upgrade 'my-designate-bind' to the new channel: 'victoria/stable' ✔
[post-upgrade] Wait 120 s for app my-designate-bind to reach the idle state. ✔
[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded ✔
[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded \Upgrade plan for 'designate-bind' to victoria
	[pre-upgrade] Upgrade software packages of 'designate-bind' from the current APT repositories
	[post-upgrade] Wait 120 s for app designate-bind to reach the idle state.
	[post-upgrade] Check if the workload of 'designate-bind' has been upgraded

[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded |
No valid input provided!
Upgrade plan for 'designate-bind' to victoria
	[pre-upgrade] Upgrade software packages of 'designate-bind' from the current APT repositories
	[post-upgrade] Wait 120 s for app designate-bind to reach the idle state.
	[post-upgrade] Check if the workload of 'designate-bind' has been upgraded

[post-upgrade] Check if the workload of 'my-designate-bind' has been upgraded \
[pre-upgrade] Upgrade software packages of 'designate-bind' from the current APT repositories ✔
[post-upgrade] Wait 120 s for app designate-bind to reach the idle state. |
[post-upgrade] Wait 120 s for app designate-bind to reach the idle state. ✔
[post-upgrade] Check if the workload of 'designate-bind' has been upgraded ✔
Upgrade completed.ck if the workload of 'designate-bind' has been upgraded -
```

After the upgrade steps of `my-designate-bind` it kept the message `Check if the workload of 'my-designate-bind' has been upgraded` instead of asking the input from the user.

rgildein:
Can we convert this to draft, it do not make sense to review it before #163.

cou/steps/execute.py
```diff
@@ -89,25 +91,53 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
...
     while result.casefold() not in AVAILABLE_OPTIONS:
-        result = (await ainput(prompt(plan.description))).casefold() if interactive else "c"
+        if (
+            not interactive
+            # sub-steps has been grouped under ApplicationUpgradeStep, so no need to prompt
+            # options for them individually
+            or isinstance(plan, UpgradeSubStep)
+            # Except for ApplicationUpgradeStep, which needs to prompt options to user,
+            # automatically proceed functionless without prompting options to user
+            or (plan.is_functionless and not isinstance(plan, ApplicationUpgradeStep))
```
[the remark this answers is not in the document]
rgildein: I think that we could define `prompt` variable for any step and by default `UpgradePlan` (`GeneralUpradeStep`) and `UpgradeSubStep` could have it set to `False`.

cou/steps/execute.py
```diff
@@ -54,29 +55,30 @@ def normal(text: str) -> str:
...
-    await step.run()
+    if indicate_progress:
+        progress_indicator.start(step.description)
+        await step.run()
+        progress_indicator.succeed()
+    else:
+        await step.run()
+
+    if indicate_progress:
+        progress_indicator.start(step.description)
```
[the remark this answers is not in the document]
agileshaw: Indeed. Thanks for pointing it out. This seems to also be the cause of the above mentioned bug

cou/steps/execute.py
```diff
@@ -89,25 +91,53 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
...
+            result = await ainput(prompt(description))
+
         match result:
-            case "c":
+            case "y":
                 logger.info("Running: %s", plan.description)
-                await _run_step(plan, interactive)
-            case "a":
+                await _run_step(plan, interactive, indicate_progress)
+            case "N":
```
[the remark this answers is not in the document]
agileshaw: I was thinking we should probably put more friction for user to abort the upgrade. But after reconsidering, I think we should keep it like other CLI tools so it matches with user's expectation. Change has been made in the new patch.

cou/steps/__init__.py
```diff
@@ -221,3 +234,83 @@ async def run(self) -> Any:
             return await self._task  # wait until task is completed
         except asyncio.CancelledError:  # ignoring asyncio.CancelledError
             logger.warning("Task %s was stopped unsafely.", repr(self))
+
+
+class UpgradePlan(BaseStep):
+    """Represents the upgrade plan.
+
+    This class is intended to be used as a higher-level group for actual upgrade steps, therefore
+    doesn't accept corotine or parallel as inputs.
```
dashmage, line 243: should be `coroutine`

cou/steps/execute.py
```diff
@@ -54,29 +55,27 @@ def normal(text: str) -> str:
...
-    return (
-        normal(parameter + " (")
-        + bold("c")
-        + normal(")ontinue/(")
-        + bold("a")
-        + normal(")bort/(")
-        + bold("s")
-        + normal(")kip:")
-    )
+    return normal(parameter + "Continue(") + bold("y") + normal("/") + bold("N") + normal("):")
```
rgildein, line 60: I think that we should use `Continue (y/n)` instead of capital `N`, because capital looks like default value, what is not true.

cou/steps/execute.py
```diff
@@ -89,25 +88,45 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
...
+async def apply_plan(plan: BaseStep, interactive: bool) -> None:
     """Apply the plan for upgrade.
 
     :param plan: Plan to be executed on steps.
-    :type plan: UpgradeStep
+    :type plan: BaseStep
     :param interactive:
     :type interactive: bool
     """
+    description = plan.description + " "
```
rgildein, line 99: Why we are adding this empty space `" "` here?

cou/steps/execute.py
```diff
@@ -89,25 +88,45 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
...
+
+    # group and print all sub-steps with hierarchy for ApplicationUpgradePlan
+    if isinstance(plan, ApplicationUpgradePlan):
+        description = str(plan) + "\n"
+        if not interactive:
+            print(plan.description)
+
     while result.casefold() not in AVAILABLE_OPTIONS:
-        result = (await ainput(prompt(plan.description))).casefold() if interactive else "c"
+        if not interactive or plan.prompt is False:
```
rgildein, line 118:
```suggestion
        if interactive is True or plan.prompt is False:
```

cou/steps/execute.py
```diff
@@ -54,29 +55,27 @@ def normal(text: str) -> str:
...
-    :type step: UpgradeStep
+    :type step: BaseStep
     :param interactive: Whether to run upgrade step in interactive mode.
     :type interactive: bool
+    :param indicate_progress: Whether to enable progress indicator for running the step.
+    :type indicate_progress: bool
     """
-    logger.debug("running step %s", step)
-    await step.run()
+    if indicate_progress:
```
rgildein, line 73: I think that progress indicator should be done only for `UpgradeStep`, so there is no need to have `indicate_progress`.

cou/steps/execute.py
```diff
@@ -89,25 +88,45 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
             await apply_plan(sub_step, interactive)
 
 
-async def apply_plan(plan: UpgradeStep, interactive: bool) -> None:
+async def apply_plan(plan: BaseStep, interactive: bool) -> None:
```
rgildein, line 91: Can we rename it to:
```suggestion
async def apply_step(step: BaseStep, interactive: bool) -> None:
```

cou/steps/execute.py
```diff
@@ -89,25 +88,45 @@ async def _run_step(step: UpgradeStep, interactive: bool) -> None:
...
+    # do nothing for empty upgrade step
+    if not plan:
+        return
+
+    # for UpgradePlan and its children class, do not show progress indicator
+    if isinstance(plan, UpgradePlan):
+        indicate_progress = False
+
+    # group and print all sub-steps with hierarchy for ApplicationUpgradePlan
+    if isinstance(plan, ApplicationUpgradePlan):
```
rgildein, line 112: If I understand it correctly, output of this will look like this:
```bash
Upgrade plan for 'rabbitmq-server' to victoria
    Upgrade software packages of 'rabbitmq-server' from the current APT repositories
    Upgrade 'rabbitmq-server' to the new channel: '3.9/stable'
    Change charm config of 'rabbitmq-server' 'source' to 'cloud:focal-victoria'
    Wait 300 s for model openstack to reach the idle state.
    Check if the workload of 'rabbitmq-server' has been upgraded
Continue(y/N)
Upgrade software packages of 'rabbitmq-server' from the current APT repositories ✔
Upgrade 'rabbitmq-server' to the new channel: '3.9/stable' ✔
...
```

Do you think printing full description in prompt is needed, since we will have indicator printing description of each sub-step.

rgildein (changes requested):
Can you provide usage example in PR descriptions, thanks.

cou/steps/execute.py
agileshaw: Do you mean `if interactive is False ...`? If interactive is true we should prompt for input instead of setting result to `y`

cou/steps/execute.py
agileshaw: For cosmetic purposes in generic cases. Otherwise if we have a single step that needs prompt, it will look like: `some step descriptionContinue(y/n)`

cou/steps/execute.py
agileshaw: Yes I think it is necessary to let the user know what will be performed before they say `yes` to it.

cou/steps/execute.py
agileshaw: That's true. The initial idea for reworking steps was different so I implemented a generic solution for progress indicator. After merging #163, I will update this accordingly.

cou/steps/execute.py
rgildein: Like this UX will be to complicated and I think it's really not necessary, because the whole plan will be print before applying. If you can share a full example in PR description we can discuss it again.

cou/steps/execute.py
rgildein: Sorry yes, I meant `interactive is False` only because you are doing `plan.prompt is False` and I prefer to stick with same usage. It means that you can use also `if not interactive or not plan.prompt`.

cou/steps/execute.py
rgildein: I will prefer to have `" Continue(y/n)"` instead of adding `" "` here.  would that be possible? If not can you add some note here why it's there. Thanks

agileshaw: The reason I didn't put the space before `Continue(y/n)` is to avoid unwanted indent in case of ApplicationUpgradePlan where `Continue(y/n)` is on a new line. I will go with your second suggestion and put a note here.

rgildein:
There is one ongoing discussion about UX, but other than that it's LGTM.

dashmage (dismissed):
LGTM :+1:

cou/steps/execute.py
```diff
@@ -54,60 +55,73 @@ def normal(text: str) -> str:
...
     :type interactive: bool
     """
-    logger.debug("running step %s", step)
-    await step.run()
+    if isinstance(step, UpgradeStep):
+        progress_indicator.start(step.description)
+        await step.run()
+        progress_indicator.succeed()
+    else:
+        await step.run()
```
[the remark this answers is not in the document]
agileshaw: IMO execution step should not have the knowledge of which type of steps has `_coro`; instead, it should just know which one needs progress indicator enabled. So personally I would prefer keeping the current implementation.

cou/steps/execute.py
```diff
@@ -54,60 +55,73 @@ def normal(text: str) -> str:
...
-    :type plan: BaseStep
+    :param step: Step to be executed.
+    :type step: BaseStep
     :param interactive:
     :type interactive: bool
     """
+    # adding a space at the end to better separate description with prompt options
+    # Example:
+    #   This is an upgrade step Continue(y/n)
+    description = step.description + " "
```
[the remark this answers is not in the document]
agileshaw: It is not being used right now, but I think it is a good practice to have a correct fallback UI for steps with single-line description and needs to be prompted (very likely to be added in the future). Otherwise, when we implement that kind of step, the UI will need additional adjustment. 

I'm changing the comments to the following:
```
    # Default description style for prompt for steps with single line description
    # Example:
    #   This is an upgrade step Continue (y/n): 
    description_to_prompt = step.description + " "
```
Would this be enough for addressing your concern here?

cou/steps/execute.py
```diff
@@ -54,60 +55,73 @@ def normal(text: str) -> str:
...
         match result:
-            case "c":
-                logger.info("Running: %s", plan.description)
-                await _run_step(plan, interactive)
-            case "a":
+            case "y":
+                logger.info("Running: %s", step.description)
+                await _run_step(step, interactive)
+            case "n":
                 logger.info("Aborting plan")
```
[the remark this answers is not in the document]
agileshaw: The problem with this wording change is that, we are actually not only aborting the current step, but the entire plan. So I'd like to keep the message as-is unless we can come up with something more accurate.

cou/steps/execute.py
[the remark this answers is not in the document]
gabrielcocenza: If there is no usage right now, I'm in favor of removing it and let the `\n` be the standard. When we have a situation where it prompts with single line, let's add it.

cou/steps/execute.py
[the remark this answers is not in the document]
gabrielcocenza: Yeah, reflecting more about it I think you are right. Let's keep it the way it is.

cou/steps/execute.py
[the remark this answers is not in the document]
gabrielcocenza: Ok. I see your point.

cou/steps/execute.py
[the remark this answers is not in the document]
agileshaw: After some thinking, I have moved the styling to `prompt` function and always display `Continue (y/n): ` on a new line.

cou/steps/execute.py
```diff
@@ -57,68 +57,86 @@ def normal(text: str) -> str:
         """
         return Style.RESET_ALL + text + Style.RESET_ALL
 
-    return normal(parameter + "Continue(") + bold("y") + normal("/") + bold("n") + normal("):")
+    return normal(parameter + "\nContinue (") + bold("y") + normal("/") + bold("n") + normal("): ")
 
 
-async def _run_step(step: BaseStep, interactive: bool) -> None:
+async def _run_step(step: BaseStep, interactive: bool, overwrite_progress: bool = False) -> None:
```
gabrielcocenza, line 63: What do you think about isolating the logic to decide if should overwrite_progress or not in a separate function and not have as parameter of `_run_step` and `apply_step`?

I think it will be cleaner and we can easily maintain.

agileshaw: I don't see a way of easily doing so because this logic requires some recursive check (all sub-steps of ApplicationUpgradePlan's sub-steps would need to set it to `True`). Plus the actual logic is a one-line code with clear comment, so I think it is ok to keep the implementation as-is.

cou/steps/execute.py
```diff
@@ -57,68 +57,86 @@ def normal(text: str) -> str:
         """
         return Style.RESET_ALL + text + Style.RESET_ALL
 
-    return normal(parameter + "Continue(") + bold("y") + normal("/") + bold("n") + normal("):")
+    return normal(parameter + "\nContinue (") + bold("y") + normal("/") + bold("n") + normal("): ")
```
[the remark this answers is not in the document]
agileshaw: Personally I don't feel like it is necessary because the colored check mark indicates it is part of a previous step. But if you think it would look better having a new line added in-between, I can add it.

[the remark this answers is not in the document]
gabrielcocenza: It's really no big deal and no blocker, but I think the display is better with an empty line between applications upgrade.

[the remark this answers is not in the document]
agileshaw: Added!

rgildein (approved):
LGTM

gabrielcocenza (approved):

Issue #154, opened 2023-10-30
COU should not wait for whole model


rgildein:
Right now the COU wait for hole model for ceph-mon, keyston and ceph-mon, however this can cause in issue, because there could be application not in idle state (e.g. telegraf), which are not relevant to OpenStack.
This can cause unnecessary waiting and eventually failing of COU.

related: #78

PR #168, opened 2023-11-14
Wait only for apps related with COU

rgildein:
Add `list_applications` function to `COUModel`, which returns all applications related with COU based on if charm is supported by COU. Use this function in `wait_for_idle` if no app was provided. Previously function was waiting for whole model if no apps was provided, but now it will wait only for apps related with COU.

fixes: #154

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -185,7 +185,7 @@ class COUModel:
 
         :param action: Action object
         :type: Action
-        :param raise_on_failure: Raise ActionFailed exception on failure, defaults to False
+        :param raise_on_failure: Whether to raise ActionFailed exception on failure, defaults to False
         :type raise_on_failure: bool
         :return: action results
         :rtype: Action
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -37,6 +37,7 @@ from cou.exceptions import (
     TimeoutException,
     UnitNotFound,
 )
+from cou.utils.openstack import is_charm_supported
 
 JUJU_MAX_FRAME_SIZE: int = 2**30
 DEFAULT_TIMEOUT: int = int(os.environ.get("COU_TIMEOUT", 10))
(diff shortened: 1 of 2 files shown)


cou/utils/juju_utils.py
```diff
@@ -259,27 +282,18 @@ async def get_status(self) -> FullStatus:
         model = await self._get_model()
         return await model.get_status()
 
-    @retry(no_retry_exceptions=(ActionFailed,))
-    async def _get_action_result(self, action: Action, raise_on_failure: bool) -> Action:
-        """Get results from action.
+    async def list_applications(self) -> dict[str, Application]:
```
gabrielcocenza, line 285: IMO the naming is confusing. I din't think we should have `list` on it and looking at the first glance it looks like that is all apps and not cou apps
```suggestion
    async def cou_applications(self) -> dict[str, Application]:
```

cou/utils/juju_utils.py
```diff
@@ -259,27 +282,18 @@ async def get_status(self) -> FullStatus:
...
-                output = await model.get_action_output(action.entity_id)
-                raise ActionFailed(action, output=output)
-
-        return result
+        model = await self._get_model()
+        return {
+            name: app
+            for name, app in model.applications.items()
+            if is_charm_supported(app.charm_name)
+        }
```
gabrielcocenza, line 296: We are not using the object `Application` so far, why to not generate the list directly?
```suggestion
        return [
            name
            for name, app in model.applications.items()
            if is_charm_supported(app.charm_name)
        ]
```

gabrielcocenza (changes requested):
Overall looks good to me, but I would suggest to rename the function and change the output of the function to what we actually will use. If in the future we need to access `Application` we can change the data structure to be a dictionary

cou/utils/juju_utils.py
rgildein: I though about changing it also after pushing, since return value is not list but dict. Renaming it to `cou_applications` do not make much sense, since it will be `COUModel.cou_applications`.

rgildein: What about simply call it `get_apps` or `get_related_apps`?

gabrielcocenza: Yeah. I agree that `COUModel.cou_applications` it's not great. I'm ok with `get_apps` or maybe `apps_supported`

rgildein: I will go with `get_supported_apps`, if that's ok with you.

gabrielcocenza: Yes. That is ok to me. Thanks

agileshaw (changes requested):

gabrielcocenza (approved):

agileshaw (dismissed):
LGTM, thanks

gabrielcocenza (dismissed):

agileshaw (approved):

gabrielcocenza (approved):

PR #169, opened 2023-11-14
Filter libjuju and websockets debug log from logfile.

gabrielcocenza:
- add those filtered logs on the log file if user uses max verbosity.

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -93,9 +93,9 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         action="count",
         dest="verbosity",
         help="Increase logging verbosity in STDOUT. Multiple 'v's yield progressively "
-        "more detail (up to 4).\nNote that by default cou use debug level at the logfile."
-        "However, at the maximum verbosity it will include debug logs from libjuju and "
-        "websockets that can generate a big logfile.",
+        "more detail (up to 4).\nNote that by default the logfile will include debug logs from "
+        "all modules with the exception of libjuju and websockets. To include the debug logs from "
+        "the excluded modules, use the maximum verbosity.",
     )
     group.add_argument(
         "--quiet",
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -93,9 +93,10 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         action="count",
         dest="verbosity",
         help="Increase logging verbosity in STDOUT. Multiple 'v's yield progressively "
-        "more detail (up to 4).\nNote that by default the logfile will include debug logs from "
-        "all modules with the exception of libjuju and websockets. To include the debug logs from "
-        "the excluded modules, use the maximum verbosity.",
+        "more detail (up to 4).\nNote that by default the logfile will include standard "
+        "logs from juju and websockets, as well as debug logs from all other modules. "
+        "To also include the debug level logs from juju and websockets modules, use the "
+        "maximum verbosity.",
     )
     group.add_argument(
         "--quiet",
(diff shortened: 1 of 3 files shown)


cou/logging.py
```diff
@@ -71,6 +71,26 @@ def setup_logging(log_level: str = "INFO") -> None:
...
+    can be huge if not filtered.
+    :param record: A LogRecord instance represents an event being logged.
+    :type record: LogRecord
+    :return: Returns false to not append record in the log file, true for appending it.
+    :rtype: bool
+    """
+    return not (
+        (record.name.startswith("juju.") or record.name.startswith("websockets."))
+        and record.levelname == "DEBUG"
+    )
```
rgildein, line 96: I think it will be effective to check first if record.levelname is DEBUG to avoid doing `startswith` twice for no debug messages.
```suggestion
    return (
        record.levelname != "DEBUG"
        or not (record.name.startswith("juju.") or record.name.startswith("websockets."))
    )
```

rgildein (changes requested):
Overall looks good to me, but I would suggest to switch order in filter function to be more effective.

rgildein (approved):
LGTM

tests/unit/test_logging.py
```diff
@@ -39,7 +43,41 @@ def test_setup_logging():
...
+            log_file_handler.addFilter.assert_not_called()
+        else:
+            log_file_handler.addFilter.assert_called_with(filter_debug_logs)
+
+
+@pytest.mark.parametrize(
+    "name, level, exp_result",
+    [
+        ("juju.client.connection", "DEBUG", False),  # juju debug is not logged
+        ("juju.client.connection", "INFO", True),  # juju debug is not logged
```
agileshaw, line 61:
```suggestion
        ("juju.client.connection", "INFO", True),  # juju info is logged
```

cou/commands.py
```diff
@@ -93,8 +93,9 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         action="count",
         dest="verbosity",
         help="Increase logging verbosity in STDOUT. Multiple 'v's yield progressively "
-        "more detail (up to 4).\nNote that this doesn't affect the verbosity in logfile, "
-        "which will always have the maximum verbosity.",
+        "more detail (up to 4).\nNote that by default cou use debug level at the logfile."
+        "However, at the maximum verbosity it will include debug logs from libjuju and "
+        "websockets that can generate a big logfile.",
```
agileshaw, line 98: The description here suggests that by default only COU logs at debug (and up) level will be included in the logfile. But that doesn't align with the actual implementation where external logs that are not in debug level will also be included in general cases. We should change the wording here to reflect this.

agileshaw (changes requested):

cou/commands.py
gabrielcocenza: Do you have a suggestion on how to re word it?

cou/commands.py
```diff
@@ -93,8 +93,9 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         action="count",
         dest="verbosity",
         help="Increase logging verbosity in STDOUT. Multiple 'v's yield progressively "
-        "more detail (up to 4).\nNote that this doesn't affect the verbosity in logfile, "
-        "which will always have the maximum verbosity.",
+        "more detail (up to 4).\nNote that by default the logfile will include debug logs from "
+        "all modules with the exception of libjuju and websockets. To include the debug logs from "
+        "the excluded modules, use the maximum verbosity.",
```
agileshaw, line 98:
```suggestion
        "more detail (up to 4).\nNote that by default the logfile will include standard logs from juju and websockets, as well as debug logs from "
        "all other modules. To also include the debug level logs from "
        "juju and websockets modules, use the maximum verbosity.",
```

May need a bit more adjustments in wording, but the idea is to correctly reflect what is implemented.

agileshaw (approved):
LGTM, thanks

rgildein (approved):
LGTM

Issue #152, opened 2023-10-30
Usage of `COU_DATA` is impossible with snap


rgildein:
Using cou as snap make `COU_DATA` unusable, because the snap has no permission to user home directory.

```bash
$ cou_rg run -v                                                                                                                                                                        
Unexpected error occurred                                                                                                                                                                                          
[Errno 13] Permission denied: '/home/ubuntu/rgildein/log'                                                                                                                                                          
Traceback (most recent call last):                                                                                                                                                                                 
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/cli.py", line 186, in entrypoint                                                                                                      
    setup_logging(log_level)
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/logging.py", line 60, in setup_logging
    pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
  File "/usr/lib/python3.10/pathlib.py", line 1175, in mkdir
    self._accessor.mkdir(self, mode)
PermissionError: [Errno 13] Permission denied: '/home/ubuntu/rgildein/log'
```

I think we should support usage of hoe directory and catching exception above with propert message how user can enable it.

agileshaw:
we might need to remove COU_DATA

PR #170, opened 2023-11-16
Remove COU_DATA env var

gabrielcocenza:
- the snap don't have permission to set the data on paths other than /home/$USER/.local/share
- added pytest_env so unit tests can have access to $HOME which is pointed to ~/tmp

Fix: #152

### DIFF
--- a/cou/utils/__init__.py
+++ b/cou/utils/__init__.py
@@ -17,4 +17,7 @@
 import os
 from pathlib import Path
 
+from halo import Halo
+
 COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou") if os.getenv("USER") else Path(".")
+progress_indicator = Halo(spinner="line", placement="right")
--- a/cou/utils/__init__.py
+++ b/cou/utils/__init__.py
@@ -17,4 +17,4 @@
 import os
 from pathlib import Path
 
-COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou")
+COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou") if os.getenv("USER") else Path(".")
--- a/cou/utils/__init__.py
+++ b/cou/utils/__init__.py
@@ -14,8 +14,7 @@
 
 """Utilities for charmed-openstack-upgrader."""
 
-
 import os
 from pathlib import Path
 
-COU_DATA = Path(os.getenv("COU_DATA", "."))
+COU_DATA = Path(f"{os.getenv('HOME')}/.local/share/cou")
--- a/cou/utils/__init__.py
+++ b/cou/utils/__init__.py
@@ -17,4 +17,4 @@
 import os
 from pathlib import Path
 
-COU_DATA = Path(f"{os.getenv('HOME')}/.local/share/cou")
+COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou")
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -14,7 +14,7 @@
 
 from collections import OrderedDict
 from pathlib import Path
-from unittest.mock import AsyncMock, MagicMock, PropertyMock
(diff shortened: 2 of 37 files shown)


cou/utils/__init__.py
```diff
@@ -14,8 +14,7 @@
 
 """Utilities for charmed-openstack-upgrader."""
 
-
 import os
 from pathlib import Path
 
-COU_DATA = Path(os.getenv("COU_DATA", "."))
+COU_DATA = Path(f"{os.getenv('HOME')}/.local/share/cou")
```
agileshaw, line 20: Putting an external discussion here for future references: `$HOME` variable has a different value inside non-classic snaps ([ref](https://snapcraft.io/docs/environment-variables#heading--home)), so logs would not be saved to the expected system path. Perhaps we should consider save logs to a directory inside $SNAP_COMMON or $SNAP_DATA instead.

tests/unit/test_logging.py
```diff
@@ -32,7 +32,10 @@ def test_filter_clears_exc_info_and_text():
 
 def test_setup_logging():
     """Test setting up logging."""
-    with patch("cou.logging.logging") as mock_logging:
+    with (
+        patch("cou.logging.logging") as mock_logging,
+        patch("cou.logging.pathlib"),
```
rgildein, line 37: I think that `COU_DATA` should be defined as auto fixture for whole unit tests, so we can be sure that no data are written to `/home/<user>/.local/share/cou`.

cou/utils/__init__.py
```diff
@@ -14,8 +14,7 @@
 
 """Utilities for charmed-openstack-upgrader."""
 
-
 import os
 from pathlib import Path
 
-COU_DATA = Path(os.getenv("COU_DATA", "."))
+COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou")
```
rgildein, line 20: Will this work fine with snap and sanp environment variables?

rgildein (changes requested):

cou/utils/__init__.py
gabrielcocenza: The problem of using [SNAP_COMMON or SNAP_DATA](https://snapcraft.io/docs/environment-variables#heading--snap-common)  it's writable just for root and that is not our case. If we were running a daemon, than this could be done

tests/unit/test_logging.py
gabrielcocenza: Done

cou/utils/__init__.py
gabrielcocenza: Yes it's working as expected. I've manually tested and reading the documentation I didn't find information like the `$HOME` that is override.

tox.ini
```diff
@@ -9,7 +9,9 @@ skip_missing_interpreters = True
 
 [testenv]
 basepython = python3
-setenv = PYTHONPATH={toxinidir}
+setenv =
+    PYTHONPATH={toxinidir}
+    USER={env:USER}
```
rgildein, line 14: I think using `passenv = USER` is better than re-setting the env variable.

tests/unit/conftest.py
```diff
@@ -568,3 +568,14 @@ def apps(status, config, model):
...
     }
+
+
+@pytest.fixture(scope="session", autouse=True)
+def cou_data(tmp_path_factory):
+    cou_test = tmp_path_factory.mktemp("cou_test")
+    with (
+        patch("cou.utils.COU_DATA", cou_test),
+        patch("cou.logging.COU_DATA", cou_test),
+        patch("cou.steps.backup.COU_DATA", cou_test),
```
rgildein, line 579: Are these two patches really needed? I think that patching only source of COU_DATA should be enough.

agileshaw (changes requested):

agileshaw (approved):
LGTM

rgildein (approved):
LGTM

Issue #127, opened 2023-10-19
Output the model name we're operating on


aieri:
If `--model` is not specified, cou will autodetermine the current juju model. This is however implicit, so I think we should add to the output which model we're actually going to work in.

PR #171, opened 2023-11-20
using current model if no model_name provided

rgildein:
- use indicator for model connection
- COUModel use current model name from Juju if no model_name provided

fixes: #127

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -37,6 +37,7 @@ from cou.exceptions import (
     TimeoutException,
     UnitNotFound,
 )
+from cou.utils.openstack import is_charm_supported
 
 JUJU_MAX_FRAME_SIZE: int = 2**30
 DEFAULT_TIMEOUT: int = int(os.environ.get("COU_TIMEOUT", 10))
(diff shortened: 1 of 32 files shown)


cou/utils/juju_utils.py
```diff
@@ -164,21 +151,15 @@ def connected(self) -> bool:
     @property
     def name(self) -> str:
         """Return model name."""
+        if self.connected:
+            return self._model.name
+
         if self._name is None:
-            self._name = self._model.name
+            # pylint: disable=protected-access
+            self._name = self._model._connector.jujudata.current_model(model_only=True)
```
gabrielcocenza, line 159: This looks like a  little hacky to me. What if we instantiate a new [FileJujuData](https://github.com/juju/python-libjuju/blob/f91507a94d7fd8ca56a50e60416d29afe2c64575/juju/client/jujudata.py#L87) by passing the output of  [juju_config_dir](https://github.com/juju/python-libjuju/blob/master/juju/utils.py#L40)?

I think it's more explicit

rgildein: I agree that it will be more explicit, but the model can be initialized with different JujuData and than you will simply provided different current model. Like this if we define model another way this part will raise an exception.

agileshaw: I can see the pros and cons from both approaches, so I don't have a strong opinion here. But for personal preferences, I'm leaning towards Gabriel's suggestion more because accessing a private variable composes more risks of code breakage.

cou/utils/juju_utils.py
```diff
@@ -136,22 +137,10 @@ class COUModel:
 
     def __init__(self, name: Optional[str]):
         """COU Model initialization with name and juju.model.Model."""
-        self._model = Model(max_frame_size=JUJU_MAX_FRAME_SIZE)
+        self._juju_data = FileJujuData()
+        self._model = Model(max_frame_size=JUJU_MAX_FRAME_SIZE, jujudata=self._juju_data)
```
gabrielcocenza, line 141: Can't we use the property directly?
```suggestion
        self._model = Model(max_frame_size=JUJU_MAX_FRAME_SIZE, jujudata=self.juju_data)
```

cou/utils/juju_utils.py
```diff
@@ -161,24 +150,23 @@ def connected(self) -> bool:
...
+
     @property
     def name(self) -> str:
         """Return model name."""
+        if self.connected:
+            return self._model.name
+
         if self._name is None:
-            self._name = self._model.name
+            # pylint: disable=protected-access
```
gabrielcocenza, line 165: I think we can remove this now
```suggestion
```

gabrielcocenza (approved):

gabrielcocenza (approved):

agileshaw (approved):
LGTM

PR #177, opened 2023-11-29
Change from rabbitmq to designate-bind for smoke test.

gabrielcocenza:
- designate-bind uses less resources and reaches idle state without the need of relating with another charms.
- this application when alone in a model will generate upgrade plans. rabbitmq had the same workload version from ussuri to yoga and when alone, the next upgrade was pointing to jammy-yoga.
- added model connection status check test


agileshaw (approved):
LGTM. Please also update the PR description to also include the newly added model connection status checking.

rgildein (approved):
LGTM

PR #178, opened 2023-11-30
Separate snap build to different job

agileshaw:
- Create 'snap-build' job to build the snap and upload as artifact to be used in future functional tests
- Upgrade checkout action to v4

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -25,14 +25,14 @@ jobs:
     with:
       python-version: ${{ matrix.python-version }}
       tox-version: '<4'
-
-  func:
-    name: Functional tests
+    
+  snap-build:
+    name: Build and install pacakges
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -25,14 +25,45 @@ jobs:
     with:
       python-version: ${{ matrix.python-version }}
       tox-version: '<4'
+    
+  snap-build:
+    name: Build and install pacakges
```
rgildein, line 30: I think it would be better to have something like "Build snap package".

rgildein (approved):
One not blocking change to name of job, other than that +2. Thanks

rgildein (approved):
LGTM

gabrielcocenza (approved):
LGTM. Thanks

PR #179, opened 2023-11-30
Smoke test for cou plan

gabrielcocenza:
check the stdout for the following scenarios:

* plan with backup
* plan without backup
* plan choosing the model

### DIFF
--- a/tests/functional/tests/cou_smoke.py
+++ b/tests/functional/tests/cou_smoke.py
@@ -1,0 +1,116 @@
+import logging
+import os
+import unittest
+from subprocess import check_call, check_output
+
+import zaza
+
+log = logging.getLogger(__name__)
+
+
+class COUSmokeTest(unittest.TestCase):
+    """COU smoke functional tests."""
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.model_name = zaza.model.get_juju_model()
+        cls.install_package()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.remove_package()
+
+    def install_package() -> None:
+        """Install cou package."""
+        cou_test_snap = os.environ.get("TEST_SNAP")
+        if cou_test_snap:
+            log.info(f"Installing {cou_test_snap}")
+            assert os.path.isfile(cou_test_snap)
+            # install the snap
+            assert check_call(f"sudo snap install --dangerous {cou_test_snap}".split()) == 0
+            # connect interfaces
+            interfaces_to_connect = [
+                "juju-client-observe",
+                "dot-local-share-cou",
+                "ssh-public-keys",
+            ]
+            for interface in interfaces_to_connect:
+                assert (
+                    check_call(
+                        f"sudo snap connect charmed-openstack-upgrader:{interface} snapd".split()
+                    )
+                    == 0
+                )
+            # make the cou alias
+            assert check_call("sudo snap alias charmed-openstack-upgrader.cou cou".split()) == 0
+
+            assert check_output("which cou".split()).decode().strip() == os.path.join(
+                "/snap/bin/cou"
+            )
+        else:
+            log.warning("Installing python package")
+            assert check_call("python3 -m pip install .".split()) == 0
+            assert (
+                check_output("which cou".split())
+                .decode()
+                .strip()
+                .startswith(os.path.join(os.getcwd(), ".tox"))
+            )
+
+    def remove_package() -> None:
+        """Remove cou package."""
+        cou_test_snap = os.environ.get("TEST_SNAP")
+        if cou_test_snap:
+            log.info("Removing snap package cou")
+            check_call("sudo snap remove charmed-openstack-upgrader --purge".split())
+        else:
+            logging.info("Uninstalling python package cou")
+            check_call("python3 -m pip uninstall --yes charmed-openstack-upgrader".split())
+
+    def test_plan_backup(self) -> None:
+        """Test plan with backup."""
+        result = check_output("cou plan".split()).decode()
+        expected_plan = (
+            "Upgrade cloud from 'ussuri' to 'victoria'\n"
+            "\tVerify that all OpenStack applications are in idle state\n"
+            "\tBackup mysql databases\n"
+            "\tControl Plane principal(s) upgrade plan\n"
+            "\t\tUpgrade plan for 'designate-bind' to victoria\n"
+            "\t\t\tUpgrade software packages of 'designate-bind' "
+            "from the current APT repositories\n"
+            "\t\t\tUpgrade 'designate-bind' to the new channel: 'victoria/stable'\n"
+            "\t\t\tWait 300s for app designate-bind to reach the idle state.\n"
+            "\t\t\tCheck if the workload of 'designate-bind' has been upgraded\n"
+        )
+        self.assertIn(expected_plan, result)
+
+    def test_plan_no_backup(self) -> None:
+        """Test plan with no backup."""
+        result = check_output("cou plan --no-backup".split()).decode()
+        expected_plan = (
+            "Upgrade cloud from 'ussuri' to 'victoria'\n"
+            "\tVerify that all OpenStack applications are in idle state\n"
+            "\tControl Plane principal(s) upgrade plan\n"
+            "\t\tUpgrade plan for 'designate-bind' to victoria\n"
+            "\t\t\tUpgrade software packages of 'designate-bind' "
+            "from the current APT repositories\n"
+            "\t\t\tUpgrade 'designate-bind' to the new channel: 'victoria/stable'\n"
+            "\t\t\tWait 300s for app designate-bind to reach the idle state.\n"
+            "\t\t\tCheck if the workload of 'designate-bind' has been upgraded\n"
+        )
+        self.assertIn(expected_plan, result)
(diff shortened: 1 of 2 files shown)


tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+
+            # check that the executable path exists
+            assert Path("/snap/bin/cou").exists()
+            cls.exc_path = "/snap/bin/cou"
+
+        else:
+            # functest already installs cou as python package, but we install again
+            # for better developer experience
+            log.warning("using cou as python package")
+            assert check_call(f"python3 -m pip install {os.getenv('PYTHONPATH')}".split()) == 0
```
rgildein, line 64: There is no need to install anything, since installation should be handled by tox.

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+            cls.exc_path = "cou"
```
rgildein, line 65: I think that this can be ambitious, because if I installed cou with snap the `cou` will represent by this or not? Can we use something else like `./tox/func/bin/cou`?

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+    def create_local_share_folder() -> None:
+        """Create the .local/share/ folder if does not exist."""
+        Path(f"/home/{os.getenv('USER')}/.local/share/").mkdir(parents=True, exist_ok=True)
+
+    @classmethod
+    def install_package(cls) -> None:
+        """Install cou package."""
+        cou_test_snap = os.environ.get("TEST_SNAP")
+        if cou_test_snap:
+            log.info(f"Installing {cou_test_snap}")
```
rgildein, line 36:
```suggestion
            log.info("Installing %s", cou_test_snap)
```

rgildein (changes requested):

tests/functional/tests/smoke.py
gabrielcocenza: cou as python package is already installed via tox and it has preference because it's on the same virtual env. If you install the snap and call just by `cou` the snap won't be called. However, this is more declarative and I think we can make something as you mentioned

tox.ini
```diff
@@ -58,6 +58,7 @@ deps = .[functests]
 passenv =
     TEST_*
     OS_*
+    USER
```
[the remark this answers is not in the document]
gabrielcocenza: When you define at the top and then in another context, it override the original

tests/functional/tests/smoke.py
agileshaw: I think which executable path `cou` refers to is depending on which `cou` is installed first. In functional tests, python package is installed first through tox, `cou` will always point to the python executable. I agree that the path should be more explicit if possible, but in case we can't get the actual path in code (e.g. `./tox/func` vs `./tox/func-smoke` created by different tox target ), I think using just `cou` with some explanatory comments is fine

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+
+        :param cmd: Command to run.
+        :type cmd: list[str]
+        :return: Response of the command with the stderr on stdout.
+        :rtype: str
+        """
+        try:
+            return check_output([self.exc_path] + cmd, stderr=STDOUT).decode()
+        except CalledProcessError as err:
+            log.error(err.output)
```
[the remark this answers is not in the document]
agileshaw: Adding `stderr=subprocess.STDOUT` parameter to `check_output` will make sure both stdout and stderr are captured

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def test_plan_backup(self) -> None:
+        """Test plan with backup."""
+        result = self.cou(["plan", "--model", self.model_name])
+        expected_plan = (
+            "Upgrade cloud from 'ussuri' to 'victoria'\n"
+            "\tVerify that all OpenStack applications are in idle state\n"
+            "\tBackup mysql databases\n"
```
agileshaw, line 92: The fact that we are able to test backup plan in a model that doesn't contain mysql is weird to me (the same reason I opened #184 ). To not confuse people, I think we should at least rename the function to something like  `test_plan_default` to not mentioning specifically mention backup.

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,132 @@
...
+            "\t\tUpgrade plan for 'designate-bind' to victoria\n"
+            "\t\t\tUpgrade software packages of 'designate-bind' "
+            "from the current APT repositories\n"
+            "\t\t\tUpgrade 'designate-bind' to the new channel: 'victoria/stable'\n"
+            "\t\t\tWait 300s for app designate-bind to reach the idle state.\n"
+            "\t\t\tCheck if the workload of 'designate-bind' has been upgraded\n"
+        )
+        self.assertIn(expected_plan, result)
+
+    def test_plan_no_backup_choosing_model(self) -> None:
```
agileshaw, line 119: Since we need to always passing in model name, this test is no longer needed (duplicate of `test_plan_no_backup`)

tests/functional/tests/smoke.py
```diff
@@ -33,44 +35,51 @@ def install_package(cls) -> None:
...
-                    )
-                    == 0
+                check_call(
+                    [
+                        "sudo",
+                        "snap",
+                        "connect",
+                        f"charmed-openstack-upgrader:{interface}",
+                        "snapd",
+                    ]
```
agileshaw, line 58: Why are we removing the assertion for interface connection calls? I think it is useful catching error early; otherwise we would scratch our head wondering why following test cases aren't passing?

tests/functional/tests/smoke.py
```diff
@@ -17,7 +17,9 @@ def setUpClass(cls) -> None:
         zaza.get_or_create_libjuju_thread()
         cls.create_local_share_folder()
         cls.model_name = zaza.model.get_juju_model()
+        cls.package_installed = False
         cls.install_package()
+        cls.package_installed = True
```
agileshaw, line 22: Since this variable is only related to snap, maybe we should rename it to `cls.snap_installed` and move setting it to `True` to be under the snap installation part of `install_package()` method? This way, we can also simplify the condition in `remove_snap_package()` method to be `if cls.snap_installed:`

tests/functional/tests/smoke.py
gabrielcocenza: Actually it was a mistake adding it in the first place. As the [documentation](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) says:
> If the return code was zero then return, otherwise raise [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError). The [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) object will have the return code in the [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError.returncode) attribute.

This means that the assert is never called. Ex:
```
assert check_call(["ls", "asdf"]) == 1, "my custom message"
ls: cannot access 'asdf': No such file or directory
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ls', 'asdf']' returned non-zero exit status 2.
```
I'll remove all asserts that are using `check_call`

agileshaw: Ah, this makes sense. :+1:

agileshaw (approved):
LGTM

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,156 @@
...
+class FuncSmokeException(Exception):
+    """Default Func Smoke exception."""
+
+
+class SmokeTest(unittest.TestCase):
+    """COU smoke functional tests."""
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        zaza.get_or_create_libjuju_thread()
```
rgildein, line 21: not-blocking: Do we nee this? we are using zaza only to get model name.

tests/functional/tests/smoke.py
```diff
@@ -0,0 +1,156 @@
...
+        cls.configure_executable_path()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.remove_snap_package()
+        zaza.clean_up_libjuju_thread()
+
+    def create_local_share_folder() -> None:
+        """Create the .local/share/ folder if does not exist."""
+        Path(f"/home/{os.getenv('USER')}/.local/share/").mkdir(parents=True, exist_ok=True)
```
rgildein, line 33: We should use `COU_DATA` and not defining it manually here.

rgildein (approved):
Two not blocking comments, but it looks good to me. Thanks

tests/functional/tests/smoke.py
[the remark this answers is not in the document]
rgildein: Yes, but we do not want to run check_output in KeyboardInterrupt interruption, since it failed and only result we get will be log message.
Now it was updated to use run :+1:.

rgildein (approved):
LGTM

agileshaw (approved):

PR #181, opened 2023-11-30
Smoke test for cou run

gabrielcocenza:
- check that designate-bind can upgrade from ussuri to victoria.

### DIFF
--- a/tests/functional/tests/cou_smoke.py
+++ b/tests/functional/tests/cou_smoke.py
@@ -114,3 +114,17 @@ class COUSmokeTest(unittest.TestCase):
             "\t\t\tCheck if the workload of 'designate-bind' has been upgraded\n"
         )
         self.assertIn(expected_plan, result)
+
+    def test_help(self) -> None:
+        """Test that help is working."""
+        assert check_call("cou --help".split()) == 0
+        assert check_call("cou --h".split()) == 0
+        assert check_call("cou plan --help".split()) == 0
+        assert check_call("cou plan -h".split()) == 0
+        assert check_call("cou run --help".split()) == 0
+        assert check_call("cou run -h".split()) == 0
+
+    def test_version(self) -> None:
+        """Test that version is working."""
+        assert check_call("cou --version".split()) == 0
+        assert check_call("cou -v".split()) == 0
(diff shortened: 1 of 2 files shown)


tests/functional/tests/cou_smoke.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+        else:
+            log.warning("Installing python package")
+            assert check_call("python3 -m pip install .".split()) == 0
+            assert (
+                check_output("which cou".split())
+                .decode()
+                .strip()
+                .startswith(os.path.join(os.getcwd(), ".tox"))
+            )
```
rgildein, line 58: I do not think that this is needed, since Python version will be installed by tox.
To me, should have fixcure/helper function to execute cou commands. This function could also handle installation via install function or we can use that install function as you did here (e.g. in setUpClass).

Both of those function should be reusable for all other tests or we can define parent SnapTestCase class.

tests/functional/tests/cou_smoke.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def remove_package() -> None:
+        """Remove cou package."""
+        cou_test_snap = os.environ.get("TEST_SNAP")
+        if cou_test_snap:
+            log.info("Removing snap package cou")
+            check_call("sudo snap remove charmed-openstack-upgrader --purge".split())
+        else:
+            logging.info("Uninstalling python package cou")
+            check_call("python3 -m pip uninstall --yes charmed-openstack-upgrader".split())
```
rgildein, line 68: There is no need to uninstall Python cou, since this env is managed by tox.

tests/functional/tests/cou_smoke.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def test_plan_backup(self) -> None:
+        """Test plan with backup."""
+        result = check_output("cou plan".split()).decode()
```
rgildein, line 72: We should not call `cou` direcly, but rather through helper function. e.g, `cou("plan", "--backup")` or `cou_run("plan")`

tests/functional/tests/tests.yaml
```diff
@@ -1,5 +1,6 @@
 tests:
   - tests.functional.tests.model.COUModelTest
+  - tests.functional.tests.cou_smoke.COUSmokeTest
```
rgildein, line 3: not-blocking: instead of `cou_smoke` we could use `smoke` and instead of `COUSmokeTest` -> `SmokeTest` or better `CLITest`. I used `COUModelTest` only because the name of class is `COUModel`.

PR #182, opened 2023-11-30
Not putting --backup in the mutually exclusive group

agileshaw:
Currently `--backup` option cannot be used with `--quiet` and `--verbose` because they are in the same mutually exclusive group. Moving it out fixes the problem


gabrielcocenza (approved):
LGTM. Thanks

gabrielcocenza (approved):

PR #183, opened 2023-12-01
fix wait for idle on slow network

rgildein:
During my manual testing, I found out two issues:

1. Using 11s for timeout in pre-upgrade step verifying that model is in idle state cause an issue on slow network. It raised an exception, saying that apps are not in idle. This was caused by slower network connection, since I was running it locally and the model was on an external OpenStack cloud. After changing timeout to 60 it works fine.  I also tested that running cou on the external cloud directly do no cause any issue.  Maybe we can set value to 60 (instead of using `DEFAULT_TIMEOUT`), but that can cause unwanted long waiting if any app is not in idle.

```bash
$ cou plan
...
Verify that all OpenStack applications are in idle state ✖
2023-12-01 17:44:45 [ERROR] Timed out waiting for model:
  rabbitmq-server/0 [idle] active: Unit is ready
  neutron-api/0 [idle] active: Unit is ready
  glance/0 [idle] active: Unit is ready
  cinder/0 [idle] active: Unit is ready
  ...  # all apps are in idle
# after change
$ COU_TIMEOUT=60 cou plan
...
Running cloud upgrade...
Verify that all OpenStack applications are in idle state ✔
Backup mysql databases ✔
...
```

2. error message from `wiat_for_idle` is not correct. Example before fix
```bash
2023-12-01 17:09:14 [ERROR] Timed out waiting for model:
  rabbitmq-server/0 [idle] active: Unit is ready
    neutron-api/0 [idle] active: Unit is ready
    glance/0 [idle] active: Unit is ready
    cinder/0 [idle] active: Unit is ready
    ...
#  example after fix
2023-12-01 17:09:14 [ERROR] Timed out waiting for model:
  rabbitmq-server/0 [idle] active: Unit is ready
  neutron-api/0 [idle] active: Unit is ready
  glance/0 [idle] active: Unit is ready
  cinder/0 [idle] active: Unit is ready
```

### DIFF
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -77,7 +77,9 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
         PreUpgradeStep(
             description="Verify that all OpenStack applications are in idle state",
             parallel=False,
-            coro=analysis_result.model.wait_for_idle(timeout=11, idle_period=10),
+            coro=analysis_result.model.wait_for_idle(
+                timeout=11, idle_period=10, raise_on_blocked=True
+            ),
         )
     )
     if backup_database:
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -46,6 +46,7 @@ from cou.exceptions import (
 from cou.steps import PreUpgradeStep, UpgradePlan
 from cou.steps.analyze import Analysis
 from cou.steps.backup import backup
+from cou.utils.juju_utils import DEFAULT_TIMEOUT
 from cou.utils.openstack import LTS_TO_OS_RELEASE, OpenStackRelease
 
 logger = logging.getLogger(__name__)
@@ -78,7 +79,11 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
             description="Verify that all OpenStack applications are in idle state",
             parallel=False,
             coro=analysis_result.model.wait_for_idle(
-                timeout=11, idle_period=10, raise_on_blocked=True
+                # NOTE (rgildein): We need to DEFAULT_TIMEOUT so it's possible to change if
+                #                  a network is too slow, this could cause an issue.
+                timeout=DEFAULT_TIMEOUT + 1,
+                idle_period=10,
+                raise_on_blocked=True,
             ),
         )
     )
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -24,7 +24,7 @@ from juju.application import Application
 from juju.client._definitions import FullStatus
 from juju.client.connector import NoConnectionException
 from juju.client.jujudata import FileJujuData
-from juju.errors import JujuError
+from juju.errors import JujuAppError, JujuError, JujuUnitError
 from juju.model import Model
 from juju.unit import Unit
 from macaroonbakery.httpbakery import BakeryException
@@ -450,6 +450,7 @@ class COUModel:
         timeout: int,
         idle_period: int = DEFAULT_MODEL_IDLE_PERIOD,
         apps: Optional[list[str]] = None,
+        raise_on_blocked: bool = False,
     ) -> None:
         """Wait for applications to reach an idle state.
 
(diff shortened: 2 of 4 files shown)


cou/steps/plan.py
```diff
@@ -77,7 +78,13 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
         PreUpgradeStep(
             description="Verify that all OpenStack applications are in idle state",
             parallel=False,
-            coro=analysis_result.model.wait_for_idle(timeout=11, idle_period=10),
+            coro=analysis_result.model.wait_for_idle(
+                # NOTE (rgildein): We need to DEFAULT_TIMEOUT so it's possible to change if
+                #                  a network is too slow, this could cause an issue.
+                timeout=DEFAULT_TIMEOUT + 1,
```
agileshaw, line 84: Instead of simply setting it to `DEFAULT_TIMEOUT + 1`, can we do `timeout=max(DEFAULT_TIMEOUT + 1, 11)`? This prevents error in case user sets a DEFAULT_TIMEOUT value smaller than 10

cou/utils/juju_utils.py
```diff
@@ -472,14 +476,22 @@ async def _wait_for_idle() -> None:
             # NOTE(rgildein): Defining wrapper so we can use retry with proper timeout
             model = await self._get_model()
             try:
-                await model.wait_for_idle(apps=apps, timeout=timeout, idle_period=idle_period)
-            except asyncio.exceptions.TimeoutError as error:
+                await model.wait_for_idle(
+                    apps=apps,
+                    timeout=timeout,
+                    idle_period=idle_period,
+                    raise_on_blocked=raise_on_blocked,
```
agileshaw, line 483: Perhaps we should also add `raise_on_error`?

agileshaw (changes requested):

cou/steps/plan.py
agileshaw: No actually I meant `max` because if `DEFAULT_TIMEOUT` is `5`, then we will have `timeout = 5+1, idle_period = 10`, which will always result in failure.

cou/utils/juju_utils.py
agileshaw: Good point :+1:

cou/steps/plan.py
```diff
@@ -78,7 +79,11 @@ async def generate_plan(analysis_result: Analysis, backup_database: bool) -> Upg
             description="Verify that all OpenStack applications are in idle state",
             parallel=False,
             coro=analysis_result.model.wait_for_idle(
-                timeout=11, idle_period=10, raise_on_blocked=True
+                # NOTE (rgildein): We need to DEFAULT_TIMEOUT so it's possible to change if
+                #                  a network is too slow, this could cause an issue.
+                timeout=DEFAULT_TIMEOUT + 1,
```
gabrielcocenza, line 84: Add explanation because the + 1 second.

gabrielcocenza (changes requested):

cou/utils/juju_utils.py
```diff
@@ -464,6 +465,9 @@ async def wait_for_idle(
         :type idle_period: int
         :param apps: Applications to wait, defaults to None
         :type apps: Optional[list[str]], optional
+        :param raise_on_blocked: If any unit or app going into "blocked" status immediately raises
```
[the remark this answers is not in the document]
rgildein: There is default value set to False, so I'm not sure what do you mean.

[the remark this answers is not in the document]
gabrielcocenza: Sorry, I use a [docstring extension](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) on vscode that automatically adds the default value on the description and I thought you didn't use a default value.

cou/utils/juju_utils.py
```diff
@@ -464,6 +465,9 @@ async def wait_for_idle(
...
+                                 WaitForApplicationsTimeout.
```
gabrielcocenza, line 469: nit:
```
WaitForApplicationsTimeout, defaults to False.
```

gabrielcocenza (approved):
LGTM. I left a non-blocker suggestion

agileshaw (commented):
Just one suggestion in-line to simply the explanatory comment. Otherwise looks good

Issue #185, opened 2023-12-05
COU should show the log file path if the user does not use quiet mode


gabrielcocenza:
Currently the information about the log file path is show just if user set the verbosity to info. Ex:
```
cou plan
Connected to 'test' ✔
Analyzing cloud... ✔
Generating upgrade plan... ✔
Upgrade cloud from 'ussuri' to 'victoria'
        Verify that all OpenStack applications are in idle state
        Backup mysql databases
        Control Plane principal(s) upgrade plan
                Upgrade plan for 'designate-bind' to victoria
                        Upgrade software packages of 'designate-bind' from the current APT repositories
                        Upgrade 'designate-bind' to the new channel: 'victoria/stable'
                        Wait 300s for app designate-bind to reach the idle state.
                        Check if the workload of 'designate-bind' has been upgraded
```

```
cou plan -vv
2023-12-05 18:44:33 [INFO] Logs of this execution can be found at /home/cocenza/.local/share/cou/log/cou-20231205184433.log
Connecting to 'admin/test' model... \2023-12-05 18:44:34 [INFO] Using model: test
Connected to 'test' ✔
Analyzing cloud... \2023-12-05 18:44:34 [INFO] Analyzing the OpenStack deployment...
2023-12-05 18:44:34 [WARNING] Failed to get origin for designate-bind, no origin config found
2023-12-05 18:44:34 [INFO] Control Plane:
designate-bind:
  model_name: test
  charm: designate-bind
  charm_origin: ch
  os_origin: ''
  channel: ussuri/stable
  units:
    designate-bind/0:
      workload_version: 9.16.1
      os_version: ussuri
Data Plane:

Current minimum OS release in the cloud: ussuri

Current minimum Ubuntu series in the cloud: focal

Analyzing cloud... ✔
Generating upgrade plan... \2023-12-05 18:44:34 [WARNING] Not triggering the workload upgrade of app designate-bind: None already set to cloud:focal-victoria
Generating upgrade plan... ✔
2023-12-05 18:44:34 [INFO] Upgrade cloud from 'ussuri' to 'victoria'
        Verify that all OpenStack applications are in idle state
        Backup mysql databases
        Control Plane principal(s) upgrade plan
                Upgrade plan for 'designate-bind' to victoria
                        Upgrade software packages of 'designate-bind' from the current APT repositories
                        Upgrade 'designate-bind' to the new channel: 'victoria/stable'
[...]

PR #186, opened 2023-12-05
Show log file path if not in quiet mode.

gabrielcocenza:
- print the log file path if not in quiet mode.
- log as info on both situation

Closes: #185


gabrielcocenza:
With those changes this is the output:
```
cou plan
Logs of this execution can be found at /home/cocenza/.local/share/cou/log/cou-20231205185005.log
Connected to 'test' ✔
Analyzing cloud... ✔
Generating upgrade plan... ✔
        Backup mysql databases
                Upgrade plan for 'designate-bind' to victoria
                        Upgrade software packages of 'designate-bind' from the current APT repositories
                        Upgrade 'designate-bind' to the new channel: 'victoria/stable'
                        Wait 300s for app designate-bind to reach the idle state.
                        Check if the workload of 'designate-bind' has been upgraded
```

Using the quiet mode
```
cou plan --quiet
        Backup mysql databases
                Upgrade plan for 'designate-bind' to victoria
                        Upgrade software packages of 'designate-bind' from the current APT repositories
                        Upgrade 'designate-bind' to the new channel: 'victoria/stable'
                        Wait 300s for app designate-bind to reach the idle state.
                        Check if the workload of 'designate-bind' has been upgraded
```

agileshaw (changes requested):
Instead of printing, we can leverage Halo in this case since we already implemented the logic to disable its output in `quiet` mode ([cou/cli.py#L188](https://github.com/canonical/charmed-openstack-upgrader/blob/main/cou/cli.py#L188))

E.g.
```
progress_indicator.start("Configuring logging...")
progress_indicator.stop_and_persist(text=f"Logs of this execution can be found at {file_name}")
```

The code change would be much smaller with this approach.

cou/logging.py
```diff
@@ -76,7 +78,10 @@ def setup_logging(log_level: str = "INFO") -> None:
 
     root_logger.addHandler(log_file_handler)
     root_logger.addHandler(console_handler)
-    logger.info("Logs of this execution can be found at %s", file_name)
+    log_file_msg = f"Logs of this execution can be found at {file_name}"
+    if not quiet:
+        print(log_file_msg)
+    logger.info(log_file_msg)
```
rgildein, line 84: What is purpose to log this message to log file? and If I do `-v` I will see print and info log message at once. I think we should remove the logging message.

rgildein (changes requested):
I agree with @agileshaw that we should use Halo indicator instead of printing message.

cou/logging.py
```diff
@@ -76,7 +76,8 @@ def setup_logging(log_level: str = "INFO") -> None:
 
     root_logger.addHandler(log_file_handler)
     root_logger.addHandler(console_handler)
-    logger.info("Logs of this execution can be found at %s", file_name)
+    progress_indicator.start("Configuring logging...")
```
agileshaw, line 79: This should be put at the beginning of the function. Otherwise we would not see the indication if something goes wrong during logger configuration.

agileshaw (dismissed):
LGTM

cou/logging.py
```diff
@@ -76,7 +78,8 @@ def setup_logging(log_level: str = "INFO") -> None:
 
     root_logger.addHandler(log_file_handler)
     root_logger.addHandler(console_handler)
-    logger.info("Logs of this execution can be found at %s", file_name)
+
+    progress_indicator.stop_and_persist(text=f"Logs of this execution can be found at {file_name}")
```
Pjack, line 82: I don't have the context of cou so I feel wried that why we need to setup progress_indicator in a "setup_logging" function.  Maybe we should decouple it?

cou/logging.py
```diff
@@ -46,6 +46,11 @@ def setup_logging(log_level: str = "INFO") -> None:
     :param log_level: Logging level, defaults to "INFO"
     :type log_level: str, optional
     """
+    progress_indicator.start("Configuring logging...")
+    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
+    file_name = f"{COU_DIR_LOG}/cou-{time_stamp}.log"
+    pathlib.Path(COU_DIR_LOG).mkdir(parents=True, exist_ok=True)
```
rgildein, line 52: not-blocking: `COU_DIR_LOG` is already a Path no need to do it again.

cou/logging.py
rgildein: You will see
```bash
$ cou run
Configuring logging... /  # the / will rotate to show progress
```
after that you will see
```bash
$ cou run
Logs of this execution can be found at <file-name>.log ✔
Connected to 'test-model' ✔
Analyzing cloud... ✔
Generating upgrade plan... ✔
...
```

rgildein (approved):
LGTM, one small not blocking comment

cou/logging.py
Pjack: BTW, Please feel free to resolve the conversation. I just want to ask some questions but not blocking the merge.

rgildein: I think I do not understand, the indicator is not sent to anywhere. Let my try to explain how it's works, maybe it will be helpful.
We are creating indicator objects in utils, with some configuration of our choice.  This indicator object is following singleton pattern, so we can use the same (per each run) indicator everywhere.
I'm glad that you are interested, so happy to provide any answer.

rgildein (approved):

agileshaw (approved):

PR #187, opened 2023-12-06
Smoke test to test help and version commands

gabrielcocenza:
- test the help and version cli commands.

### DIFF
--- a/tests/functional/tests/smoke.py
+++ b/tests/functional/tests/smoke.py
@@ -122,6 +122,21 @@ class SmokeTest(unittest.TestCase):
         """
         return run([self.exc_path] + cmd, capture_output=True, text=True)
 
+    def test_help(self) -> None:
+        """Test that help command is working."""
+        cmds = ["", "plan", "run"]
+        help_options = ["-h", "--help"]
+        for cmd in cmds:
+            for help in help_options:
+                help_cmd = [cmd, help] if cmd else [help]
+                self.assertTrue(self.cou(help_cmd).returncode == 0)
+
+    def test_version(self) -> None:
+        """Test that version command is working."""
+        version_cmds = ["--version", "-V"]
+        for cmd in version_cmds:
+            self.assertTrue(self.cou([cmd]).returncode == 0)
+
     def test_plan_default(self) -> None:
         """Test plan with backup."""
         result = self.cou(["plan", "--model", self.model_name]).stdout
--- a/tests/functional/tests/smoke.py
+++ b/tests/functional/tests/smoke.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import os
 import unittest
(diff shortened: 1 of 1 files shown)


tests/functional/tests/smoke.py
```diff
@@ -122,6 +122,21 @@ def cou(self, cmd: list[str]) -> CompletedProcess:
         """
         return run([self.exc_path] + cmd, capture_output=True, text=True)
 
+    def test_help(self) -> None:
+        """Test that help command is working."""
+        cmds = ["", "plan", "run"]
+        help_options = ["-h", "--help"]
+        for cmd in cmds:
+            for help in help_options:
```
rgildein, line 130: Exactly for this I prefer to use pytest. Now it's too late, but I would be better to use fixtures for installing snap and cou wrapper.

rgildein (dismissed):
LGTM

tests/functional/tests/smoke.py
chanchiwai-ray: Do you mean the for-loop?

rgildein: I was going to suggest that, because I saw it in hw-observer PR. But when I was looking on it at pypi.org I found out that this package is not maintained often (last change is from 10 months ago and previous one is 3 years old). That's why I do not think that it's good to be used, I spoke about it with @dashmage.

gabrielcocenza: It's not like pytest fixture, but using [SubTest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) is a better approach because it will show which sub test failed.

I'll change to something like:
```python
    def test_help(self) -> None:
        """Test that help command is working."""
        help_options = itertools.product(["", "plan", "run"], ["-h", "--help"])
        for cmd, help in help_options:
            help_cmd = [cmd, help] if cmd else [help]
            with self.subTest(help_cmd):
                self.assertTrue(self.cou(help_cmd).returncode == 0)
```

agileshaw (approved):
LGTM. I think SubTest fits the need here (and cleaner than parameterize).

tests/functional/tests/smoke.py
rgildein: I did not know about [SubTest](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests), that's really nice <3. 

> is a better approach because it will show which sub test failed.

If you use fixtures and parametrize from pytest it show also which tests and option failed and code is a little bit nicer with decorator instead of context manager. But I really like the subTest.

gabrielcocenza: I think my sentence was poor. I meant that was a better approach than what I had before and not that is a better approach than pytest or  parametrized that also has this feature of showing which sub test failed.

PR #188, opened 2023-12-06
Smoke test for cou run

gabrielcocenza:
- check that designate-bind can upgrade from ussuri to victoria.


rgildein (approved):
Nice we can run two upgrades. LGTM

agileshaw (approved):
Nice way of checking upgrade success:+1:

Issue #189, opened 2023-12-06
Remove functional tests using the OpenStack bundle


gabrielcocenza:
Currently, COU is using a heavy OpenStack bundle to just test the backup functionality. It's possible to mock the part that checks if there is a relation between keystone and mysql and use a lighter bundle containing just 3 units of mysql-innodb-cluster and 1 unit of designate-bind.

Pros:
- faster testing
- less time running the CI
- No need to have a dedicated github runner to run the project
- Same test coverage with less time and less resources

Cons:
- Miss the plan generation and run of more complex bundles.

The upgrade of more complex bundles probably will happen on SolQA environment (TBD) .

PR #190, opened 2023-12-06
Remove OpenStack bundle and make the backup test in the smoke bundle

gabrielcocenza:
Currently, COU is using a heavy OpenStack bundle to just test the backup functionality. It's possible to mock the part that checks if there is a relation between keystone and mysql and use a lighter bundle containing just 3 units of mysql-innodb-cluster and 1 unit of designate-bind.

The upgrade of more complex bundles probably will happen on SolQA environment (TBD).

Pros:
- No need to have a dedicated github runner to run the project
- Same functional test coverage with less time and less resources.

Cons:
- Miss the plan generation and run of more complex bundles.

Closes: #189


tests/functional/tests/backup.py
```diff
@@ -17,21 +16,27 @@
...
-
-        sync_connect()
-        backup_file = sync_backup(model)
-        logger.info("Backup file: %s", backup_file)
-        assert os.path.getsize(backup_file) > 0
-        self.addCleanup(os.remove, backup_file)
-        self.addCleanup(zaza.clean_up_libjuju_thread)
+        # NOTE(gabrielcocenza) mocking _check_db_relations to return True allows
+        # to test that backup without the need of deploying a heavy OpenStack bundle.
+        with patch("cou.steps.backup._check_db_relations") as mock_relations:
```
rgildein, line 37: I would prefer to run backup with snap, but I see that it would not be possible.

rgildein (approved):
LGTM

tests/functional/tests/smoke.py
```diff
@@ -122,35 +122,43 @@ def cou(self, cmd: list[str]) -> CompletedProcess:
...
         return run([self.exc_path] + cmd, capture_output=True, text=True)
 
-    def test_plan_default(self) -> None:
-        """Test plan with backup."""
-        result = self.cou(["plan", "--model", self.model_name]).stdout
-        expected_plan = (
+    def generate_expected_plan(self, backup: bool = True) -> str:
+        """Generate the expected plan for the smoke bundle.
+
+        :param backup: Weather the plan should contains the backup, defaults to True
```
agileshaw, line 128:
```suggestion
        :param backup: Whether the plan should contain the backup step, defaults to True
```

tests/functional/tests/backup.py
```diff
@@ -17,21 +16,27 @@
...
-        sync_connect = zaza.sync_wrapper(model.connect)
-
-        sync_connect()
-        backup_file = sync_backup(model)
-        logger.info("Backup file: %s", backup_file)
-        assert os.path.getsize(backup_file) > 0
-        self.addCleanup(os.remove, backup_file)
-        self.addCleanup(zaza.clean_up_libjuju_thread)
+        # NOTE(gabrielcocenza) mocking _check_db_relations to return True allows
+        # to test that backup without the need of deploying a heavy OpenStack bundle.
```
agileshaw, line 36:
```suggestion
        # backup test without the need of deploying a heavy OpenStack bundle.
```

agileshaw (changes requested):
Overall looks good. Just a few typo and grammar fixes as suggested inline.

tests/functional/tests/backup.py
```diff
@@ -17,21 +16,27 @@
 class BackupTest(unittest.TestCase):
     """Code for backup test."""
 
+    def setUp(self) -> None:
+        zaza.get_or_create_libjuju_thread()
+        model_name = zaza.model.get_juju_model()
+        self.model = COUModel(model_name)
```
agileshaw, line 22: Don't we need to connect to the model by running `self.model.connect()`?

agileshaw (dismissed):
LGTM, with one non-blocking discussion above

tests/functional/tests/backup.py
rgildein: I do not think there is any need for that since it's calling the same functions as this. But I agree that it could be confusing, that we are not connecting to model.

Issue #173, opened 2023-11-28
Improve UI/UX with CLI restructuring


agileshaw:
The current `plan` sub-command provides a full cloud upgrade plan. This may not actually be as useful as we thought to user because:
1. Running `cou run` also print out the same plan before proceeding. 
2. If the time gap between running `cou plan` and `cou run` is too big, the steps in `cou plan` may no longer accurately reflect the actual cloud state.
3. COU is a stateless tool so it's not possible to apply a previously generated plan

So after discussion, we plan to modify the CLI in following ways:
- drop both plan and run sub-commands
- when running `cou` with no options, the user will get:
    - a printout of the plan
    - a question - do you want to continue? (y/N) - no is the default option
    - detect if run in a script (without a tty) and auto-answer no instead of hanging.
- when running `cou --run`,  the execution starts without prompt after printing out plan (equivalent to answering `y` in the previous mode but without additional interaction)

agileshaw:
After rediscussion, we will keep the current CLI structure (not dropping the `plan` and `run` sub-commands). Instead, we will make the following improvements:
- add a message at the end of the plan to say that the actual process may differ because the plan will be recalculated at upgrade time
- rename `run` to `upgrade`
- `cou upgrade`: add a prompt after the plan to ask the user whether they want to continue
- replace `--no-interactive` with `--auto-approve` and remove `--interactive` (since it is already the default)

PR #196, opened 2023-12-11
Add CLI prompts

agileshaw:
- Move text styling code from steps/execute.py to a newly created utils/text_styler.py to be shared by multiple modules
- Re-structured prompt handling with high versatility though parameters
- Add prompt to warn user about the potential discrepancy between plan and actual upgrade steps
- Add prompt to let the user confirm about proceeding with plan
- Create dedicated function for print and logging debug message
- Accept also full word "yes" and "no" as valid user input

relates-to: #173


cou/cli.py
```diff
@@ -72,6 +74,17 @@ def _missing_(cls, value: object) -> Enum:
         raise ValueError(f"{value} is not a valid member of VerbosityLevel.")
 
 
+def prompt_message(parameter: str) -> str:
```
gabrielcocenza, line 77: I'm a little hesitant of having two function with the same name `prompt_message` here and at `execute.py`. If we can't have a single function to deal with both cases, I think we should name then differently

tests/unit/test_cli.py
```diff
@@ -122,15 +124,72 @@ async def test_run_upgrade_quiet(
...
-    mock_logger.info.assert_called_once_with(plan)
-    mock_apply_step.assert_called_once_with(plan, True)
+    mock_logger.debug.assert_called_once_with(plan)
+    mock_apply_step.assert_called_once_with(plan, False)
     mock_print.call_count == expected_print_count
     mock_manually_upgrade.assert_called_once()
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("input_value", ["n", "x"])
```
gabrielcocenza, line 138: nit: Add N option
```suggestion
@pytest.mark.parametrize("input_value", ["n", "N", "x"])
```

tests/unit/test_cli.py
```diff
@@ -122,15 +124,72 @@ async def test_run_upgrade_quiet(
...
+    mock_input,
+    mock_apply_step,
+    mock_analyze_and_plan,
+    mock_manually_upgrade,
+):
+    plan = UpgradePlan(description="Upgrade cloud from 'ussuri' to 'victoria'")
+    plan.add_step(PreUpgradeStep(description="backup mysql databases", parallel=False))
+    mock_analysis_result = MagicMock()
+    mock_analyze_and_plan.return_value = (mock_analysis_result, plan)
+    mock_input.side_effect = "y"
```
gabrielcocenza, line 181: nit: parameterize and add `Y` option too

gabrielcocenza (commented):
Generally LGTM, but I left some points

cou/cli.py
agileshaw: My initial idea was to have them sharing the same function, but was worried that maybe it is a bit mixed in logic. I will upload a patch with that approach and we can discuss how much we like it.

cou/utils/text_styler.py
```diff
@@ -37,3 +39,35 @@ def normal(text: str) -> str:
     :rtype: str
     """
     return Style.RESET_ALL + text + Style.RESET_ALL
+
+
+def prompt_message(parameter: str, default: Optional[str] = None) -> str:
```
gabrielcocenza, line 44: I think `default_choice` represents better this function argument. `default` is a little vague.

gabrielcocenza (commented):
I definitely prefer this approach. Thanks. I think we can rename one parameter, but LGTM

cou/cli.py
```diff
@@ -149,7 +155,26 @@ async def run_upgrade(
...
+        match prompt_input:
+            case "y":
+                logger.info("Start the upgrade.")
+            case "n":
+                logger.info("Exiting COU without running upgrades.")
+                return
+            case _:
+                print("No valid input provided! Exiting COU without upgrades.")
+                logger.debug("No valid input provided! Exiting COU without upgrades.")
+                return
```
rgildein, line 177: Can we move this to a separate function?

rgildein (changes requested):

cou/utils/text_styler.py
```diff
@@ -0,0 +1,73 @@
...
+    :rtype: str
+    :raise ValueError: raise ValueError if default choice is invalid
+    """
+    continue_option = "y"
+    abort_option = "n"
+
+    if not default_choice:  # use all lowercases if no default is passed
+        pass
+    elif default_choice.casefold() == "y":
+        continue_option = "Y"
```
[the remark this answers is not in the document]
agileshaw: Because when "y" is the default choice, we should display it in upper case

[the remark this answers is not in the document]
rgildein: So why not simply use `default_choice.upper()`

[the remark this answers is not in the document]
agileshaw: Because we need to check whether the `default_choice` is valid, and also 'y' and 'n' are in different position of the message so we need to know which one needs to be in uppercase and which one should remain lower.

cou/utils/text_styler.py
```diff
@@ -0,0 +1,73 @@
...
+    else:
+        raise ValueError(f"Invalid default choice: {default_choice}")
+
+    return (
+        normal("\n" + parameter + "\nContinue (")
+        + bold(continue_option)
+        + normal("/")
+        + bold(abort_option)
+        + normal("): ")
+    )
```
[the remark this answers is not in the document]
agileshaw: For options I will group them together. But we can not do the same for parameter because we will lose "Continue" with this approach.

cou/cli.py
```diff
@@ -149,7 +155,26 @@ async def run_upgrade(
...
+
+    if interactive:
+        prompt_input = (
+            await ainput(
+                prompt_message("Would you like to start the upgrade?", default_choice="n")
+            )
+        ).casefold()
+
+        match prompt_input:
+            case "y":
```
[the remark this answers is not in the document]
agileshaw: I can add `yes` or `no`, I don't think it's necessary to remove the single-letter input.

cou/utils/text_styler.py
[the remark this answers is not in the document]
rgildein: Not at all, you will do `["<current-parameter>", "Continue"]`. But maybe we do not need it and it could be just 
```python
options = normal("/").join(bold(option) for option in options)
return  normal("\n" + parameter + "\nContinue (") + options + normal("): ") 
```
And the `options` can be defined as input parameter with default value, or we just use constant for it.

cou/utils/text_styler.py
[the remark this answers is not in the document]
rgildein: I do not think we need to check if default value is valid, since those values are cumming from source code and not from any config, user inputs, etc.

cou/utils/text_styler.py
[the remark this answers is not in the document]
rgildein: and you can do also
```python
options = normal("/").join(bold(option.upper() if option == default else option) for option in options)
```

cou/cli.py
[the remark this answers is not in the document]
rgildein: OK, I agree.

cou/utils/text_styler.py
rgildein: heh, I do not like the `default_choice` is too long for no purpose. But I see your point, if we add `choices` or `options` input parameter (see my comment below) using only `default` will be even more intuitive.

cou/utils/text_styler.py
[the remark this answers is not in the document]
agileshaw: To me I prefer having an exception here even if it only been used internally. This makes the program more robust.

cou/utils/text_styler.py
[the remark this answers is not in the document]
agileshaw: I see your point now. Sure we can pack the option iteration in a one-liner.

cou/utils/text_styler.py
agileshaw: Since we are just using `Continue?` check, we should always have only `y` and `n` as options, at least in the print message. So I prefer keeping the options a constant in the function instead of making it a parameter

cou/cli.py
```diff
@@ -126,9 +128,13 @@ async def get_upgrade_plan(model_name: Optional[str], backup_database: bool) ->
     :type backup_database: bool
     """
     analysis_result, upgrade_plan = await analyze_and_plan(model_name, backup_database)
-    logger.info(upgrade_plan)
+    logger.debug(upgrade_plan)
     print(upgrade_plan)  # print plan to console even in quiet mode
     manually_upgrade_data_plane(analysis_result)
+    print(
+        "Please note that the actually upgrade steps could be different "
+        "because the plan will be re-calculated at upgrade time."
```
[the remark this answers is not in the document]
agileshaw: I have made the message clearer in the new patch.

cou/utils/__init__.py
```diff
@@ -15,9 +15,23 @@
...
 from pathlib import Path
+from typing import Any
 
 from halo import Halo
 
 COU_DATA = Path(f"/home/{os.getenv('USER')}/.local/share/cou") if os.getenv("USER") else Path(".")
 progress_indicator = Halo(spinner="line", placement="right")
+
+
+def print_and_debug(message: Any, logger: Logger) -> None:
```
rgildein, line 28: not-blocking: We can use `inspect.stack()` to get info about file, which is calling this function, and use it for logger definition. Like this, logger will not need to be passed as argument.

cou/cli.py
```diff
@@ -89,6 +89,28 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
...
 
+async def continue_upgrade() -> bool:
+    """Determine whether to continue with the upgrade based on user input.
+
+    :return: Boolean value indicate whether to continue with the upgrade.
+    :rtype: bool
+    """
+    prompt_input = (
+        await ainput(prompt_message("Would you like to start the upgrade?", default_choice="n"))
+    ).casefold()
```
rgildein, line 100: Now I realized, what is the purpose of defualt_choise if we do not use it?
```suggestion
    default_choice = "n"
    prompt_input = await ainput(prompt_message("Would you like to start the upgrade?", default_choice="n"))
    prompt_input = (prompt_input or default_choice).casefold()
```

cou/utils/text_styler.py
```diff
@@ -41,33 +39,24 @@ def normal(text: str) -> str:
...
-    if not default:  # use all lowercases if no default is passed
-        pass
-    elif default.casefold() == "y":
-        continue_option = "Y"
-    elif default.casefold() == "n":
-        abort_option = "N"
-    else:
-        raise ValueError(f"Invalid default choice: {default}")
+    if default_choice and default_choice.casefold() not in choices:
+        raise ValueError(f"Invalid default choice: {default_choice}")
```
rgildein, line 56: not-blocking: I really do not see any benefit using this, if you provide a wrong default value, it will be simply ignored.

rgildein (changes requested):
Default value is not used.

cou/cli.py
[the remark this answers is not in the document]
rgildein: Thanks

cou/utils/text_styler.py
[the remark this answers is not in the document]
rgildein: This make program being same, but more robust to developer error. We spent some time to discussion it and yet do not see that the default value is not used if user pressed only enter.
IMO instead of validating potential developer error, we can design it to have single function handling the input and provide the output.

cou/utils/text_styler.py
rgildein: That's why at the beginning I suggest to make `parameters` input argument instead, so we could use it without continue. However, right now we do not have use case for this, but I think it will come, and then we refactor all this functions again.

tests/unit/test_cli.py
```diff
@@ -122,15 +124,74 @@ async def test_run_upgrade_quiet(
...
-    mock_logger.info.assert_called_once_with(plan)
-    mock_apply_step.assert_called_once_with(plan, True)
+    mock_logger.debug.assert_called_once_with(plan)
+    mock_apply_step.assert_called_once_with(plan, False)
     mock_print.call_count == expected_print_count
     mock_manually_upgrade.assert_called_once()
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("input_value", ["n", "N", "x"])
```
gabrielcocenza: add the new option `no`

tests/unit/test_cli.py
```diff
@@ -122,15 +124,74 @@ async def test_run_upgrade_quiet(
...
+    mock_analyze_and_plan.assert_awaited_once_with(None, True)
+    mock_input.assert_has_awaits(
+        [call(cli.prompt_message("Would you like to start the upgrade?", default_choice="n"))]
+    )
+    mock_apply_step.assert_not_awaited()
+    mock_manually_upgrade.assert_not_called()
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("input_value", ["y", "Y"])
```
gabrielcocenza: add the new option `yes`

cou/utils/__init__.py
```diff
@@ -14,10 +14,61 @@
...
+    :param message: The object to print and log.
+    :type message: Any
+    """
+    print(message)
+
+    logger = logging.getLogger(inspect.stack()[1].filename)
+    logger.debug(message)
+
+
+# pylint: disable=dangerous-default-value
```
rgildein, line 44: You should not use dangerous default valu, so I will suggest using None instead of list of "y" and "n".

rgildein (changes requested):
One last issue, suggested also by pylint.

rgildein (approved):
LGTM thanks

gabrielcocenza (approved):
LGTM

Issue #199, opened 2023-12-12
Verification of the model using wait_for_idle should consider also if application is active


gabrielcocenza:
I was testing cou on the following bundle and as it is possible to see, one unit of mysql was in maintenance:

```
App                   Version  Status       Scale  Charm                 Channel        Rev  Exposed  Message
designate-bind        9.16.1   active           1  designate-bind        ussuri/stable   98  no       Unit is ready
mysql-innodb-cluster  8.0.35   maintenance      3  mysql-innodb-cluster  8.0/stable     107  no       Paused. Use 'resume' action to resume normal service.

Unit                     Workload     Agent  Machine  Public address  Ports  Message
designate-bind/0*        active       idle   0        10.101.82.190          Unit is ready
mysql-innodb-cluster/0*  active       idle   1        10.101.82.32           Unit is ready: Mode: R/W, Cluster is NOT tolerant to any failures. 1 member is not active.
mysql-innodb-cluster/1   active       idle   2        10.101.82.151          Unit is ready: Mode: R/O, Cluster is NOT tolerant to any failures. 1 member is not active.
mysql-innodb-cluster/2   maintenance  idle   3        10.101.82.241          Paused. Use 'resume' action to resume normal service.
```
This initial check should fail in such situation and also after the upgrade if some unit is still on maintenance.

PR #201, opened 2023-12-13
Check if applications status are active on wait_for_idle.

gabrielcocenza:
- Without status check, if a unit is on maintenance it will proceed to upgrade
  or pass the post-upgrade check after upgrading.
 - change function name from wait_for_idle to wait_for_active_and_idle to reflect
  the changes.

Closes: #199

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -697,5 +697,5 @@ class OpenStackApplication:
         return PostUpgradeStep(
             description=description,
             parallel=False,
-            coro=self.model.wait_for_idle(self.wait_timeout, apps=apps),
+            coro=self.model.wait_for_active_and_idle(self.wait_timeout, apps=apps),
         )
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -697,5 +697,5 @@ class OpenStackApplication:
         return PostUpgradeStep(
             description=description,
             parallel=False,
-            coro=self.model.wait_for_active_and_idle(self.wait_timeout, apps=apps),
+            coro=self.model.wait_for_active_idle(self.wait_timeout, apps=apps),
         )
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -445,14 +445,14 @@ class COUModel:
             switch=switch,
         )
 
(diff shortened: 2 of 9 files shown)


agileshaw (changes requested):
As discussed in the sync, it would be better changing the function name to `wait_for_active`.

Additionally, please also update PR description (status check is true for all cases, not only in pre-upgrade) and linked issue title (still mentioning `wait_for_active`)

cou/apps/base.py
```diff
@@ -697,5 +697,5 @@ def _get_wait_step(self) -> PostUpgradeStep:
         return PostUpgradeStep(
             description=description,
             parallel=False,
-            coro=self.model.wait_for_idle(self.wait_timeout, apps=apps),
+            coro=self.model.wait_for_active_and_idle(self.wait_timeout, apps=apps),
```
rgildein, line 700: IMO shorter name is better and `wait_for_active` should be enough, since `active` means that it's idle. Or we can do at least `wait_for_active_idle`. What do you say?

cou/utils/juju_utils.py
```diff
@@ -445,14 +445,14 @@ async def upgrade_charm(
...
-    async def wait_for_idle(
+    async def wait_for_active_and_idle(
         self,
         timeout: int,
         idle_period: int = DEFAULT_MODEL_IDLE_PERIOD,
         apps: Optional[list[str]] = None,
         raise_on_blocked: bool = False,
     ) -> None:
-        """Wait for applications to reach an idle state.
+        """Wait for application(s) to reach active and idle state.
```
rgildein, line 455:
```suggestion
        """Wait for application(s) to reach active idle state.
```

rgildein (commented):
LGTM, only name of functions, is unnecessary too long.

cou/apps/base.py
gabrielcocenza: I always prefer been explicit so, `wait_for_active_idle` sounds better to me. I'm going to change it

agileshaw (approved):
LGTM

rgildein (approved):
LGTM, thanks

PR #204, opened 2023-12-15
Mark plan generation as failed when reaching highest supporting release

agileshaw:
If the cloud version is detected as the highest supporting release, mark plan generation as failed then print the message.


agileshaw:
Closing this PR in favor of #205

PR #205, opened 2023-12-15
fix finishing indicator if HighestReleaseAchieved

rgildein:
Without indicator finished, the message will break previous indicator.
- add success for progess_indicator after `HighestReleaseAchieved`
- improve message for `HighestReleaseAchieved`
- improve message for logging indicator

original stdout
```bash
$ cou upgrade --model openstack
Logs of this execution can be found at /home/rgildein/.local/share/cou/log/cou-20231215205252.log  
Connected to 'openstack' ✔
Analyzing cloud... ✔
Generating upgrade plan... \The cloud is already at the latest OpenStack release 'yoga' compatible with series 'focal', and COU does not support series upgrade. Please manually upgrade series and run COU again. 
```

new stdout
```bash
$ cou upgrade --model openstack
Full execution log: '/home/rgildein/.local/share/cou/log/cou-20231215205143.log'  
Connected to 'openstack' ✔
Analyzing cloud... ✔
Generating upgrade plan... ✔
No upgrades available for OpenStack Yoga on Ubuntu Focal.
Newer OpenStack releases may be available after upgrading to a later Ubuntu series.
```

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -82,9 +82,7 @@ def get_log_level(quiet: bool = False, verbosity: int = 0) -> str:
     :return: Log level.
     :rtype: str
     """
-    if quiet:
-        return "CRITICAL"
-    return VerbosityLevel(verbosity).name
+    return "CRITICAL" if quiet else VerbosityLevel(verbosity).name
 
 
 async def continue_upgrade() -> bool:
@@ -220,8 +218,8 @@ def entrypoint() -> None:
         loop = asyncio.get_event_loop()
         loop.run_until_complete(_run_command(args))
     except HighestReleaseAchieved as exc:
+        progress_indicator.succeed()
         print(exc)
-        sys.exit(0)
     except TimeoutException:
         progress_indicator.fail()
         print("The connection was lost. Check your connection or increase the timeout.")
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -287,7 +287,7 @@ def test_entrypoint_highest_release(mock_run_command, mock_indicator):
 
     cli.entrypoint()
 
-    mock_indicator.stop.assert_called_once_with()
+    mock_indicator.succeed.assert_called_once_with()
     mock_indicator.stop.assert_called_once_with()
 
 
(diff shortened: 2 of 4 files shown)


cou/cli.py
```diff
@@ -220,8 +218,8 @@ def entrypoint() -> None:
         loop = asyncio.get_event_loop()
         loop.run_until_complete(_run_command(args))
     except HighestReleaseAchieved as exc:
+        progress_indicator.succeed()
```
agileshaw, line 221: As discussed elsewhere, it would be better to use `stop_and_persist` here.

cou/cli.py
```diff
@@ -220,8 +218,8 @@ def entrypoint() -> None:
...
         print(exc)
-        sys.exit(0)
```
agileshaw, line 224: I think having the `exit(0)` is clearer, but I'm not against removing it. However, you would need to delete the corresponding unit test `tests/unit/test_cli.py::test_entrypoint_highest_release`

agileshaw (changes requested):

cou/cli.py
rgildein: It's default behavior, so it's not important at all.

agileshaw: With your change the previous test implementation would fail. But I see you updated it, so that's good.

tests/unit/test_cli.py
```diff
@@ -285,10 +285,10 @@ def test_entrypoint_highest_release(mock_run_command, mock_indicator):
     """Test TimeoutException exception during entrypoint execution."""
     mock_run_command.side_effect = HighestReleaseAchieved
 
-    with pytest.raises(SystemExit, match="0"):
-        cli.entrypoint()
+    cli.entrypoint()
 
     mock_indicator.stop.assert_called_once_with()
+    mock_indicator.stop.assert_called_once_with()
```
agileshaw, line 291: Why duplicating this line?

agileshaw (approved):

agileshaw (changes requested):
Previous approval was given accidentally.

agileshaw (approved):
LGTM

gabrielcocenza (approved):
LGTM. Thanks

PR #213, opened 2024-01-11
add first draft of pseudo code for data plane upgrade

rgildein:
The idea behind this approach is to handle all the machines and applications in them through Nova-compute applications. That way, not much code will change, what is benefit.

I prepared a diagram for a more detailed description.
cou data-plane v1

Known issues:
 - Nova-compute application needs to be responsible for upgrading apps deployed under the same machine.
   - Steps like config changes must be done as pre-checks in nova-compute for all apps.

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -59,6 +59,16 @@ class ApplicationUnit:
 
 
 @dataclass
+class ApplicationMachine:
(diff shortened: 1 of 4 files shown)


rgildein:
After review, we decide to go another direction.

PR #215, opened 2024-01-11
Sane checks for cli inputs

gabrielcocenza:
- introduce machine class
- sane checks for machine, hostname and azs filtration


rgildein:
These changes were already addressed in #222.

PR #217, opened 2024-01-15
drop smoke tests

rgildein:
After we decided to drop small OpenStack bundle and run all tests again the smoke tests, there is no need to have smoke tests definition in tox or Makefile.


rgildein:
> Looks good. If we are dropping these smoke targets, I think we can also remove `smoke_bundle` in `tests.yaml`?

We could :+1:

agileshaw (approved):

gabrielcocenza (approved):

PR #219, opened 2024-01-16
add UnitUpgradeStep, improve add_step

rgildein:
Add UnitUpgradeStep and is_unit_upgrade_step, also use sub_steps as property to prevent defining it as not list of BaseStep's. Raising TypeError in add_step if object is not based on BaseStep.


cou/steps/__init__.py
```diff
@@ -286,9 +309,24 @@ class UpgradeStep(BaseStep):
...
 
 
 class PostUpgradeStep(UpgradeStep):
     """Represents the post-upgrade step."""
+
+
+def is_unit_upgrade_step(step: BaseStep) -> bool:
+    """Check if it's unit upgrade step.
+
+    :param step: step
```
gabrielcocenza, line 327:
```suggestion
    :param step: Upgrade step
```

gabrielcocenza (commented):
Generally LGTM. I have some small comments that are non-blockers

It's not  totally clear to me the impact of those changes in the big picture, but I can see that the introduction of unit upgrade step might be necessary.

cou/steps/__init__.py
```diff
@@ -286,9 +309,24 @@ class UpgradeStep(BaseStep):
...
+    :type step: BaseStep
+    :return: True or False if it's BaseStep
```
agileshaw, line 329:
```suggestion
    :return: True or False if it's UnitUpgradeStep
```

cou/steps/__init__.py
```diff
@@ -286,9 +309,24 @@ class UpgradeStep(BaseStep):
     prompt: bool = False
 
 
+class UnitUpgradeStep(UpgradeStep):
+    """Represents the upgrade step for individual unit."""
```
agileshaw, line 313:
```suggestion
    """Represents the upgrade step for an individual unit."""
```

cou/steps/__init__.py
```diff
@@ -195,13 +195,36 @@ def done(self) -> bool:
...
+        """Return list of sub-steps.
+
+        :return: List of BaseStep.
+        :rtype: List[BaseStep]
+        """
+        return self._sub_steps
+
+    @sub_steps.setter
+    def sub_steps(self, steps: Iterable[BaseStep]) -> None:
+        """Set list of sub-steps.
```
agileshaw, line 209:
```suggestion
        """Set a list of sub-steps.
```

cou/steps/__init__.py
agileshaw: I don't think it's a good idea making this change because `Upgrade step` has a special meaning in our codebase (`UpgradeStep`).

agileshaw (changes requested):
Some docstring changes suggested in line. Otherwise looks good.

agileshaw (approved):

gabrielcocenza (approved):

PR #220, opened 2024-01-16
Update breaking changes from starter-pack and improve workflow

agileshaw:
- There are recent breaking changes in canonical/sphinx-docs-starter-pack that causes our workflow to fail. Align those changes in COU to fix the issue.
- Update the workflow to only trigger documentation checks if a PR modifies `docs/`. This will prevent un-necessary runs. I'm keeping the on.push as is because I think it is good to have a way to notify us about dependency breakage (those issues are not affecting live doc site).


docs/Makefile
```diff
@@ -93,9 +93,9 @@ woke: woke-install
...
 
 pa11y: pa11y-install html
-	find $(BUILDDIR) -name *.html -exec $(PA11Y) {} \;
+	find $(BUILDDIR) -name *.html -print0 | xargs -n 1 -0 $(PA11Y)
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	. $(VENV); $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	. $(VENV); $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
gabrielcocenza, line 101: nit: add end line

gabrielcocenza (commented):
Generally, LGTM. I'm wondering if it would be good or possible to have in our Makefile a option to sync with the upstream documentation as it happens for example on [charm-helpers](https://github.com/openstack/charm-nova-compute/blob/650f3a5d511690ec27648b30f3b24532378a33a1/Makefile#L15-L17) on classic charms.

agileshaw:
@gabrielcocenza Right now there's no clean way to automate the upstream tracking because we are modifying the files from starter-pack. Documentation team is aware of the limitation and will work on supporting such an automation. But until then, there's nothing we can do from our side.

gabrielcocenza (approved):

PR #221, opened 2024-01-16
Custom COU Namespace

gabrielcocenza:
- create a new NamedTuple class to be the COU Namespace. With that, type is still checked and few arguments are necessary to pass to the functions that use the arguments passed by the user. Moreover, docstrings are shorter and easier to maintain

- Global variables for "data-plane" and "control-plane".


cou/commands.py
```diff
@@ -316,13 +319,32 @@ def create_subparsers(parser: argparse.ArgumentParser) -> argparse._SubParsersAc
     return subparsers
 
 
-def parse_args(args: Any) -> argparse.Namespace:  # pylint: disable=inconsistent-return-statements
+class Namespace(NamedTuple):
```
rgildein, line 322: The name is strange, I will rename it to something like  `CLIargs`.

cou/cli.py
```diff
@@ -131,21 +125,19 @@ async def analyze_and_plan(
     progress_indicator.succeed()
 
     progress_indicator.start("Generating upgrade plan...")
-    upgrade_plan = await generate_plan(analysis_result, backup_database)
+    upgrade_plan = await generate_plan(analysis_result, args)
```
rgildein, line 128: One of disadvantage of this is, that I do not know what arguments are really needed by function and that we are passing more that it's needed (e.g. `args.verbosity`, `args.command`, ...)

cou/cli.py
```diff
@@ -154,24 +146,14 @@ async def get_upgrade_plan(model_name: Optional[str], backup_database: bool) ->
...
-    :type backup_database: bool
-    :param prompt: Whether to prompt to run upgrade interactively.
-    :type prompt: bool
-    :param quiet: Whether to run upgrade in quiet mode.
-    :type quiet: bool
+    :param args: CLI arguments
+    :type args: Namespace
     """
-    analysis_result, upgrade_plan = await analyze_and_plan(model_name, backup_database)
+    prompt = not args.auto_approve
```
rgildein, line 155: Can we do this in the `Namespace`? So here I can do simply `args.prompt`. If yes, I think that the `Namespace` could be renamed to something even different from `CLIargs` (from my comment below).

cou/commands.py
```diff
@@ -316,13 +319,32 @@ def create_subparsers(parser: argparse.ArgumentParser) -> argparse._SubParsersAc
...
+    """Mock Namespace used for type hinting purposes.
```
rgildein, line 323: We are not mocking, but wrapping. Can you change that, thanks?

rgildein (changes requested):
I will suggest renaming `Namespace` to something else and maybe add `prompt` as property or additional argument set during init process.
I see some benefits passing CLI args to everywhere, so I will give +1.

cou/commands.py
rgildein: or `COUargs`

rgildein: Now I realized that this also could be defined as frozen dataclass, lake that it will not be possible to change any arguments in any step accidentally.

cou/cli.py
gabrielcocenza: Yes, but at the same time I prefer to know that the function is using something that comes from the CLI arguments instead of `backup_database`. For someone that never saw the code, looking just at this argument it's almost impossible to figure out that came from the user's input

rgildein: I'll not say impossible, but anyway mixing logic like this in most case is more harmful than useful.

agileshaw: I agree with Robert on this. When going through the changes, I find it strange that the entirety of `CLIargs` is passed to `generate_plan`, in which only one value is actually needed. I think `backup_database` is clearer, and people can simply go back to the calling function to see where the value come from.

This is also true for [`get_log_level`](https://github.com/canonical/charmed-openstack-upgrader/blob/main/cou/cli.py#L75) and [`COUModel()`](https://github.com/canonical/charmed-openstack-upgrader/blob/main/cou/cli.py#L122), which I think it's good that you kept unchanged.

agileshaw (commented):
Overall looks good. Just one comment inline.

cou/cli.py
rgildein: Since this will never be used as Python package it's fine, but if I'm wrong and some will use it as Python package how that will look like? IMO it will be really strange to use, for example generating plan, and you need to provide full `CLIargs` object, where there will be parts which are not used at all.


I do not want to block this, so I'll resolve this conversation.

cou/commands.py
```diff
@@ -316,13 +320,44 @@ def create_subparsers(parser: argparse.ArgumentParser) -> argparse._SubParsersAc
...
+    subcommand: Optional[str] = None  # for help option
+    machines: Optional[list[str]] = None
+    hostnames: Optional[list[str]] = None
+    availability_zones: Optional[list[str]] = None
+
+    @property
+    def prompt(self) -> bool:
+        """Whether if COU should prompt to the user.
+
+        :return: Prompt if true, otherwise don't prompt.
```
rgildein, line 348:
```suggestion
        :return: Prompt if auto_approve is True, otherwise don't prompt.
```

rgildein (approved):
I approve, but I still don't see it as a big benefit.

cou/cli.py
gabrielcocenza: @agileshaw with the next changes we would need to pass more than just `backup_database`. [ pre_plan_sanity_checks](https://github.com/canonical/charmed-openstack-upgrader/pull/222/files#diff-abbb76655f6d778c97d8958cb0158d4fcc2f87854044f43695bcc788b76489e5R211) will need to have `upgrade_groups` and also `machines`, `hosrnames`, `avaiabilty-zones` and etc. The number of inputs to generate plan will start to be quiet big and against the clean code principles.

@rgildein as far as I know, just the map between workload and OpenStack release might be a python package in the future.

rgildein: I'm sorry, but what is "against the clean code principles"? I'm not aware of any official guideline like this in any PEP.

agileshaw (approved):
I see the benefit of having less arguments passed into functions, but I think we are hyper-correcting it a bit with this PR and I have concerns about what this extra layer of abstraction may bring. But this is not a blocker, so I'm also approving it.

PR #222, opened 2024-01-16
Add data-plane pre upgrade sanity checks

gabrielcocenza:
- check if control-plane is already upgraded if user pass "data-pane" upgrade group
- check if data-plane apps exists if user pass "data-pane" upgrade group
- splitted the logic on determine_upgrade_target to several functions

This PR should be merged after #221


cou/steps/plan.py
```diff
@@ -52,24 +54,162 @@
...
+
+    :param analysis_result: Analysis result.
+    :type analysis_result: Analysis
+    :raises OutOfSupportRange: When series is not supported.
+    """
+    supporting_lts_series = ", ".join(LTS_TO_OS_RELEASE)
+    # series already checked at is_valid_openstack_cloud
+    if (
+        current_series := analysis_result.current_cloud_series
+    ) and current_series not in LTS_TO_OS_RELEASE:
```
rgildein, line 105: This should work fine and it is more readable. 
```suggestion
    if (
        current_series := analysis_result.current_cloud_series
        and current_series not in LTS_TO_OS_RELEASE
    ):
```

cou/steps/plan.py
```diff
@@ -52,24 +54,162 @@
...
+    :raises OutOfSupportRange: When the upgrade scope is not supported
+        by the current series.
+    :return: The target OS release to upgrade the cloud to.
+    :rtype: OpenStackRelease
+    """
+    if (
+        (current_os_release := analysis_result.current_cloud_os_release)
+        and (current_series := analysis_result.current_cloud_series)
+        and not (target := current_os_release.next_release)
+    ):
```
rgildein, line 180: Since this is used also outside of this if + for return we should define it before it. 
```suggestion
    target = current_os_release.next_release
    if (
        (current_os_release := analysis_result.current_cloud_os_release)
        and (current_series := analysis_result.current_cloud_series)
        and not target
    ):
```

cou/steps/plan.py
```diff
@@ -52,24 +54,162 @@
...
+    if (
+        current_series
+        and (supporting_os_release := ", ".join(LTS_TO_OS_RELEASE[current_series]))
+        and str(current_os_release) not in supporting_os_release
+        or str(target) not in supporting_os_release
+    ):
+        raise OutOfSupportRange(
+            f"Unable to upgrade cloud from `{current_series}` to '{target}'. "
+            "Both the from and to releases need to be supported by the current "
+            "Ubuntu series '{current_series}': {supporting_os_release}."
```
rgildein, line 195:
```suggestion
            f"Ubuntu series '{current_series}': {supporting_os_release}."
```

tests/unit/steps/test_steps_plan.py
```diff
@@ -216,15 +222,107 @@ def test_determine_upgrade_target_no_upgrade_available():
...
+@pytest.mark.parametrize(
+    "current_os_release, current_series",
+    [
+        (OpenStackRelease("yoga"), "jammy"),
+        (OpenStackRelease("train"), "bionic"),
+    ],
+)
+def test_pre_plan_sane_checks_is_supported_series(current_os_release, current_series, cli_args):
+    mock_analysis_result = MagicMock(spec=Analysis)()
+    with pytest.raises(OutOfSupportRange):
```
rgildein, line 244: We should try to match also error message in any pytest.raises (here and in other tests), which is checking our own error with custom message.

rgildein (changes requested):

cou/steps/plan.py
gabrielcocenza: I've tried before and It does not work.[0](https://peps.python.org/pep-0572/#exceptional-cases)

cou/steps/plan.py
gabrielcocenza: type checker will complain that `current_os_release` can be None and won't have `next_release`. It's a question of choosing add a `ignore` because is already checked on other function or leave this way that is more bullet proof.

cou/steps/plan.py
rgildein: Ok, but this should:
```python
if (
        (current_series := analysis_result.current_cloud_series)
        and current_series not in LTS_TO_OS_RELEASE
):
```
However, IMO this does not bring any benefit than using:
```python
current_series = analysis_result.current_cloud_series
if current_series not in LTS_TO_OS_RELEASE:
```
Since, we will define that variable in any case, and it will not happen only in some cases.

cou/steps/plan.py
```diff
@@ -52,24 +54,162 @@
...
+
+    :param analysis_result: Analysis result.
+    :type analysis_result: Analysis
+    :raises HighestReleaseAchieved: When the OpenStack release is the last supported by the series.
+    """
+    if (
+        (current_os_release := analysis_result.current_cloud_os_release)
+        and (current_series := analysis_result.current_cloud_series)
+        and str(current_os_release) == LTS_TO_OS_RELEASE[current_series][-1]
+    ):
```
rgildein, line 123: Same here, what is the benefit doing it like this? Why not simply
```suggestion
    if str(analysis_result.current_cloud_os_release) == LTS_TO_OS_RELEASE[analysis_result.current_cloud_series][-1]:
```
or maybe even better would be:
```python
last_supported = LTS_TO_OS_RELEASE.get(analysis_result.current_cloud_series, [])[-1]
if str(analysis_result.current_cloud_os_release) == last_supported:
```
Like this, we show that we do not care if `current_cloud_series` exists in `LTS_TO_OS_RELEASE`

cou/steps/plan.py
rgildein: I do not understand how this is more "bullet proof"? I think it is the same approach, but yours is less readable.
The whole idea behind assignment expression is to use it with function e.g. `if target := foo():` so the output is catch. However you are using it for properties and I do not see any benefit of it, it's only less redable.

rgildein (changes requested):

cou/steps/plan.py
gabrielcocenza: mypy is a nightmare when we have `Optional[str]`. This code results in:
`error: Argument 1 to "get" of "dict" has incompatible type "str | None"; expected "str"  [arg-type]` that is why sometimes I prefer to use empty string instead of `None`.

We can disable the [type-checker](https://github.com/canonical/charmed-openstack-upgrader/pull/215/files#diff-abbb76655f6d778c97d8958cb0158d4fcc2f87854044f43695bcc788b76489e5R112) because `current_os_release` and `current_series` are already checked in another sanity check, however you seem to not like the approach.

cou/steps/plan.py
gabrielcocenza: Your solution don't pass mypy. if we define `target` before we need to check that `current_os_release` is not None.

If the walrus operator makes less readable we can define `current_os_release`, `current_series` above and check target.

rgildein: I know, but mypy is just not able to check in assignment expression, you are not checking that is not none. So finding a way around mypy is not "better" solution. If you really want to fix that issue mypy is pointing out, you should check that it is not None also here.

cou/steps/plan.py
rgildein: No, I do not like much the ignores, because when we are using mypy we rely on it, and when we disable it randomly it can cause an issue in the future. 

```python
series = analysis_result.current_cloud_series or ""
last_supported = LTS_TO_OS_RELEASE.get(series, [])[-1]
```
Or maybe we just need helper function to actually get `last_supported_series`?

cou/steps/plan.py
```diff
@@ -53,6 +54,150 @@
...
+    current_series = analysis_result.current_cloud_series
+    if current_series not in LTS_TO_OS_RELEASE:
+        raise OutOfSupportRange(
+            f"Cloud series '{current_series}' is not a Ubuntu LTS series supported by COU. "
+            f"The supporting series are: {supporting_lts_series}"
+        )
+
+
+def is_highest_release_achieved(analysis_result: Analysis) -> None:
+    """Check if the highest OpenStack release is reached by the ubuntu series.
```
agileshaw, line 111:
```suggestion
    """Check if the highest OpenStack release is reached for the current ubuntu series.
```

cou/steps/plan.py
```diff
@@ -53,6 +54,150 @@
...
+    if not analysis_result.min_os_version_data_plane:
+        raise DataPlaneCannotUpgrade(
+            "Cannot find data-plane apps. Is this a valid OpenStack cloud?"
+        )
+    if not is_control_plane_upgraded(analysis_result):
+        raise DataPlaneCannotUpgrade("Please, upgrade control-plane before data-plane")
+
+
+def is_control_plane_upgraded(analysis_result: Analysis) -> bool:
+    """Check if control plane is already upgraded.
```
agileshaw, line 149:
```suggestion
    """Check if control plane has been fully upgraded.
```

cou/steps/plan.py
```diff
@@ -53,6 +54,150 @@
...
+
+    Control-plane will be considered as upgraded when the OpenStack version of it
+    is bigger than the data-plane.
```
agileshaw, line 152:
```suggestion
    is higher than the data-plane.
```

cou/steps/plan.py
agileshaw: I don't see how the change @rgildein suggested would work, calling `current_os_release` before the variable is defined.

For the mypy error, would passing in `current_os_release` as parameter of the function with type check solves the problem? That is:
 ```python
def determine_upgrade_target(current_os_release: OpenStackRelease, current_series: str):
```

**Review by agileshaw — CHANGES_REQUESTED** [review]

File: cou/steps/plan.py (hunk 2 above)
Comment (gabrielcocenza) [reply cou/steps/plan.py:180]: >  but mypy is just not able to check in assignment expression,  you are not checking that is not none

Where did you see this? It looks like they support since [2019](https://github.com/python/mypy/pull/6899). It's also on the changelog from their [blog](https://mypy-lang.blogspot.com/2019/09/mypy-730-released.html)

I don't see the problem. checking a variable like `if my_var:` is enough to be sure that it's not None and continue.

@agileshaw I don't see the benefit of having a function like this. probably we will use like this:
 ```python
 determine_upgrade_target(analysis_result.current_cloud_os_release, analysis_result.current_cloud_series)
 ```
 So the asignament of the function needs to be:
 ```python
 def determine_upgrade_target(current_os_release: Optional[OpenStackRelease], current_series: Optional[str]):
 ```

Comment (agileshaw) [reply cou/steps/plan.py:177]: I see. To me this is not a blocker, but if you choose to use `# type: ignore` at the end of the function, it would be better to add a comment explaining we have already checked the type before calling this function.

**Review by agileshaw — APPROVED** [review]
LGTM, but with one small non-blocking suggestion

File: cou/steps/plan.py (hunk 2 above)
Comment (rgildein) [reply cou/steps/plan.py:177]: @agileshaw You are right me suggestion will not work. 
I saw that it should support it, but tell me what is the difference between 
```python
    if (
        (current_os_release := analysis_result.current_cloud_os_release)
        and (current_series := analysis_result.current_cloud_series)
        and not (target := current_os_release.next_release)
    ):
```
and
```python
    current_os_release = analysis_result.current_cloud_os_release
    current_series = analysis_result.current_cloud_series
    target = current_os_release.next_release
    
    if not target:
        ...
```
Did you check that parameter is None somewhere? IMO if mypy is not raising any error, is due that it could not check properly types of those porperties.

rgildein: I missed that walrus operator is doing also checking, so it means:
```python
if current_os_release and current_series and not target:
    ...
```
What is fine I guess, since these variables will not be empty string, list, etc., or do they?

rgildein (approved):

cou/steps/plan.py
```diff
@@ -53,6 +54,147 @@
 logger = logging.getLogger(__name__)
 
 
+def pre_plan_sanity_checks(args: CLIargs, analysis_result: Analysis) -> None:
+    """Pre checks to generate the upgrade plan.
```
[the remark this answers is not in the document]
gabrielcocenza: I see little benefit on doing this. In `Java` for example, the checked exception forced the signature of every method would list all of the exceptions that it could pass to its caller and has a high price. If we start getting deep into this we could have a situation where we would need to go three or more levels checking all the exceptions that can happen.

On theory a change at a low level of the software would imply on updating on the higher level and I don't think it worth.

cou/steps/plan.py
gabrielcocenza: @agileshaw Samuel did a good comment about changing the name of the functions and I  ended up taking your approach. This way we don't need to use the `# type: ignore` at the end of the function and if the current structure we need to check twice, maybe it's not a bad deal checking once in the `determine_upgrade_target`. We will need to add another checks, so the structure is ready for checking azs, machines, hostnames and etc.

cou/steps/plan.py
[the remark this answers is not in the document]
rgildein: @samuelengineer that's good point, I lost myself with warlock operators in this PR. 
@gabrielcocenza I see your point, however all those `verify_*` function are simple helper functions, and we expected this function to raise those exceptions. If I look at the function name `pre_plan_sanity_checks`, and then the description, I could not tell what this function actually do? It's not returning anything and not mention raising anything.

Please stop referring to that book (I'm going to buy and read it, because there could be good point). I believe that you're misinterpreting things from it, or it has written not for Python. Thanks

rgildein (approved):

agileshaw (approved):
LGTM

PR #223, opened 2024-01-17
run func tests on GitHub instead of self-hosted

rgildein:
Running functional tests on GitHub runner instead of self-hosted. With GitHub runner we do not need to have concurrency set-up.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -66,11 +66,11 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-      - name: Setup Juju environment
+      - name: Setup Juju 2.9/stable environment
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          juju-channel: stable
+          juju-channel: 2.9/stable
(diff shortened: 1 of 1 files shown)


agileshaw (commented):
If we no longer uses self-hosted runners, then can't we just call reusable workflows in `bootstack-action` for checks like other repos?

rgildein:
> If we no longer uses self-hosted runners, then can't we just call reusable workflows in `bootstack-action` for checks like other repos?

Maybe we can, I just tried to not using it. We can use bootstack-actions in other PR.

agileshaw:
> Maybe we can, I just tried to not using it. We can use bootstack-actions in other PR.

It would be a much simpler change with more straightforward maintainability using bootstack-actions.

rgildein:
> > Maybe we can, I just tried to not using it. We can use bootstack-actions in other PR.
> 
> It would be a much simpler change with more straightforward maintainability using bootstack-actions.

If you want you con provide PR for that, with this I was going to tests if it works. I'm ok to merge this, since it's not different that what we had expect, no need to have our own runner.

rgildein:
replaced by #244

PR #224, opened 2024-01-18
Add force flag to generate plan and upgrate non-empty hypervisors

gabrielcocenza:
- introduction of "--force" that can be used on "plan" and "upgrade"
- usage of argparse.SUPPRESS to not [overwrite]( https://stackoverflow.com/questions/62904585/support-global-arguments-before-or-after-sub-command-in-argparse) the parent parser value
- more unit tests changing the sequence of subcommands to show that overwriting it's not happening anymore.


agileshaw (changes requested):
Thanks for observing the issue and apply fixes. A few comments inline.

agileshaw (approved):
LGTM :)

rgildein (approved):
LGTM

PR #225, opened 2024-01-19
Add instance count helper function and refactor action running methods

agileshaw:
- Created a helper function _get_instance_count() to get the instance count of a nova-compute unit
- Refactor to action related methods in COUModel

### DIFF
--- a/cou/utils/app_utils.py
+++ b/cou/utils/app_utils.py
@@ -53,11 +53,11 @@ async def upgrade_packages(
 async def get_instance_count(unit: str, model: COUModel) -> int:
     """Get instance count on a nova-compute unit.
 
-    :param unit: Name of the unit where the action runs on.
+    :param unit: Name of the nova-compute unit where the action runs on.
     :type unit: str
     :param model: COUModel object
     :type model: COUModel
-    :return: Instance count of the unit
+    :return: Instance count of the nova-compute unit
     :rtype: int
     :raises ValueError: When the action result is not valid.
     """
--- a/cou/utils/app_utils.py
+++ b/cou/utils/app_utils.py
@@ -65,7 +65,7 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
     action = await model.run_action(unit_name=unit, action_name=action_name)
 
     if (
-        instance_count := action.results.get("instance-count").strip()
+        instance_count := action.results.get("instance-count", "").strip()
     ) and instance_count.isdigit():
         return int(instance_count)
 
--- a/cou/utils/app_utils.py
+++ b/cou/utils/app_utils.py
@@ -50,6 +50,31 @@ async def upgrade_packages(
         await model.run_on_unit(unit_name=unit, command=command, timeout=600)
 
 
+async def get_instance_count(unit: str, model: COUModel) -> int:
+    """Get instance count on a nova-compute unit.
+
+    :param unit: Name of the unit where the action runs on.
+    :type unit: str
+    :param model: COUModel object
+    :type model: COUModel
+    :return: Instance count of the unit
+    :rtype: int
+    :raises ValueError: When the action result is not valid.
+    """
+    action_name = "instance-count"
+    action = await model.run_action(unit_name=unit, action_name=action_name)
+
+    if (
+        instance_count := action.results.get("instance-count").strip()
+    ) and instance_count.isdigit():
+        return int(instance_count)
+
+    raise ValueError(
+        f"No valid instance count value found in the result of {action_name} action "
+        f"running on '{unit}': {action.results}"
+    )
+
+
 async def set_require_osd_release_option(unit: str, model: COUModel) -> None:
     """Check and set the correct value for require-osd-release on a ceph-mon unit.
 
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -168,27 +168,27 @@ class COUModel:
         return self._name
 
     @retry(no_retry_exceptions=(ActionFailed,))
-    async def _get_action_result(self, action: Action, raise_on_failure: bool) -> Action:
-        """Get results from action.
+    async def _get_waited_action_object(self, action: Action, raise_on_failure: bool) -> Action:
+        """Get waited action object.
+
+        To access action data from the returned action object, use `action_obj.data`, which
+        contains action parameters, results, status, and metadata. Alternatively, it is possible
+        to access action results directly with `action_obj.results`.
 
         :param action: Action object
         :type: Action
         :param raise_on_failure: Whether to raise ActionFailed exception on failure, defaults
                                  to False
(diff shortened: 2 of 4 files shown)


cou/utils/app_utils.py
```diff
@@ -50,6 +50,31 @@ async def upgrade_packages(
...
+    :type model: COUModel
+    :return: Instance count of the unit
+    :rtype: int
+    :raises ValueError: When the action result is not valid.
+    """
+    action_name = "instance-count"
+    action = await model.run_action(unit_name=unit, action_name=action_name)
+
+    if (
+        instance_count := action.results.get("instance-count").strip()
```
rgildein, line 68: not-blocking: What is the benefit of walrus operator here? IMO it is making it only less readable.

rgildein: After second thought, this is actually blocking comment, because if `"instance-count"` does not exist in action results, the get function will return `None`. Like this function can end up in `AttributeError`.

cou/utils/juju_utils.py
```diff
@@ -168,27 +168,27 @@ def name(self) -> str:
...
-        result = await action.wait()
-        if raise_on_failure:
-            model = await self._get_model()
-            status = await model.get_action_status(uuid_or_prefix=action.entity_id)
-            if status != "completed":
-                output = await model.get_action_output(action.entity_id)
-                raise ActionFailed(action, output=output)
+        action_obj = await action.wait()
+        if raise_on_failure and action_obj.status != "completed":
+            raise ActionFailed(action, output=action_obj.data)
```
rgildein, line 189: Previous approach was chosen by OpenStak team in zaza, I'm not sure why. @gabrielcocenza do you know if it's important?

cou/utils/juju_utils.py
```diff
@@ -168,27 +168,27 @@ def name(self) -> str:
         return self._name
 
     @retry(no_retry_exceptions=(ActionFailed,))
-    async def _get_action_result(self, action: Action, raise_on_failure: bool) -> Action:
-        """Get results from action.
+    async def _get_waited_action_object(self, action: Action, raise_on_failure: bool) -> Action:
```
rgildein, line 171: not-blocking: In output, we are saying that it's awaited, so why not also `_get_awaited_action_object`? IMO changing the name is not necessary.

rgildein (changes requested):

cou/utils/app_utils.py
agileshaw: I chose walrus operator because this variable is not used outside this conditional block. 

I agree with the issue you described and will add a default value for "get()" in this case.

cou/utils/juju_utils.py
agileshaw: `unit.run_action` is also a sync command with `awaits`, so I think `_get_awaited_action_object` would a misleading name. I would prefer "waited" because it indicates the usage of `action.wait()` function.

Also, since the function is not returning just the action results but a action object with results as part of it, I renamed the function for more clarity.

cou/utils/juju_utils.py
agileshaw: I just checked zaza, they are doing the [same type of comparison](https://github.com/openstack-charmers/zaza/blob/master/zaza/model.py#L1099) (I'm surprised that they are even using the same variable name as mine :laughing: ). The only difference is how they are getting the output, which I don't know if there's any additional benefits getting from model instead of the action itself.

rgildein: I think that this is fine, I have no idea why it was like this in the first place, and I like your version.

rgildein (approved):
LGTM

cou/utils/app_utils.py
```diff
@@ -50,6 +50,31 @@ async def upgrade_packages(
         await model.run_on_unit(unit_name=unit, command=command, timeout=600)
 
 
+async def get_instance_count(unit: str, model: COUModel) -> int:
```
gabrielcocenza, line 53:
```suggestion
async def get_nova_compute_instance_count(unit: str, model: COUModel) -> int:
```

cou/utils/app_utils.py
```diff
@@ -50,6 +50,31 @@ async def upgrade_packages(
...
+    """Get instance count on a nova-compute unit.
+
+    :param unit: Name of the unit where the action runs on.
+    :type unit: str
+    :param model: COUModel object
+    :type model: COUModel
+    :return: Instance count of the unit
```
gabrielcocenza, line 60:
```suggestion
    :return: Instance count of the nova-compute unit
```

cou/utils/juju_utils.py
gabrielcocenza: I have no idea....hahahaha I also enjoyed the approach :+1:

tests/unit/utils/test_app_utils.py
```diff
@@ -78,6 +79,38 @@ async def test_application_upgrade_packages_with_hold(model):
...
+    ],
+)
+async def test_get_instance_count_invalid_result(model, result_key, value):
+    model.run_action.return_value = mocked_action = AsyncMock(spec_set=Action).return_value
+    mocked_action.results = {"Code": "0", result_key: value}
+
+    with pytest.raises(ValueError):
+        await app_utils.get_instance_count(unit="nova-compute/0", model=model)
+
+
```
gabrielcocenza, line 113: Does it worth adding a test that shows the behavior if we pass a unit different than `nova-compute` to `.get_instance_count`

gabrielcocenza (commented):
I left some non-blocker comments, but I enjoyed the approach. Thanks

cou/utils/app_utils.py
agileshaw: I would prefer keeping the function name short. I improved the docstrings based on your suggestion so I think it is clearer now.

tests/unit/utils/test_app_utils.py
agileshaw: Probably not at this stage since we are using it only internally. We didn't have unit checks for other helper functions either.

cou/utils/app_utils.py
rgildein: Since we are passing the name of unit, it can be `cinder/0` or `Harry Potter`, where both values should fail. So we do not check the name, so I do not see any need to rename this function.

tests/unit/utils/test_app_utils.py
rgildein: The `model.run_action` will raise `ActionFailed`, so we do not need to test more.

rgildein (approved):

cou/utils/app_utils.py
gabrielcocenza: So it's two against one :sweat_smile: . No big deal, solving it.

gabrielcocenza (approved):

PR #226, opened 2024-01-23
Machine Class

gabrielcocenza:
- Introduction of the machine class with the following fields:
  - hostname, az, id and is_data_plane
- juju_utils generate COU machine class based on juju Application data
- Apps receive the machines with units deployed as attribute during the creation of the instance.
- changed conftests to have a single source of truth for the unit tests.

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -24,6 +24,7 @@ from typing import Any, Optional
 from juju.client._definitions import ApplicationStatus, UnitStatus
 from ruamel.yaml import YAML
 
+from cou.apps.machine import Machine
 from cou.exceptions import (
     ApplicationError,
     HaltUpgradePlanGeneration,
@@ -54,8 +55,8 @@ class ApplicationUnit:
 
     name: str
     os_version: OpenStackRelease
+    machine: Machine
     workload_version: str = ""
-    machine: str = ""
 
 
 @dataclass
@@ -99,6 +100,7 @@ class OpenStackApplication:
     config: dict
     model: COUModel
     charm: str
+    machines: dict[str, Machine]
     charm_origin: str = ""
     os_origin: str = ""
     origin_setting: Optional[str] = None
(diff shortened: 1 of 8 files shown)


cou/utils/juju_utils.py
```diff
@@ -499,3 +509,27 @@ async def _wait_for_active_idle() -> None:
...
+
+        :return: Dictionary of the machines found in the model. E.g: {'0': Machine0}
+        :rtype: dict[str, Machine]
+        """
+        juju_applications = await self.get_applications()
+        machines: dict[str, Machine] = {}
+        for app in juju_applications.values():
+            charm_name = app.charm_name
+            for unit in app.units:
+                unit_machine = unit.machine.data
```
rgildein, line 524: I think it would be easier if we can do (not sure if it's possible):
```python
for juju_machine in model.machines:
    is_data_plane = any(app.charm_name in DATA_PLANE_CHARMS for app in juju_machine.apps)
    machine = Machine(
        machine_id=juju_machine["id"],
        hostname=juju_machine["hostname"],
        az=juju_machine["hardware-characteristic"].get("availability-zone"),
        is_data_plane=is_data_plane
    )
    ...
```

cou/apps/base.py
```diff
@@ -370,16 +372,8 @@ def channel_codename(self) -> OpenStackRelease:
         :return: OpenStackRelease object
         :rtype: OpenStackRelease
         """
-        try:
```
gabrielcocenza, line 373: I was not able to find why coverage wasn't complaining about this before. We are already checking if channel is valid, so I was not able to write a unit tests that would trigger this `ValueError` and that is why I'm removing it

cou/steps/analyze.py
```diff
@@ -105,7 +91,11 @@ async def create(cls, model: juju_utils.COUModel) -> Analysis:
 
         control_plane, data_plane = cls._split_apps(apps)
 
-        return Analysis(model=model, apps_data_plane=data_plane, apps_control_plane=control_plane)
+        return Analysis(
+            model=model,
+            apps_data_plane=data_plane,
+            apps_control_plane=control_plane,
+        )
```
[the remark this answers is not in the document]
gabrielcocenza: Probably I've added a new parameter, removed and then the multiple lines kept. Set back again to single line

PR #227, opened 2024-01-25
Add base architecture of HypervisorUpgradePlanner

rgildein:
This commit introduce the HypervisorUpgradePlanner, which takes apps as argument and group them by AZ and hypervisor. Like this it can easily generate plan to upgrade all hypervisors.

### DIFF
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -293,8 +293,8 @@ class UpgradePlan(BaseStep):
         logger.debug("No coroutine to run for %s", repr(self))
 
 
-class AvailabilityZoneUpgradePlan(UpgradePlan):
-    """Represents the plan for upgrading availability zone upgrade.
+class HypervisorGroupUpgradePlan(UpgradePlan):
+    """Represents the plan for group of hypervisors.
 
     This class is intended to be used as group for hypervisor upgrade steps, therefore
     doesn't accept coroutine or parallel as inputs.
@@ -317,7 +317,7 @@ class HypervisorUpgradePlan(BaseStep):
     """Represents the plan for hypervisor upgrade.
 
     This class is intended to be used as a group for unit-level upgrade steps, which are
-    grouped by single hypervisor. It doesn't accept coroutine or parallel as inputs.
+    grouped by a single hypervisor. It doesn't accept coroutine or parallel as inputs.
 
     All sub-steps will run in parallel.
     """
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -293,6 +293,16 @@ class UpgradePlan(BaseStep):
         logger.debug("No coroutine to run for %s", repr(self))
 
 
+class AvailabilityZoneUpgradePlan(UpgradePlan):
+    """Represents the plan for upgrading availability zone upgrade.
+
+    This class is intended to be used as group for hypervisor upgrade steps, therefore
+    doesn't accept coroutine or parallel as inputs.
+    """
+
+    prompt: bool = True
+
+
 class ApplicationUpgradePlan(UpgradePlan):
     """Represents the plan for application-level upgrade.
 
@@ -301,6 +311,26 @@ class ApplicationUpgradePlan(UpgradePlan):
     """
 
     prompt: bool = True
+
+
+class HypervisorUpgradePlan(BaseStep):
+    """Represents the plan for hypervisor upgrade.
+
+    This class is intended to be used as a group for unit-level upgrade steps, which are
+    grouped by single hypervisor. It doesn't accept coroutine or parallel as inputs.
+
+    All sub-steps will run in parallel.
+    """
+
+    prompt: bool = False
+
+    def __init__(self, description: str):
+        """Initialize upgrade plan.
+
+        :param description: Description of the step.
+        :type description: str
+        """
+        super().__init__(description=description, parallel=True, coro=None)
 
 
 class UpgradeStep(BaseStep):
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -1,0 +1,152 @@
+# Copyright 2024 Canonical Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Hypervisor planner."""
+import logging
+from dataclasses import dataclass
+
+from cou.apps.base import ApplicationUnit, OpenStackApplication
+from cou.steps import (
+    HypervisorUpgradePlan,
+    PostUpgradeStep,
+    PreUpgradeStep,
+    UpgradePlan,
+)
+from cou.utils.openstack import OpenStackRelease
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class Hypervisor:
+    """Hypervisor containing units for multiple applications."""
+
+    name: str
+    units: list[ApplicationUnit]
+
+
+@dataclass(frozen=True)
+class AZ:
+    """Availability Zone hypervisors in the az."""
+
+    name: str
+    hypervisors: list[Hypervisor]
+
+
+def verify_apps(apps: list[OpenStackApplication]) -> None:
+    """Verify OpenStack applications.
+
+    This function will verify that all machines contain nova-compute application.
+
+    :param apps: list of OpenStack applications
+    :type apps: list[OpenStackApplication]
+    """
+    raise NotImplementedError
+
+
+class HypervisorUpgradePlanner:
+    """Planer for all hypervisor updates.
+
+    This planner is meant to be used to upgrade machines contains the nova-compute application.
+    """
+
+    def __init__(self, apps: list[OpenStackApplication]) -> None:
+        """Initialize the Hypervisor class.
+
+        :param apps: list of OpenStack applications
+        :type apps: list[OpenStackApplication]
+        """
+        verify_apps(apps)
+        self._apps = apps
+
+    @property
+    def apps(self) -> list[OpenStackApplication]:
+        """Return list of apps in hypervisor class."""
+        return self._apps
+
+    @property
+    def azs(self) -> list[AZ]:
+        """Returns list of AZs defined in individual applications.
+
+        Each AZ contains a sorted list hypervisors, where each hypervisor has sorted list of units.
+        The order of units depends on the order in which they needs to be upgraded.
+        eg.
+        az1:
+        - hypervisor0:
+          - nova-compute/0
+          - cinder/0
+        - hypervisor1:
+          - nova-compute/1
+          - cinder/1
+        ...
+        """
+        raise NotImplementedError
+
+    def _generate_pre_upgrade_plan(self) -> PreUpgradeStep:
+        """Generate pre upgrade plan all application.
+
+        This section should create a plan with steps like changing Juju config option, etc.
+
+        :return: Pre-upgrade step with all needed pre-upgrade steps.
+        :rtype: PreUpgradeStep
+        """
+        raise NotImplementedError
+
+    def _generate_hypervisor_upgrade_plan(self, hypervisor: Hypervisor) -> HypervisorUpgradePlan:
+        """Genarete upgrade plan for single hypevisors.
+
+        Each hypervisor upgrade consists of a UnitUpgradeStep, so all substeps should be based
+        on UnitUpgradeStep.
+
+        :param hypervisor: hypervisor object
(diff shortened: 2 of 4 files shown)


cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+
+        :param hypervisor: hypervisor object
+        :type hypervisor: Hypervisor
+        :return: Upgrade plan for hypervisor.
+        :rtype: HypervisorUpgradePlan
+        """
+        raise NotImplementedError
+
+    def _generate_post_upgrade_plan(self) -> PostUpgradeStep:
+        """Generate post upgrade plan all application.
```
agileshaw, line 120:
```suggestion
        """Generate post upgrade plan for all applications.
```

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+        self._apps = apps
+
+    @property
+    def apps(self) -> list[OpenStackApplication]:
+        """Return list of apps in hypervisor class."""
+        return self._apps
+
+    @property
+    def azs(self) -> list[AZ]:
+        """Returns list of AZs defined in individual applications.
```
agileshaw, line 80:
```suggestion
        """Returns a list of AZs defined in individual applications.
```

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+        The order of units depends on the order in which they needs to be upgraded.
+        eg.
+        az1:
+        - hypervisor0:
+          - nova-compute/0
+          - cinder/0
+        - hypervisor1:
+          - nova-compute/1
+          - cinder/1
+        ...
```
agileshaw, line 92: Need to add return and return type

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+        verify_apps(apps)
+        self._apps = apps
+
+    @property
+    def apps(self) -> list[OpenStackApplication]:
+        """Return list of apps in hypervisor class."""
+        return self._apps
+
+    @property
+    def azs(self) -> list[AZ]:
```
agileshaw, line 79: It's unclear to me how the AZs are determined through application. But since this is only the base structure, we can leave it as "magic" behind the scene

pyproject.toml
```diff
@@ -80,7 +80,15 @@ exclude = [
...
 source = ["."]
-omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py"]
+omit = [
+    "tests/**",
+    "docs/**",
+    "lib/**",
+    "snap/**",
+    "build/**",
+    "setup.py",
+    "cou/steps/hypervisor.py",
```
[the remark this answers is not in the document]
rgildein: This is for dev branch and without this CI will failes, which we do not want to do. 
We will do PR from dev branch to main, and if we missing thinks like this <==> bad review, if you have a better idea how we can not break CI and not putting it here, I'll go with it.

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    def __init__(self, apps: list[OpenStackApplication]) -> None:
+        """Initialize the Hypervisor class.
+
+        :param apps: list of OpenStack applications
+        :type apps: list[OpenStackApplication]
+        """
+        verify_apps(apps)
+        self._apps = apps
```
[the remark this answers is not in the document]
rgildein: I was thinking about it, but I do not want to set those apps multiple times (what setter allows), so I decide to stick with this approach.

cou/steps/hypervisor.py
rgildein: yes, this will be more tricky than just for loop, but I did not want to focus on that here. Next PR

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+    """Availability Zone hypervisors in the az."""
+
+    name: str
+    hypervisors: list[Hypervisor]
+
+
+def verify_apps(apps: list[OpenStackApplication]) -> None:
+    """Verify OpenStack applications.
+
+    This function will verify that all machines contain nova-compute application.
```
gabrielcocenza, line 50: I didn't understand what you are going to do here. Are you going to some how filter all the applications that are colocated with nova-compute?

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class Hypervisor:
+    """Hypervisor containing units for multiple applications."""
+
+    name: str
+    units: list[ApplicationUnit]
```
gabrielcocenza, line 36: Isn't this a machine in the end of the day? I'm not seeing much difference if the Machine class says which units are deployed on it. It seems that this Hypervisor is behaving like a "special" machine because it will have nova-compute on it.

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+    @property
+    def azs(self) -> list[AZ]:
+        """Returns list of AZs defined in individual applications.
+
+        Each AZ contains a sorted list hypervisors, where each hypervisor has sorted list of units.
+        The order of units depends on the order in which they needs to be upgraded.
+        eg.
+        az1:
+        - hypervisor0:
+          - nova-compute/0
```
gabrielcocenza, line 87: nit: cinder will be upgraded first than nova-compute. It might be good changing this example to match reality.

pyproject.toml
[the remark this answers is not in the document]
gabrielcocenza: I think it's good to add a comment like `TODO` to remove it later.

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+class Hypervisor:
+    """Hypervisor containing units for multiple applications."""
+
+    name: str
+    units: list[ApplicationUnit]
+
+
+@dataclass(frozen=True)
+class AZ:
+    """Availability Zone hypervisors in the az."""
```
gabrielcocenza, line 41:
```suggestion
    """Collection of hypervisors in an Availability Zone."""
```

pyproject.toml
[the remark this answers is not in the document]
agileshaw: Yea I suggested this because there were no CI for this branch. But now that you've updated `check.yaml`, I think we can keep it as is. Would be nice to track it somehow, but I don't have better ideas.

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,152 @@
...
+@dataclass(frozen=True)
+class Hypervisor:
+    """Hypervisor containing units for multiple applications."""
+
+    name: str
+    units: list[ApplicationUnit]
+
+
+@dataclass(frozen=True)
+class AZ:
```
gabrielcocenza, line 40: I don't think naming AZ here is good because an AZ is more than just a collection of hypervisors. Maybe we can call something else like `HypervisorsGrouping` or `HypervisorsCollection`.

If for some reason in the future someone wants to group the Hypervisors by a different algorithm like just for empty VMs, this class will still  have a representative name.

cou/steps/hypervisor.py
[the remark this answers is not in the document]
gabrielcocenza: Maybe we can use [cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property)?

gabrielcocenza (changes requested):

pyproject.toml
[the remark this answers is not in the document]
rgildein: TODO will cause further issues with linting. Yes, because of your comment, I realized that CI is not working.

rgildein:
> Generally looks good for what this PR is intended to do, but there are pieces in the overall approach remain unclear to me.
> 
> It seems that Machine and Hypervisor classes are of the same scope, but Machine contains hardware information whereas Hypervisor contains software resources (which units are running on). But not sure how the logic between OpenStack Application, ApplicationUnit, Machine, Hypervisor, AZ would be established.

No the Hypervisor class contain information about OpenStack Applications, which indeed contain information about units, I want to use their functions to generate plan to generate plan itself.

cou/steps/hypervisor.py
rgildein: It's possible that we will use the machine class, I'm not sure until it will be used. This was done before Machine class exists and now, you are maybe right we do not need it anymore.

cou/steps/hypervisor.py
rgildein: The name is verify, I just want to make sure that machine passing to the HypervisorUpgradePlanner via apps has nova-compute on it, so yes I will verify that apps are either nova-compute or colocated apps.

It's possible that this will change to filter function and we will use it during creation of HypervisorUpgradePlanner.

cou/steps/hypervisor.py
rgildein: That sound like good idea, since both AZ and the Hypervisors are just placeholder for grouped objects, we could easily used just `list[lists[lists[ApplicationUnit]]]`. In next PR, where I will introduce the functionality of grouping we can change this names (potentially use  also Machine class instead of Hypervisor).

These objects are just for better imagination for now, we can discuss further comment in PR we they will be used.

cou/steps/hypervisor.py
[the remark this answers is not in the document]
rgildein: The cached_property is for situation where you have dynamic getting of  those objects, e.g. if we have `get_application` function. Not for static object, where we pass list of apps.

cou/steps/hypervisor.py
agileshaw: I actually don't mind the separation between machine and hypervisor, but also have no objection on making them associated/inherited.

agileshaw (commented):
> No the Hypervisor class contain information about OpenStack Applications, which indeed contain information about units, I want to use their functions to generate plan to generate plan itself.

From your current implementation, Hypervisor class has only a list of ApplicationUnits but no correlation of OpenStackApplication class itself. So I'm not sure I understand what you mean here by "contain information about OpenStack Applications"

rgildein:
> > No the Hypervisor class contain information about OpenStack Applications, which indeed contain information about units, I want to use their functions to generate plan to generate plan itself.
> 
> From your current implementation, Hypervisor class has only a list of ApplicationUnits but no correlation of OpenStackApplication class itself. So I'm not sure I understand what you mean here by "contain information about OpenStack Applications"

Sorry, I did not understand your message properly. The `HypervisorUpgradePlanner` has all information about `OpenStackApplication` and the `Hypervisor` is just a group of units, where these units can be from multiple applications.

cou/steps/hypervisor.py
gabrielcocenza: I think we don't need to double check that the machines passed to the hypervisors are nova-compute, because this check should happen before when we are validating user input.

However I think we'll need to do other stuff, so I'm not blocking this.

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,164 @@
...
+    PreUpgradeStep,
+    UpgradePlan,
+)
+from cou.utils.openstack import OpenStackRelease
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class HypervisorMachine:
```
gabrielcocenza, line 32: Is it too soon to inherit from `Machine`? I guess in the end of the day we will end up with the normal machine class. This hypervisor machine is missing IMO a easy way to access `az`. If we are grouping by AZ, the way it is right now we would need to access units -> unit -> machine -> az

gabrielcocenza (commented):
The PR has some merge conflicts with the dev branch. Generally LGTM, but I left some non blockers comments

cou/steps/__init__.py
```diff
@@ -293,6 +293,16 @@ async def run(self) -> None:
         logger.debug("No coroutine to run for %s", repr(self))
 
 
+class AvailabilityZoneUpgradePlan(UpgradePlan):
```
agileshaw, line 296: non-blocking: we are using `HypervisorGroup` class for type-agnostic grouping of the hypervisor machines, I think this should be updated to follow the same approach

cou/steps/hypervisor.py
```diff
@@ -0,0 +1,164 @@
...
+    This function will verify that all machines contain nova-compute application.
+
+    :param apps: list of OpenStack applications
+    :type apps: list[OpenStackApplication]
+    """
+    raise NotImplementedError
+
+
+class HypervisorUpgradePlanner:
+    """Planer for all hypervisor updates.
```
agileshaw, line 62:
```suggestion
    """Planner for all hypervisor updates.
```

cou/steps/hypervisor.py
agileshaw: I share a similar concern (which is partly what I wanted to express in my initial review comments). IIUC the logic chain is even longer when trying to instantiate a `HypervisorMachines` class: we need to go through `HypervisorUpgradePlanner` -> OpenStackApplications -> units -> unit -> machine -> AZ.

rgildein: Yes, it's since the Machine will change. We can discuss this later when it will be used, these are just pseudo code.

gabrielcocenza: Ok, closing to see the future changes.

gabrielcocenza (approved):

agileshaw (approved):

PR #228, opened 2024-01-25
Sanity checks for machines id, hostnames and availability zones

gabrielcocenza:
- verify if the parameters passed exist and if they are data-plane related
- fix lint to pass the recent [black](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#configuration) changes


cou/steps/plan.py
```diff
@@ -172,6 +174,138 @@ def determine_upgrade_target(analysis_result: Analysis) -> OpenStackRelease:
     return target
 
 
+def verify_data_plane_cli_input(args: CLIargs, analysis_result: Analysis) -> None:
+    """Sane checks from the parameters passed in the cli to upgrade data-plane.
```
agileshaw, line 178:
```suggestion
    """Sanity checks from the parameters passed in the cli to upgrade data-plane.
```

cou/steps/analyze.py
```diff
@@ -203,3 +204,38 @@ def _get_minimum_cloud_series(self) -> Optional[str]:
...
+    def data_plane_machines(self) -> dict[str, Machine]:
+        """Data-plane machines of the model.
+
+        :return: Data-plane machines of the model.
+        :rtype: dict[str, Machine]
+        """
+        return {
+            machine_id: app.machines[machine_id]
+            for app in self.apps_data_plane
+            for machine_id in app.machines
```
agileshaw, line 218: Nice compact piece of code :+1:

cou/steps/plan.py
```diff
@@ -172,6 +174,138 @@ def determine_upgrade_target(analysis_result: Analysis) -> OpenStackRelease:
...
+    }
+
+    # NOTE(gabrielcocenza) simple deployments may not have AZs. mypy does not understand
+    # that None is being removed from set, so the type goes from set[Optional[str]] to
+    # set[str]. As we can be sure of that, we can cast with the correct type to mypy.
+    all_azs_uncast.discard(None)
+    data_plane_azs_uncast.discard(None)
+
+    all_azs = cast(set[str], all_azs_uncast)
+    data_plane_azs = cast(set[str], data_plane_azs_uncast)
```
agileshaw, line 255: Perhaps it's better to also handle this in COUModel like how you are getting the machines?

cou/steps/plan.py
```diff
@@ -172,6 +174,138 @@ def determine_upgrade_target(analysis_result: Analysis) -> OpenStackRelease:
...
+
+def parametrize_cli_inputs(cli_input: list[str]) -> set[str]:
+    """Parametrize the cli inputs.
+
+    :param cli_input: cli inputs.
+    :type cli_input: list[str]
+    :return: A set of elements passed in the cli.
+    :rtype: set[str]
+    """
+    return {raw_item.strip() for raw_items in cli_input for raw_item in raw_items.split(",")}
```
agileshaw, line 278: Wouldn't it makes more sense to handle this in the CLIArgs class?

agileshaw (changes requested):

cou/steps/plan.py
gabrielcocenza: I think it does. Thanks!

cou/steps/analyze.py
```diff
@@ -203,3 +204,38 @@ def _get_minimum_cloud_series(self) -> Optional[str]:
             (app.series for app in self.apps_control_plane + self.apps_data_plane),
             default=None,
         )
+
+    @property
+    def data_plane_machines(self) -> dict[str, Machine]:
```
rgildein, line 209: I still do not understand why I would need to have dictionary instead list of machine.

rgildein: BTW, here is better place where you can use cached_property.

cou/steps/plan.py
rgildein, line 255:
```suggestion
    all_azs_uncast: set[str] = {
        machine.az for machine in analysis_result.machines.values() if machine.az is not None
    }
    data_plane_azs_uncast: set[str] = {
        machine.az for machine in analysis_result.data_plane_machines.values() if machine.az is not None
    }
```

cou/steps/plan.py
```diff
@@ -172,6 +174,138 @@ def determine_upgrade_target(analysis_result: Analysis) -> OpenStackRelease:
...
+
+
+def data_plane_membership_check(
```
rgildein, line 281: I think we should rename it to `check_data_plane_membership`.

rgildein (changes requested):
Small change required, other than that +1

cou/steps/analyze.py
gabrielcocenza: Here it could be a list, I don't see a problem. However the machines passed here it's a dictionary that makes easier for populating units.

```python
apps = {
            AppFactory.create(
                name=app,
                status=app_status,
                config=await model.get_application_config(app),
                model=model,
                charm=await model.get_charm_name(app),
                machines={
                    unit_status.machine: juju_machines[unit_status.machine]
                    for unit_status in app_status.units.values()
                },
            )
            for app, app_status in juju_status.applications.items()
            if app_status
        }
        
def _populate_units(self) -> None:
    """Populate application units."""
    if not self.is_subordinate:
        for name, unit in self.status.units.items():
            compatible_os_version = self._get_latest_os_version(unit)
            self.units.append(
                ApplicationUnit(
                    name=name,
                    workload_version=unit.workload_version,
                    os_version=compatible_os_version,
                    machine=self.machines[unit.machine],
                )
            )
```

I'm thinking to keep the current data structure and if we change the way that we populate units, we can re-think and maybe change to a list. I prefer to stick with dictionary because it can easily access if necessary and there is no difference from using `.values()` to have a list of machines

cou/steps/plan.py
gabrielcocenza: Way better :+1: . Thanks

tests/unit/steps/test_steps_plan.py
```diff
@@ -441,3 +451,156 @@ def test_analysis_not_print_warn_manually_upgrade(mock_print, model, apps):
...
+
+    if hostnames:
+        mock_verify_machines.assert_not_called()
+        mock_verify_hostnames.assert_called_once()
+        mock_verify_azs.assert_not_called()
+
+    if azs:
+        mock_verify_machines.assert_not_called()
+        mock_verify_hostnames.assert_not_called()
+        mock_verify_azs.assert_called_once()
```
agileshaw, line 503: non-blocking: If we need to have this many conditions, it would be better separating them into different functions instead of forcing them into one parametrized test function.

tests/unit/steps/test_steps_plan.py
```diff
@@ -441,3 +451,156 @@ def test_analysis_not_print_warn_manually_upgrade(mock_print, model, apps):
...
+
+
+@pytest.mark.parametrize(
+    "machine",
+    [
+        {NOVA_MACHINES[0]},
+        {NOVA_MACHINES[0], NOVA_MACHINES[1], NOVA_MACHINES[2]},
+    ],
+)
+def test_verify_data_plane_cli_machines(apps, model, machine):
```
agileshaw, line 513: non-blocking: We could merge `verify_data_plane_cli_input` test with machine input to this test. Similar for hostname and az tests below.

agileshaw (approved):
LGTM. I left some suggestion on refactoring test codes, but they are not blockers.

cou/steps/analyze.py
rgildein: Ok, I agree, and I'm working on that change, so we will see.

cou/commands.py
```diff
@@ -151,7 +151,7 @@ def get_dataplane_common_opts_parser() -> argparse.ArgumentParser:
         action="append",
         help="Specify machine id(s) to upgrade.\nThis option accepts a single machine id as well "
         "as a stringified comma-separated list of ids,\nand can be repeated multiple times.",
-        dest="machines",
+        dest="raw_machines",
```
rgildein, line 154: Instead of this, we can define our own action to get all machines we need.

```python
class SplitArgs(argparse._AppendAction):
    def __call__(self, parser, namespace, values, option_string=None):
        for value in values.split(","):
            super().__call__(parser, namespace, value, option_string)
...
    dp_upgrade_group.add_argument(
        "--machine",
        "-m",
        action=SplitArgs,
...
```
Like this, we can parse machine defined as list separated by comma or multiple arguments.


However, I do not see a reason why we are supporting both? It would not be enough to support only one way? e.g. `--machine 1 --machine 2` or `--machines 1,2`

rgildein (changes requested):

tests/unit/test_commands.py
```diff
@@ -553,9 +553,9 @@ def test_parse_args_plan(args, expected_CLIargs):
                 auto_approve=False,
                 backup=True,
                 force=False,
-                raw_machines=None,
-                raw_hostnames=["1", "2,3"],
-                raw_availability_zones=None,
+                machines=None,
+                hostnames={"1", "2", "3"},
```
agileshaw, line 557: non-blocking: since you are using set, it would be good to add a test for an input like `-m 1,2,3 -m 3,5`

agileshaw (approved):
TIL as well. Looks good, just one non-blocking suggestion inline.

cou/commands.py
```diff
@@ -71,6 +75,27 @@ def add_argument(self, action: argparse.Action) -> None:
...
+    ) -> None:
+        logger.debug("SplitArgs: %s %s %s %s", parser, namespace, values, option_string)
+        new_items = {item.strip() for item in values.split(",")}
+        cli_input = getattr(namespace, self.dest) or set()
+        cli_input.update(new_items)
+        setattr(
+            namespace,
+            self.dest,
+            cli_input,
+        )
```
rgildein, line 96: not-blocking: 
```suggestion
        setattr(namespace, self.dest, cli_input)
```

rgildein (approved):

pyproject.toml
```diff
@@ -7,6 +7,7 @@ ignore = ["C901", "D100", "D101", "D102", "D103", "W503", "W504"]
 exclude = ['.eggs', '.git', '.tox', '.venv', '.build', 'build', 'report', 'docs']
 max-line-length = 99
 max-complexity = 10
+extend-ignore = ["E203", "E704"]
```
gabrielcocenza, line 10: Lint was not passing with the recent changes on [black](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#configuration). That is why this config was necessary

rgildein (approved):
LGTM

agileshaw (approved):

PR #229, opened 2024-01-26
Filter hypervisors.

gabrielcocenza:
- do the nova-compute instance-count to find empty hypervisors
- integrate the force flag to include non-empty hypervisors

PR shoul be merged after #228

### DIFF
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -397,7 +397,6 @@ async def filter_hypervisors_machines(args: CLIargs, analysis_result: Analysis) 
         return {machine for machine in hypervisors_machines if machine.machine_id in cli_machines}
 
     if cli_hostnames := args.hostnames:
-        print(cli_hostnames)
         return {machine for machine in hypervisors_machines if machine.hostname in cli_hostnames}
 
     if cli_azs := args.availability_zones:
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -379,7 +379,7 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     return plan
 
 
-async def filter_hypervisors_machines(args: CLIargs, analysis_result: Analysis) -> set[Machine]:
+async def filter_hypervisors_machines(args: CLIargs, analysis_result: Analysis) -> list[Machine]:
     """Filter the hypervisors to generate plan and upgrade.
 
     :param args: CLI arguments
@@ -387,45 +387,45 @@ async def filter_hypervisors_machines(args: CLIargs, analysis_result: Analysis) 
     :param analysis_result: Analysis result
     :type analysis_result: Analysis
     :return: hypervisors filtered to generate plan and upgrade.
-    :rtype: set[Machine]
+    :rtype: list[Machine]
     """
     hypervisors_machines = await _get_upgradable_hypervisors_machines(args.force, analysis_result)
 
     if cli_machines := args.machines:
-        return {machine for machine in hypervisors_machines if machine.machine_id in cli_machines}
+        return [machine for machine in hypervisors_machines if machine.machine_id in cli_machines]
 
     if cli_hostnames := args.hostnames:
-        return {machine for machine in hypervisors_machines if machine.hostname in cli_hostnames}
+        return [machine for machine in hypervisors_machines if machine.hostname in cli_hostnames]
 
     if cli_azs := args.availability_zones:
-        return {machine for machine in hypervisors_machines if machine.az in cli_azs}
+        return [machine for machine in hypervisors_machines if machine.az in cli_azs]
 
     return hypervisors_machines
 
 
 async def _get_upgradable_hypervisors_machines(
     cli_force: bool, analysis_result: Analysis
-) -> set[Machine]:
+) -> list[Machine]:
     """Get the hypervisors that are possible to upgrade.
 
     :param cli_force: If force is used, it gets all hypervisors, otherwise just the empty ones
     :type cli_force: bool
     :param analysis_result: Analysis result
     :type analysis_result: Analysis
-    :return: Set of nova-compute units to upgrade
-    :rtype: set[Machine]
+    :return: List of nova-compute units to upgrade
+    :rtype: list[Machine]
     """
-    nova_compute_units = {
+    nova_compute_units = [
         unit
         for app in analysis_result.apps_data_plane
         for unit in app.units
         if app.charm == "nova-compute"
-    }
+    ]
 
     if cli_force:
-        return {unit.machine for unit in nova_compute_units}
+        return [unit.machine for unit in nova_compute_units]
 
-    return await get_empty_hypervisors(list(nova_compute_units), analysis_result.model)
+    return await get_empty_hypervisors(nova_compute_units, analysis_result.model)
 
 
 async def create_upgrade_group(
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -15,7 +15,7 @@
 """Upgrade planning utilities."""
 
 import logging
-from typing import Callable, Optional, cast
+from typing import Callable
 
 # NOTE we need to import the modules to register the charms with the register_application
 # decorator
@@ -176,7 +176,7 @@ def determine_upgrade_target(analysis_result: Analysis) -> OpenStackRelease:
 
 
 def verify_data_plane_cli_input(args: CLIargs, analysis_result: Analysis) -> None:
-    """Sane checks from the parameters passed in the cli to upgrade data-plane.
+    """Sanity checks from the parameters passed in the cli to upgrade data-plane.
 
     :param args: CLI arguments
     :type args: CLIargs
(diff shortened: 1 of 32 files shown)


cou/steps/plan.py
```diff
@@ -375,6 +379,65 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
+    :type analysis_result: Analysis
+    :return: Set of nova-compute units to upgrade
+    :rtype: set[Machine]
+    """
+    hypervisors_units = _get_hypervisors_units(analysis_result.apps_data_plane)
+
+    if cli_force:
+        return {unit.machine for unit in hypervisors_units}
+
+    hypervisors_units_name = [unit.name for unit in hypervisors_units]
```
agileshaw, line 425: nit: any reason we are using list instead of set here? We are using set everywhere else in this function so was wondering if I missed anything.

agileshaw (commented):
Generally LGTM. Just some non-blocking suggestions

tests/unit/steps/test_steps_plan.py
```diff
@@ -638,4 +582,219 @@ def test_verify_data_plane_cli_azs_raise_cannot_find():
...
+        (False, [0, 1], [0, 1]),
+        (False, [0], [0]),
+        (True, [0, 1, 2], [0, 1, 2]),
+        (True, [2], [2]),
+        (True, [0], [0]),
+    ],
+)
+@pytest.mark.asyncio
+@patch("cou.steps.plan._get_hypervisors_machines_possible_to_upgrade")
+async def test_filter_hypervisors_machines_by_machine_id(
```
rgildein, line 602: This test should test only the `filter_hypervisors_machines` by simulating all type of inputs from user, e.g. `--hosntame bla`, `--machine 1,2` and mock the `_get_hypervisors_machines_possible_to_upgrade` function.

Like that, we will actually test only the functionality of that single function instead of multiple function at once + we should avoid using Nova machines, you should be able to use simple mocks.

tests/unit/steps/test_steps_plan.py
```diff
@@ -638,4 +582,219 @@ def test_verify_data_plane_cli_azs_raise_cannot_find():
...
+        apps_machines["nova-compute"][NOVA_MACHINES[0]],
+        apps_machines["nova-compute"][NOVA_MACHINES[1]],
+    }
+    if force:
+        hypervisors_to_upgrade.add(apps_machines["nova-compute"][NOVA_MACHINES[2]])
+
+    return expected_hypervisors, hypervisors_to_upgrade
+
+
+def _prepare_cli_args(cli_args, machines, hostnames, az, force):
```
rgildein, line 754: This is not good behavior to change input argument, we should avoid usage of these and especially in tests, where it must be obvious what is going on.

rgildein (changes requested):

cou/steps/plan.py
gabrielcocenza: When I use the asyncio.gather the input needs to match with the exactly order of the inputs. Set are a little bit unpredictable the sequence, so using list I feel more safe from bugs.

What I'm trying to say that in a input like ["A", "B", "C"] in a function `foo(input)` I would like to be sure that the result of `foo(A)` won't be considered as output of `foo(B)` or `foo(C)`. Playing around with set it seems that they always return the same sequence when you interact, so you might be right that its not necessary.

agileshaw: Right, since we need to `zip` units and instance count together. Makes sense :ok_hand:

cou/steps/plan.py
```diff
@@ -327,6 +329,8 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     :rtype: UpgradePlan
     """
     pre_plan_sanity_checks(args, analysis_result)
+    hypervisors = await filter_hypervisors_machines(args, analysis_result)
+    logger.info("Hypervisors selected: %s", hypervisors)
```
rgildein, line 333: This should work only with hypervisors sub-command, or not?

cou/steps/plan.py
```diff
@@ -375,6 +379,55 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
+    """Get the hypervisors that are possible to upgrade.
+
+    :param cli_force: If force is used, it gets all hypervisors, otherwise just the empty ones
+    :type cli_force: bool
+    :param analysis_result: Analysis result
+    :type analysis_result: Analysis
+    :return: Set of nova-compute units to upgrade
+    :rtype: set[Machine]
+    """
+    nova_compute_units = {
```
rgildein, line 418: Why we are creating set and converting it to list in 428, I think we should do list here instead. The set does not bring any benefit, since units are unique.

tests/unit/conftest.py
```diff
@@ -683,3 +694,11 @@ def cli_args() -> MagicMock:
     """
     # spec_set needs an instantiated class to be strict with the fields.
     return MagicMock(spec_set=CLIargs(command="plan"))()
+
+
+def _generate_mock_machine(machine_id, hostname, az):
```
rgildein, line 699: Why this is "private" function? You are using it outside of conftest + this is helper function in tests, so there is not much reason to use "_".

tests/unit/utils/test_nova_compute.py
```diff
@@ -0,0 +1,87 @@
...
+    [
+        ([(0, 0), (1, 0), (2, 0)], {"0", "1", "2"}),
+        ([(0, 1), (1, 0), (2, 0)], {"1", "2"}),
+        ([(0, 1), (1, 3), (2, 0)], {"2"}),
+        ([(0, 1), (1, 3), (2, 5)], set()),
+    ],
+)
+@pytest.mark.asyncio
+@patch("cou.utils.nova_compute.get_instance_count")
+@patch("cou.utils.nova_compute.asyncio.gather", new_callable=AsyncMock)
```
rgildein, line 68: not-blocking: IMO, it would be cleaner to define `mock_instance_count.side_effect` and do not mock the gather here.

rgildein (changes requested):
Small changes required, but it looks good to me.

cou/steps/plan.py
gabrielcocenza: Not necessarily. The function abstracts the details and if nothing is provided, it will return all empty hypervisors and if force is used all hypervisors. Machines, hostnames and azs are just filters to create a sub-set of the hypervisors.

cou/steps/plan.py
gabrielcocenza: It was a way to be sure that we wouldn't have duplicate machines, but I think you are right. I'll change the data structure to be lists

rgildein (approved):
LGTM

cou/steps/plan.py
rgildein: If user use `cou upgrade controle-plane` we do not need to do this, so I think we should add one if condition to skip this in that case.

rgildein (changes requested):

cou/steps/plan.py
agileshaw: Good catch. I'm thinking it would be better if we create different functions, which will be called by `generate_plan`, for generating plans for control-plane, data-plane, and hypervisors. Then we can move this hypervisor filter to the one(s) needing it.

cou/steps/plan.py
```diff
@@ -375,6 +379,55 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
+    hypervisors_machines = await _get_upgradable_hypervisors_machines(args.force, analysis_result)
+
+    if cli_machines := args.machines:
+        return [machine for machine in hypervisors_machines if machine.machine_id in cli_machines]
+
+    if cli_hostnames := args.hostnames:
+        return [machine for machine in hypervisors_machines if machine.hostname in cli_hostnames]
+
+    if cli_azs := args.availability_zones:
+        return [machine for machine in hypervisors_machines if machine.az in cli_azs]
```
agileshaw, line 401: non-blocking: though this function is already pretty readable, it has similar code across these conditions which seems redundant. An idea for refactoring:
```python
cli_filters = [
        ('machine_id', args.machines),
        ('hostname', args.hostnames),
        ('az', args.availability_zones),
    ]
for attr, filter in cli_filters:
    if filter:
          return [machine for machine in hypervisors_machines if getattr(machine, attr) in filter]
```
This is also easily extendable if we need to add more filters in the future

rgildein: It's extendable only for machines filters, I do not think we will have much, and to me, it's less readable with more space for an issue.

cou/steps/plan.py
rgildein: No need, since this will be not used only if user call `cou upgrade controle-plane`. In other cases, we will use it.

cou/steps/plan.py
agileshaw: IMO this is less prone to error because you don't need to write this list comprehension multiple times `[machine for machine in hypervisors_machines if machine.az in cli_azs]` and have the chance to mistype something. It's also easier for reviews because they are using a unified way of iteration.

rgildein (approved):
Agreed that skipping hypervisors will be introduced with hypervisors sub-command.

cou/steps/plan.py
gabrielcocenza: We'll need to introduce the `hypervisor` command in a future PR and it's when we are going to fix this to not compute when using `control-plane`.

Because it's a dev branch, we decided to move forward to unblock the next PRs

agileshaw: Ack :ok_hand:

PR #231, opened 2024-02-01
Introduce pause and resume be able to upgrade OpenStack applications unit by unit

gabrielcocenza:
- enable action-managed-upgrade to be able to upgrade unit by unit
- pause OpenStack application unit
- resume OpenStack application unit

others:
- compare_step_coroutines able to compare two coroutines if the sequence of parameters respect the assignament of the function

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -34,6 +34,7 @@ from cou.steps import (
     ApplicationUpgradePlan,
     PostUpgradeStep,
     PreUpgradeStep,
+    UnitUpgradeStep,
     UpgradeStep,
 )
 from cou.utils.app_utils import upgrade_packages
@@ -641,30 +642,30 @@ class OpenStackApplication:
             coro=self.model.set_application_config(self.name, {"action-managed-upgrade": True}),
         )
 
-    def _get_pause_unit_step(self, unit: ApplicationUnit) -> UpgradeStep:
-        """Get the plan to pause a unit to upgrade.
+    def _get_pause_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
+        """Get the step to pause a unit to upgrade.
 
         :param unit: Unit to be paused.
         :type unit: ApplicationUnit
-        :return: Plan to pause a unit.
-        :rtype: UpgradeStep
+        :return: Step to pause a unit.
+        :rtype: UnitUpgradeStep
         """
-        return UpgradeStep(
+        return UnitUpgradeStep(
             description=f"Pause the unit: '{unit.name}'.",
             coro=self.model.run_action(
                 unit_name=unit.name, action_name="pause", raise_on_failure=True
             ),
         )
 
-    def _get_resume_unit_step(self, unit: ApplicationUnit) -> UpgradeStep:
+    def _get_resume_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
         """Get the step to resume a unit after upgrading the workload version.
 
         :param unit: Unit to be resumed.
         :type unit: ApplicationUnit
-        :return: Plan to resume a unit.
-        :rtype: UpgradeStep
+        :return: Step to resume a unit.
+        :rtype: UnitUpgradeStep
         """
-        return UpgradeStep(
+        return UnitUpgradeStep(
             description=(f"Resume the unit: '{unit.name}'."),
             coro=self.model.run_action(
                 unit_name=unit.name, action_name="resume", raise_on_failure=True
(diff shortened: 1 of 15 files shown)


cou/apps/base.py
```diff
@@ -624,6 +624,58 @@ def _get_disable_action_managed_plan(self, parallel: bool = False) -> UpgradeSte
...
+        :type unit: ApplicationUnit
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return: Plan to pause a unit.
+        :rtype: UpgradeStep
+        """
+        return UpgradeStep(
+            description=f"Pause the unit: '{unit.name}'.",
+            parallel=parallel,
+            coro=self.model.run_action(unit_name=unit.name, action_name="pause"),
```
agileshaw, line 660: Question: is "pause" and "resume" action idempotent?

cou/utils/app_utils.py
```diff
@@ -75,6 +75,30 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
...
+async def disable_nova_compute_scheduler(unit: str, model: COUModel) -> None:
+    """Disable nova-compute scheduler, so the unit cannot create new VMs.
+
+    :param unit: Name of the nova-compute unit where the action runs on.
+    :type unit: str
+    :param model: COUModel object
+    :type model: COUModel
+    """
+    action_name = "disable"
+    await model.run_action(unit_name=unit, action_name=action_name)
```
agileshaw, line 99: I don't understand why we can't add `model.run_action` as coro into an upgrade step of nova-compute application class (similar to how you are handling `pause` and `resume` to base application class). Any specific reason that I'm missing?

gabrielcocenza: I see your point. I was kind following the implementation of instance-count, but giving a second thought those actions probably won't be called outside the Nova-Compute class as it happens with instance-count.

I think it's ok leaving this way until the next PR for the nova-compute implementation. If I confirm this, I can move it to the Nova-Compute class. What do you think?

cou/apps/base.py
```diff
@@ -624,6 +624,58 @@ def _get_disable_action_managed_plan(self, parallel: bool = False) -> UpgradeSte
...
+        """Get the plan to pause a unit to upgrade.
+
+        :param unit: Unit to be paused.
+        :type unit: ApplicationUnit
+        :param parallel: Parallel running, defaults to False
+        :type parallel: bool, optional
+        :return: Plan to pause a unit.
+        :rtype: UpgradeStep
+        """
+        return UpgradeStep(
```
rgildein, line 657: We should check if this action exits or not? Or maybe define this only in application with such action.
For a situation like this, it would be nice to have mixing classes.

cou/utils/app_utils.py
rgildein: I agree, these functions are useless as utils, since they will be used only in the NovaCompute class.

rgildein (changes requested):

cou/apps/base.py
rgildein: IMO, this should be ok or if any issue happens, it should be responsibility of the charm itself.

cou/apps/base.py
rgildein: We will handle it later or maybe not care.

cou/utils/app_utils.py
gabrielcocenza: Removed from this PR. It's a simple function that can be introduced with NovaCompute class.

cou/apps/base.py
rgildein: @gabrielcocenza checked the nova-compute charm, and it looks like safe to do it. We will test the dev/data-plane before merging it to the main, so further test will happen.

cou/apps/base.py
```diff
@@ -640,40 +638,37 @@ def _get_enable_action_managed_plan(self, parallel: bool = False) -> UpgradeStep
             description=(
                 f"Change charm config of '{self.name}' 'action-managed-upgrade' to True."
             ),
-            parallel=parallel,
             coro=self.model.set_application_config(self.name, {"action-managed-upgrade": True}),
         )
 
-    def _get_pause_unit(self, unit: ApplicationUnit, parallel: bool = False) -> UpgradeStep:
+    def _get_pause_unit_step(self, unit: ApplicationUnit) -> UpgradeStep:
```
rgildein, line 644: We need to change this and following one to `UnitUpgradeStep`.

rgildein (changes requested):

rgildein (approved):

PR #232, opened 2024-02-02
Drop units populations from OpenStackApplication

rgildein:
Like, this we could make OpenStackApplication frozen, so it will represent static data in time.


cou/utils/juju_utils.py
```diff
@@ -526,3 +562,32 @@ async def get_model_machines(self) -> dict[str, Machine]:
...
+        app = await self._get_application(name)
+        full_status = await self.get_status()
+
+        # Note(rgildein): Retrieving units from state so the suboridnate application returns
+        # empty lists.
+        return [
+            Unit(
+                name=unit_name,
+                is_leader=unit.leader,
+                machine=Machine(
```
rgildein, line 584: Missing workload_version.

gabrielcocenza, line 584: Considering that the code should have a single way of doing one thing, I'm against creating machines while populating units because of some reasons.

1.  `get_model_machines` already does this job. If we take that approach, which I think we shouldn't, then we need to remove from the `juju_utils`.

2. This way we'll create the same machine more than once. In a hyper-converged  environment the same machine will have more than one App. So when you create a `Machine 0` for` app A`, it will create the same machine for `app B`. It's not a big efficiency problem, but why not access the already created machines using `O(1)` instead of `O(n)`.

3. I think we have an amazing opportunity in our hands to identify nova-compute machines and collocated apps with nova-compute. We are looping thru all applications to populate units and we already know the machines. If we have one new field in machine, `charms_deployed`,  we can have a new property on Machine class to identify hypervisors and collocated apps. E.g:

```python
class Machine:
    """Representation of a juju machine."""

    machine_id: str
    hostname: str
    az: Optional[str]  # simple deployments may not have azs
    charms_deployed: set[str] = field(default_factory=lambda: set())
    
    @property
    def is_hypervisor_machine(self) -> bool:
        return "nova-compute" in self.charms_deployed
```

With that we'll have an opportunity to have different treatment on `OpenStackApplication` depending if the machines are hypervisor or not to determine `current_os_release` without the need of instantiating Applications with just some units.

IMO it will be nice to ask nova-compute application or any collocated unit it's current OpenStack release and have the real knowledge about the application. Moreover, I think it's going to be easier to put the collocated apps with nova-compute on the `HypervisorUpgradePlanner`

gabrielcocenza (commented):
After some thoughts, I think I'm ok with the idea of the Application being responsible to know the `OpenStack version` from the application and units.

I liked the idea of populating the units outside the factory, but I don't completely agree with the way that you are proposing the Machine instantiation on the unit level. I think we can take a good advantage of that if we create the machines outside of the units and insert some application and/or unit knowledge into the machines

rgildein:
replaced by #242

PR #234, opened 2024-02-06
Use UnitUpgradeStep for packages upgrade

rgildein:
Like this, we can upgrade packages in individual step for each unit and run these steps in parallel or sequentially. Right now, it's run sequentially, so no change in previous approach.


cou/utils/app_utils.py
```diff
@@ -27,13 +26,11 @@
 logger = logging.getLogger(__name__)
 
 
-async def upgrade_packages(
-    units: Iterable[str], model: COUModel, packages_to_hold: Optional[list]
-) -> None:
+async def upgrade_packages(unit: str, model: COUModel, packages_to_hold: Optional[list]) -> None:
     """Run package updates and upgrades on each unit of an Application.
```
agileshaw, line 30:
```suggestion
    """Run package updates and upgrades on an Application unit.
```

cou/apps/base.py
```diff
@@ -513,13 +514,21 @@ def _get_upgrade_current_release_packages_plan(self, parallel: bool = False) ->
         :return: Plan for upgrading software packages to the latest of the current release.
         :rtype: PreUpgradeStep
         """
-        return PreUpgradeStep(
+        step = PreUpgradeStep(
```
agileshaw, line 517: Should we change this to a plan instead of step? It have empty coro and allows its substeps to be run in parallel. I'm pointing this out because in #235 we are removing all `parallel` param in step functions, and we've been trying to draw a clear line between plan and step.

tests/unit/apps/test_auxiliary.py
```diff
@@ -109,14 +110,21 @@ def test_auxiliary_upgrade_plan_ussuri_to_victoria_change_channel(
...
+
     upgrade_steps = [
-        PreUpgradeStep(
-            description=(
-                f"Upgrade software packages of '{app.name}' from the current APT repositories"
-            ),
-            parallel=False,
-            coro=app_utils.upgrade_packages(app.status.units.keys(), model, None),
-        ),
+        expected_upgrade_package_step,
```
agileshaw, line 127: These tests is getting more and more complicated because of the multi-level steps structure. We should consider refactoring it

agileshaw (changes requested):

tests/unit/apps/test_auxiliary.py
rgildein: Totally agree, these tests are not even useful for actual testing.

cou/apps/base.py
```diff
@@ -513,13 +514,21 @@ def _get_upgrade_current_release_packages_plan(self, parallel: bool = False) ->
...
             description=(
                 f"Upgrade software packages of '{self.name}' from the current APT repositories"
             ),
             parallel=parallel,
```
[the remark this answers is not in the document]
gabrielcocenza: I was to say that I think this should be True by default. I don't see a problem too.

gabrielcocenza (commented):
Changes looks good. I'm just wondering if those updates should run by default on parallel

cou/apps/base.py
rgildein: After second thought, I do not see reason to make it as plan, since this is a sub-step of application plan. I agree that it does not have coro, but we want to make it parallel = True and plan do not have such option. Do you think we should add PreUpgradePlan? I think that it's not needed.

rgildein:
closing in favor of PR against main branch

PR #237, opened 2024-02-07
Feature/bseng 1937/upgrade unit by unit

rgildein:
Add UnitUpgradeStep and is_unit_upgrade_step, also use sub_steps as
property to prevent defining it as not list of BaseStep's. Raising
TypeError in add_step if object is not based on BaseStep.

Use UnitUpgradeStep, we can upgrade packages in individual step for each unit and
run these steps in parallel or sequentially. Right now, it's run
sequentially, so no change in previous approach.


cou/steps/__init__.py
```diff
@@ -286,9 +309,24 @@ class UpgradeStep(BaseStep):
...
+
 class PreUpgradeStep(UpgradeStep):
     """Represents the pre-upgrade step."""
 
 
 class PostUpgradeStep(UpgradeStep):
     """Represents the post-upgrade step."""
+
+
+def is_unit_upgrade_step(step: BaseStep) -> bool:
```
agileshaw, line 324: Question: what is the intention for adding this check? I don't see this being used yet so was curious.

agileshaw (commented):
Looks good, and glad that we are updating documentation to align. I just have one question.

tests/unit/apps/test_auxiliary.py
```diff
@@ -670,4 +712,5 @@ def test_mysql_innodb_cluster_upgrade(status, config, model):
     ]
     add_steps(expected_plan, upgrade_steps)
 
+    print(upgrade_plan, expected_plan)
```
gabrielcocenza, line 715: I guess you forgot to remove it...

cou/steps/__init__.py
gabrielcocenza: Same question

rgildein: I used cherry-pick and forgot about it. I removed it.

gabrielcocenza (approved):
LGTM

agileshaw (approved):
LGTM

PR #239, opened 2024-02-07
Introduction of NovaCompute class

gabrielcocenza:
- upgrade unit by unit
- change in the base to accept list of ApplicationUnit to generate plan
- disable or enable action-managed-upgrade depending if units are passed
- HaltUpgradeExecution to not run the sub_steps of a BaseStep
- force argument into generate_upgrade_plan


cou/steps/execute.py
```diff
@@ -36,13 +37,17 @@ async def _run_step(step: BaseStep, prompt: bool, overwrite_progress: bool = Fal
...
+        if isinstance(step, UpgradeStep):
+            progress_indicator.start(step.description)
+            await step.run()
+            if not overwrite_progress:
+                progress_indicator.succeed()
+        else:
+            await step.run()
+    except HaltUpgradeExecution:
+        progress_indicator.fail()
+        step.sub_steps = []
```
rgildein, line 50: We should not change a sub-step of any step in the plan during execution, that's not right. You can cancel them, but I do not think that it's good to have sub-steps for `UnitUpgradeStep(description="Run the instance-count to upgrade", ...)`. I'm ok to handle it in execution, but I will expect that it works like `continue`.

rgildein (changes requested):

cou/apps/core.py
```diff
@@ -40,3 +44,106 @@ class Octavia(OpenStackApplication):
...
+                self._get_openstack_upgrade_step(unit),
+                self._get_resume_unit_step(unit),
+            ]
+
+            if empty_hypervisor_check_step:
+                [empty_hypervisor_check_step.add_step(step) for step in other_steps]
+                unit_steps.add_step(empty_hypervisor_check_step)
+            else:
+                [unit_steps.add_step(step) for step in other_steps]
+
```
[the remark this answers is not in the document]
gabrielcocenza: Those steps are going to run if the empty hypervisor step pass the check. This gives a clear dependency hierarchy and if it fails, the next step is to enable the nova-compute scheduler.

cou/steps/execute.py
rgildein: From our discussion:
```python
try:
    await apply_step(sub_step, prompt, overwrite_substeps_progress)
except COUBreakException:
    progress_indicator.failed()
    break
```

cou/utils/nova_compute.py
```diff
@@ -60,3 +64,17 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
         f"No valid instance count value found in the result of {action_name} action "
         f"running on '{unit}': {action.results}"
     )
+
+
+async def get_instance_count_to_upgrade(unit: ApplicationUnit, model: COUModel) -> None:
+    unit_instance_count = await get_instance_count(unit, model)
+    if unit_instance_count != 0:
+        model.run_action(unit_name=unit.name, action_name="enable", raise_on_failure=True)
```
rgildein, line 72: Maybe we will need to try catch this and if that failed, send `SIGINT` or `SIGTERM`. But maybe it's not needed, and it's safe to continue even if it failed.

cou/steps/execute.py
```diff
@@ -59,7 +60,20 @@ async def _run_step(step: BaseStep, prompt: bool, overwrite_progress: bool = Fal
...
+                await apply_step(sub_step, prompt, overwrite_substeps_progress)
+            except HaltUpgradeExecution:
+                sub_step_index = step.sub_steps.index(sub_step)
+                non_executed_sub_steps = [
+                    sub_step.description for sub_step in step.sub_steps[sub_step_index + 1 :]
+                ]
+                logger.warning(
+                    "Step: '%s' failed to complete execution. Steps '%s' will be skipped.",
+                    sub_step.description,
+                    ", ".join(non_executed_sub_steps),
```
rgildein, line 73: not-blocking: I think that this will not be readable for user, and we should use new line, but this can be improved during manual test phase.

cou/utils/nova_compute.py
```diff
@@ -60,3 +64,25 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
         f"No valid instance count value found in the result of {action_name} action "
         f"running on '{unit}': {action.results}"
     )
+
+
+async def get_instance_count_to_upgrade(unit: ApplicationUnit, model: COUModel) -> None:
```
rgildein, line 69: This name is really strange, we should change it.

cou/utils/nova_compute.py
```diff
@@ -60,3 +64,25 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
...
+
+    :param unit: Unit to check if there are VMs running.
+    :type unit: ApplicationUnit
+    :param model: COUModel
+    :type model: COUModel
+    :raises HaltUpgradeExecution: When a unit has VMs running.
+    """
+    unit_instance_count = await get_instance_count(unit.name, model)
+    if unit_instance_count != 0:
+        await model.run_action(unit_name=unit.name, action_name="enable", raise_on_failure=True)
```
rgildein, line 80: I thought we agreed that this needs to be caught and log.error message should be provided in that situation.

cou/apps/core.py
```diff
@@ -40,3 +45,117 @@ class Octavia(OpenStackApplication):
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade steps,
+        :type units: Optional[list[ApplicationUnit]]
+        :raises ApplicationError: If no units are passed to upgrade Nova Compute
+        :return: List of upgrade steps.
+        :rtype: list[UpgradeStep]
+        """
+        if not units:
+            raise ApplicationError("No units passed to upgrade Nova Compute.")
```
rgildein, line 75: This should not be here, it should be possible to upgrade all units, when we do not pass list of units.

cou/apps/core.py
```diff
@@ -40,3 +45,117 @@ class Octavia(OpenStackApplication):
...
+            parallel=True,
+        )
+
+        for unit in units:
+            unit_steps = UnitUpgradeStep(description=f"Upgrade plan for unit: {unit.name}")
+            unit_steps.add_step(self._get_disable_scheduler_step(unit))
+            unit_steps.add_step(self._get_empty_hypervisor_step(unit))
+            unit_steps.add_step(self._get_pause_unit_step(unit, self._dependency_of_steps))
+            unit_steps.add_step(self._get_openstack_upgrade_step(unit, self._dependency_of_steps))
+            unit_steps.add_step(self._get_resume_unit_step(unit, self._dependency_of_steps))
```
rgildein, line 99:
```suggestion
            unit_steps.add_step(self._get_pause_unit_step(unit, not self.force))
            unit_steps.add_step(self._get_openstack_upgrade_step(unit,  not self.force))
            unit_steps.add_step(self._get_resume_unit_step(unit,  not self.force))
```
And even for force I start to have feeling, that it should be argument for `upgrade_steps` and not class argument.

rgildein (changes requested):

cou/apps/core.py
gabrielcocenza: I agree that this behavior should happen for OpenStack Application, but not for NovaCompute. If no units are passed to NovaCompute is a strange situation. Maybe that means that it has no units to upgrade for a certain target.

cou/apps/core.py
gabrielcocenza: The problem is that `force` is very specific for `NovaCompute` and the `upgrade_steps` is inherited from `OpenStackApplication`. Adding a new parameter for just one class does not looks right to me.

IMO `not self.force` is very out of context. Having a property adds a little overhead, but it's more expressive.

rgildein, line 99: If we move force as argument for `generate_upgrade_plan` instead of argument for whole application, we can use something like this:

```suggestion
            if force is True:
                unit_steps.add_step(self._get_empty_hypervisor_step(unit))

            is_dependent = not force
            unit_steps.add_step(self._get_pause_unit_step(unit, is_dependent))
            unit_steps.add_step(self._get_openstack_upgrade_step(unit, is_dependent))
            unit_steps.add_step(self._get_resume_unit_step(unit, is_dependent))
```

cou/steps/execute.py
```diff
@@ -59,7 +60,24 @@ async def _run_step(step: BaseStep, prompt: bool, overwrite_progress: bool = Fal
...
+            except HaltUpgradeExecution:
+                sub_step_index = step.sub_steps.index(sub_step)
+                non_executed_sub_steps = [
+                    sub_step.description for sub_step in step.sub_steps[sub_step_index + 1 :]
+                ]
+                logger.warning(
+                    (
+                        "Step: '%s' from '%s' failed to complete execution. "
+                        "The following steps will be skipped:\n %s"
+                    ),
```
rgildein, line 74: not-blocking: The brackets are not needed.

cou/utils/nova_compute.py
```diff
@@ -59,3 +63,34 @@ async def get_instance_count(unit: str, model: COUModel) -> int:
...
+                unit_name=unit.name, action_name="enable", raise_on_failure=True
+            )
+            logger.info("Successfully enabled nova-compute scheduler from unit %s.", unit.name)
+        except ActionFailed:
+            logger.error("Failed to enable nova-compute scheduler from unit %s.", unit.name)
+        logger.warning(
+            (
+                "Unit: %s has %s VMs running. The upgrade on this unit cannot happen "
+                "unless it's empty or force flag is used"
+            ),
```
rgildein, line 92: not-blocking: same here

rgildein (approved):
LGTM, small comments.

tests/unit/apps/test_core.py
```diff
@@ -842,17 +848,17 @@ def test_nova_compute_get_disable_scheduler_step(model):
...
-    units = [
-        ApplicationUnit(f"nova-compute/{unit_num}", MagicMock(), MagicMock(), MagicMock())
+    units = {
+        f"nova-compute/{unit_num}": COUUnit(f"nova-compute/{unit_num}", MagicMock(), MagicMock())
         for unit_num in range(3)
-    ]
-    app = NovaCompute(app_name, status, {}, model, charm, {})
+    }
+    app = NovaCompute(
+        app_name, "", charm, channel, {}, {}, model, "cs", "focal", [], units, "21.0.1"
```
rgildein, line 861: Should we use `ch`?

rgildein (approved):
LGTM

cou/steps/execute.py
```diff
@@ -59,7 +60,24 @@ async def _run_step(step: BaseStep, prompt: bool, overwrite_progress: bool = Fal
         logger.debug("running all sub-steps of %s step sequentially", step)
         for sub_step in step.sub_steps:
             logger.debug("running sub-step %s of %s step", sub_step, step)
-            await apply_step(sub_step, prompt, overwrite_substeps_progress)
+            try:
+                await apply_step(sub_step, prompt, overwrite_substeps_progress)
+            except HaltUpgradeExecution:
```
agileshaw, line 65: IIUC based on the execution logic here, if instance check on vm failed, none of the steps in the same level will be executed, including `_get_enable_scheduler_step()` even though it doesn't have the `dependent` parameter. I don't think this is the correct behavior.

rgildein: The step raining `HaltUpgradeExecution` will handle the enabling itself. The dependent is only to visually improve UX.
With saying `progress_indicator.fail()` you will see something like this:
```bash
# pritning plan
...
Upgrade plan for unit: nova-compute/2
    Disable nova-compute scheduler from unit: 'nova-compute/2'.
    Check if unit nova-compute/2 has no VMs running before upgrading.
    ├── Pause the unit: 'nova-compute/2'.
    ├── Upgrade the unit: 'nova-compute/2'.
    ├── Resume the unit: 'nova-compute/2'.
    Enable nova-compute scheduler from unit: 'nova-compute/2'.
...
# executing plan
Upgrade plan for unit: nova-compute/1 ✔
Upgrade plan for unit: nova-compute/2 ✖
Upgrade plan for unit: nova-compute/3 ...
...
```

agileshaw: Yes, but in the log, it will say that `Enable nova-compute scheduler from unit: 'nova-compute/2'.` step is skipped, right? That's is giving the user a false sense that nova-compute scheduler is still disabled.

agileshaw: Wouldn't it be better if we don't enable the scheduler in "Check if unit nova-compute/2 has no VMs running before upgrading." but actually run  `Enable nova-compute scheduler from unit: 'nova-compute/2'` step?

gabrielcocenza: I guess we'll need to get a little dipper into this in another PR. For this one I'm removing the independent steps like `Enable nova-compute scheduler from unit: 'nova-compute/2'.` from the warning log so the user get less confuse about if the scheduler is enable or not.

I still think that it would be an easier implementation if the "dependency" was  as substeps of the step that can Halt. We can talk about this later

agileshaw: I also think it would be better expressing dependencies with one more level of sub-steps. The overall logic would require more thorough design, especially considering the problem I pointed out in the other comment. 

For this PR, improving the warning message is enough.

agileshaw (approved):
Approving this PR as it serves a good starting point of nova-compute upgrade. The remaining issues can be addressed in future PRs.

PR #240, opened 2024-02-08
Introduced COUApplication, COUUnit and COUMachine

rgildein:
Add COUApplication, COUUnit and COUMachine to juju_utils and create ger_applications function, which is obtaining all application from model with all information required for COU.


cou/utils/juju_utils.py
```diff
@@ -266,6 +313,38 @@ async def connect(self) -> None:
...
+        # note(rgildein): We get the applications from the Juju status, since we can get more
+        #                 information the status than from objects. e.g. workload_version for unit
+        full_status = await self.get_status()
+        machines = await self.get_model_machines()
+
+        return {
+            app: COUApplication(
+                name=app,
+                can_upgrade_to=status.can_upgrade_to,
+                charm=status.charm,
```
rgildein, line 331: This is not what we need, value like this look like `'ch:amd64/focal/ubuntu-24'` and it should be just `ubuntu`.

cou/utils/juju_utils.py
```diff
@@ -266,6 +313,38 @@ async def connect(self) -> None:
...
+        """Return list of applications with all relevant information.
+
+        :returns: list of application with all information
+        :rtype: list[COUApplication]
+        """
+        # note(rgildein): We get the applications from the Juju status, since we can get more
+        #                 information the status than from objects. e.g. workload_version for unit
+        full_status = await self.get_status()
+        machines = await self.get_model_machines()
+
```
gabrielcocenza, line 326: I think you can also get the applications like: `model.applications` and with that access the charm_name and the config without the need of using `self.get_application_config`

gabrielcocenza (changes requested):
Small details, almost ready for merging

cou/utils/juju_utils.py
```diff
@@ -266,6 +320,39 @@ async def connect(self) -> None:
         )
 
     @retry
+    async def get_applications(self) -> dict[str, COUApplication]:
```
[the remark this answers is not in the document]
rgildein: It's COUModel, saying that it returns cou_application is redundant.

cou/utils/juju_utils.py
```diff
@@ -266,6 +320,39 @@ async def connect(self) -> None:
...
+            app: COUApplication(
+                name=app,
+                can_upgrade_to=status.can_upgrade_to,
+                charm=model.applications[app].charm_name,
+                charm_channel=status.charm_channel,
+                config=await model.applications[app].get_config(),
+                machines={unit.machine: machines[unit.machine] for unit in status.units.values()},
+                model=self,
+                series=status.series,
+                subordinate_to=status.subordinate_to,
```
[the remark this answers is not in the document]
rgildein: We are not using it, that's true, but I see benefit leaving it there. I'll move the property `is_subordinate` to the COUApplication, what do you say?

cou/utils/juju_utils.py
```diff
@@ -131,6 +133,67 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:  # pylint: disable=W9011
...
+
+
+@dataclass(frozen=True)
+class COUApplication:
+    """Representation of an single application."""
+
+    # pylint: disable=too-many-instance-attributes
+
+    name: str
+    can_upgrade_to: str
```
agileshaw, line 177: Question: why is this attribute needed?

agileshaw (commented):
Overall looks good. Two small comments inline.

cou/utils/juju_utils.py
rgildein: It is defined in `OpenStackApplication`, but never used. So it's not needed.

rgildein: Should I remove it?

agileshaw: Just checked, it is being used to determine whether we are at the latest revision of the current channel ([line](https://github.com/canonical/charmed-openstack-upgrader/blob/bc6fe5d89f784d0a8e57056fea9e1297bd9fb8d3/cou/apps/base.py#L392)). So we should leave it.

agileshaw (approved):
LGTM

cou/utils/juju_utils.py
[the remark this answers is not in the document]
rgildein: I removed it in the #242, where the real implementation happen.

PR #242, opened 2024-02-09
feature: using COUApplication as base for OpenStackApplication

rgildein:
Like this, we moved usage of libjuju even more to juju_utils + now OpenStackApplication are frozen static object, which are initialized without any post initialization steps.

This is based on #240


cou/utils/juju_utils.py
```diff
@@ -131,6 +133,67 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:  # pylint: disable=W9011
     return _wrapper
 
 
+@dataclass(frozen=True)
+class COUMachine:
```
aieri, line 137: nit: does the `COU` prefix mean much here? These classes seem quite generic

cou/apps/base.py
```diff
@@ -643,7 +580,7 @@ def _get_enable_action_managed_step(self) -> UpgradeStep:
             coro=self.model.set_application_config(self.name, {"action-managed-upgrade": True}),
         )
 
-    def _get_pause_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
+    def _get_pause_unit_step(self, unit: COUUnit) -> UnitUpgradeStep:
```
gabrielcocenza, line 583: Update docstrings

cou/apps/base.py
```diff
@@ -658,7 +595,7 @@ def _get_pause_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
             ),
         )
 
-    def _get_resume_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
+    def _get_resume_unit_step(self, unit: COUUnit) -> UnitUpgradeStep:
```
gabrielcocenza, line 598: Update docstrings

cou/apps/base.py
```diff
@@ -673,7 +610,7 @@ def _get_resume_unit_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
             ),
         )
 
-    def _get_openstack_upgrade_step(self, unit: ApplicationUnit) -> UnitUpgradeStep:
+    def _get_openstack_upgrade_step(self, unit: COUUnit) -> UnitUpgradeStep:
```
gabrielcocenza, line 613: Update docstrings

gabrielcocenza (changes requested):
Just some small typos

cou/utils/juju_utils.py
rgildein: No, it's just to differentiate from juju Machine, Application and Unit. I do not think that it's necessary, but I followed the same approach as we did with COUModel.

aieri (approved):
This looks good, I just feel I'm missing the big picture. We can discuss it next ime

cou/apps/base.py
```diff
@@ -92,31 +61,21 @@ class OpenStackApplication:
...
     def __hash__(self) -> int:
         """Hash magic method for Application.
 
         :return: Unique hash identifier for Application object.
         :rtype: int
         """
-        return hash(f"{self.name}{self.charm}")
+        return hash(f"{self.name}({self.charm})")
 
     def __eq__(self, other: Any) -> bool:
```
gabrielcocenza, line 80: is it still necessary to have `hash` and `eq` if now the dataclass is frozen?

cou/apps/factory.py
```diff
@@ -34,48 +32,37 @@ class AppFactory:
...
+        if is_charm_supported(app.charm):
+            app_class = cls.charms.get(app.charm, OpenStackApplication)
             return app_class(
-                name=name,
-                status=status,
-                config=config,
-                model=model,
-                charm=charm,
-                machines=machines,
+                name=app.name,
```
gabrielcocenza, line 49: I think you can use:
```python
from dataclasses import asdict
return app_class(**asdict(app))
```

tests/unit/apps/test_auxiliary.py
```diff
@@ -30,91 +31,113 @@
...
+    """
+    machines = {"0": MagicMock(spec_set=COUMachine)}
     app = RabbitMQServer(
-        "rabbitmq-server",
-        status["rabbitmq_server"],
-        config["auxiliary_ussuri"],
-        model,
-        "rabbitmq-server",
-        apps_machines["rmq"],
+        name="rabbitmq-server",
```
gabrielcocenza, line 48: I'm against creating a application instance every time to test something. It's very repetitive and does not make part of the test itself, so I think it should be moved somehow to conftest.

I know that the current unit tests are not ideal, but this can also add pain for the future. If a new field is addes/removed this would result into a lot of broken unit tests and the changes will be spread instead of changing on a single place

cou/apps/factory.py
gabrielcocenza: From the sync it was said that the problem was because asdict creates a deepcopy and this can be problematic to model. Maybe this can be an alternative for shallow copy

```python
return app_class(**dict((field.name, getattr(obj, field.name)) for field in fields(app)))
```

tests/unit/apps/test_auxiliary.py
```diff
@@ -30,91 +31,113 @@
...
-            os_version=OpenStackRelease("yoga"),
-            workload_version="3.8",
-            machine=apps_machines["rmq"]["0/lxd/19"],
-        )
-    ]
-    rmq_status = status["rabbitmq_server"]
-    rmq_status.charm = "cs:amd64/focal/rabbitmq-server-638"
-    rmq_status.charm_channel = "stable"
+def test_auxiliary_app_cs(model):
+    """Test aucilary application from charm store."""
```
aieri, line 77:
```suggestion
    """Test auxiliary application from charm store."""
```

cou/apps/base.py
rgildein: Unfortunately, they are needed, since class is frozen.

```bash
Full execution log: '/home/ubuntu/.local/share/cou/log/cou-20240215220116.log'  
Connected to 'zaza-ce1e95799ba2' ✔
Analyzing cloud... \2024-02-15 22:01:16 [ERROR] Unexpected error occurred.
2024-02-15 22:01:16 [ERROR] unhashable type: 'dict'
```

gabrielcocenza (approved):
LGTM

agileshaw (commented):
LGTM

PR #243, opened 2024-02-15
integrate the force flag from cli with the generate_upgrade_plan

gabrielcocenza:
- update unit tests

Should be merged after #239


agileshaw (approved):
LGTM

cou/apps/base.py
```diff
@@ -411,19 +411,19 @@ def post_upgrade_steps(self, target: OpenStackRelease) -> list[PostUpgradeStep]:
     def generate_upgrade_plan(
         self,
         target: OpenStackRelease,
+        force: bool,
```
rgildein, line 414: Can we use default value to `False`?

rgildein (approved):
One question, but that's not blocking one.

cou/apps/base.py
gabrielcocenza: We can, but I prefer not to so it forces passing the CLI parameter for it.

rgildein: Why? What is the benefit of such approach?

rgildein: Discussed and we will change it latter, if we do not forget.

PR #245, opened 2024-02-15
Upgrade software packages of individual units instead of all units.

gabrielcocenza:
- pre_upgrade_steps accepts units as parameters
- if no units is passed, the pre-upgrade will upgrade the packages of all units
- if units are passed, upgrade the packages of those units.

PR should be merged after #239


cou/utils/nova_compute.py
```diff
@@ -22,6 +22,8 @@
 
 logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
```
agileshaw, line 25: Duplicated logger

agileshaw (changes requested):
Overall LGTM. Just one duplicated line coming from merge.

PR #246, opened 2024-02-16
Implement AZ grouping mechanism

rgildein:
Add AZ grouping mechanism for `HypervisorUpgradePlanner` and add also showcase for generating upgrade steps.

### DIFF
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -14,11 +14,11 @@
 
 """Hypervisor planner."""
 import logging
+from collections import defaultdict
 from dataclasses import dataclass
 from typing import Any
 
 from cou.apps.base import OpenStackApplication
-from cou.exceptions import ApplicationError
 from cou.steps import (
     HypervisorUpgradePlan,
     PostUpgradeStep,
@@ -45,11 +45,35 @@ class HypervisorGroup:
         """Equal magic method for HypervisorGroup.
 
         :param other: HypervisorGroup object to compare.
-        :type other: any
+        :type other: Any
         :return: True if equal False if different.
         :rtype: bool
         """
+        if not isinstance(other, HypervisorGroup):
+            return NotImplemented
+
         return other.name == self.name
+
+
+class AZs(defaultdict):
+    """AZs dictionary object with default value HypervisorGroup."""
+
+    def __init__(self) -> None:
+        """Initialize the Hypervisor class.
+
+        The AZs represent default dict, with predefined HypervisorGroup as default value.
+        """
+        super().__init__(default_factory=None)
+
+    def __missing__(self, key: str) -> HypervisorGroup:
+        """Handle missing key in AZs.
+
+        If key is missing in the AZs dict, the default value
+        HypervisorGroup(name=key, apps=defaultdict(list) will be used. Like this we can always
+        access the az["my-az"].name == "my-az" or az["my-az"].apps["my-app"].
+        """
+        self[key] = HypervisorGroup(name=key, apps=defaultdict(list))
+        return self[key]
 
 
 def verify_apps(apps: list[OpenStackApplication]) -> None:
@@ -90,7 +114,7 @@ class HypervisorUpgradePlanner:
         return self._apps
 
     @property
-    def azs(self) -> dict[str, HypervisorGroup]:
+    def azs(self) -> AZs:
         """Returns a list of AZs defined in individual applications.
 
         Each AZ contains a dictionary of application name and all units in the AZ.
@@ -115,24 +139,12 @@ class HypervisorUpgradePlanner:
         :rtype: dict[str, HypervisorGroup]
         :raises ApplicationError: if there is unit without az defined
         """
-        azs = {}
+        azs = AZs()
         for app in self.apps:
             for unit in app.units.values():
-                az = unit.machine.az
-
-                if az is None:
-                    raise ApplicationError(
-                        f"The application {app.name} has a machine {unit.machine} without az"
-                    )
-
-                if az not in azs:
-                    # define empty hypervisor group for az
-                    azs[az] = HypervisorGroup(name=az, apps={})
-
-                if app.name not in azs[az].apps:
-                    # define empty list of units for each app, so unit can be added
-                    azs[az].apps[app.name] = []
-
+                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
+                #                 belong to a single group.
+                az = unit.machine.az or ""
                 azs[az].apps[app.name].append(unit)
 
         return azs
@@ -150,7 +162,7 @@ class HypervisorUpgradePlanner:
     def _generate_hypervisor_group_upgrade_plan(
         self, target: OpenStackRelease, group: HypervisorGroup
     ) -> HypervisorUpgradePlan:
-        """Genarete upgrade plan for a single hypervisor group.
+        """Generate upgrade plan for a single hypervisor group.
 
         :param target: OpenStack codename to upgrade.
         :type target: OpenStackRelease
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -39,7 +39,7 @@ class HypervisorGroup:
     """
 
     name: str
-    apps: dict[str, list[COUUnit]]
+    app_units: dict[str, list[COUUnit]]
 
     def __eq__(self, other: Any) -> bool:
         """Equal magic method for HypervisorGroup.
@@ -72,7 +72,7 @@ class AZs(defaultdict):
         HypervisorGroup(name=key, apps=defaultdict(list) will be used. Like this we can always
         access the az["my-az"].name == "my-az" or az["my-az"].apps["my-app"].
         """
-        self[key] = HypervisorGroup(name=key, apps=defaultdict(list))
+        self[key] = HypervisorGroup(name=key, app_units=defaultdict(list))
         return self[key]
 
 
@@ -101,7 +101,6 @@ class HypervisorUpgradePlanner:
         :param apps: sorted list of OpenStack applications
         :type apps: list[OpenStackApplication]
         """
-        # verify_apps(apps)
         self._apps = apps
 
     @property
@@ -145,7 +144,7 @@ class HypervisorUpgradePlanner:
                 # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
                 #                 belong to a single group.
                 az = unit.machine.az or ""
-                azs[az].apps[app.name].append(unit)
+                azs[az].app_units[app.name].append(unit)
 
         return azs
 
@@ -173,7 +172,7 @@ class HypervisorUpgradePlanner:
         """
         plan = HypervisorUpgradePlan(description=f"Upgrade plan for '{group.name}' to {target}")
         for app in self.apps:
-            if app.name not in group.apps:
+            if app.name not in group.app_units:
                 logger.debug(
                     "skipping application %s because it is not part of group %s",
                     app.name,
@@ -181,7 +180,7 @@ class HypervisorUpgradePlanner:
                 )
                 continue
 
-            units = group.apps[app.name]
+            units = group.app_units[app.name]
             logger.info("generating upgrade steps for %s app and %s units", app.name, units)
             plan.sub_steps = app.upgrade_steps(target, units)  # type: ignore[call-arg, assignment]
 
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -15,8 +15,10 @@
 """Hypervisor planner."""
 import logging
 from dataclasses import dataclass
+from typing import Any
 
 from cou.apps.base import OpenStackApplication
+from cou.exceptions import ApplicationError
 from cou.steps import (
     HypervisorUpgradePlan,
     PostUpgradeStep,
@@ -29,15 +31,7 @@ from cou.utils.openstack import OpenStackRelease
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
-class HypervisorMachine:
-    """Hypervisor machine containing units for multiple applications."""
-
-    name: str
-    units: list[COUUnit]
-
-
-@dataclass(frozen=True)
+@dataclass
 class HypervisorGroup:
     """Group of hypervisors.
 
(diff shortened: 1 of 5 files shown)


cou/steps/hypervisor.py
```diff
@@ -84,25 +90,52 @@ def apps(self) -> list[OpenStackApplication]:
...
+                        f"The application {app.name} has a machine {unit.machine} without az"
+                    )
+
+                if az not in azs:
+                    # define empty hypervisor group for az
+                    azs[az] = HypervisorGroup(name=az, apps={})
+
+                if app.name not in azs[az].apps:
+                    # define empty list of units for each app, so unit can be added
+                    azs[az].apps[app.name] = []
```
agileshaw, line 134: Would be simpler if we do:
```suggestion
                if az not in azs:
                    # define empty hypervisor group for az
                    azs[az] = HypervisorGroup(name=az, apps={app.name: []})
                elif app.name not in azs[az].apps:
                    # define empty list of units for each app, so unit can be added
                    azs[az].apps[app.name] = []
```

cou/steps/hypervisor.py
```diff
@@ -114,20 +147,33 @@ def _generate_pre_upgrade_plan(self) -> PreUpgradeStep:
...
-    def _generate_hypervisor_upgrade_plan(
-        self, hypervisor: HypervisorMachine
+    def _generate_hypervisor_group_upgrade_plan(
+        self, target: OpenStackRelease, group: HypervisorGroup
     ) -> HypervisorUpgradePlan:
-        """Genarete upgrade plan for a single hypervisor.
-
-        Each hypervisor upgrade consists of a UnitUpgradeStep, so all substeps should be based
-        on UnitUpgradeStep.
+        """Genarete upgrade plan for a single hypervisor group.
```
agileshaw, line 153:
```suggestion
        """Generate upgrade plan for a single hypervisor group.
```

cou/steps/hypervisor.py
```diff
@@ -29,23 +31,25 @@
...
+
+    def __eq__(self, other: Any) -> bool:
+        """Equal magic method for HypervisorGroup.
+
+        :param other: HypervisorGroup object to compare.
+        :type other: any
+        :return: True if equal False if different.
+        :rtype: bool
+        """
+        return other.name == self.name
```
agileshaw, line 52: If we are accepting `Any` type for comparison, we should consider the case where that object doesn't have `name` attribute. The current code would raise un-catched error.

agileshaw (changes requested):

cou/steps/hypervisor.py
rgildein: That will not work, since with first if you created whole ` azs[az]` and in second you add empty list for app. 

We can do it, but like this:
```python
if az not in azs:
  # define empty hypervisor group for az
  azs[az] = HypervisorGroup(name=az, apps={app.name: []})
  azs[az].apps[app.name] = []
elif app.name not in azs[az].apps:
  # define empty list of units for each app, so unit can be added
  azs[az].apps[app.name] = []
```
and there is no benefit doing it like that.

cou/steps/hypervisor.py
rgildein: Good point, I copy and paste this from other class, so I will fix it there also.

cou/steps/hypervisor.py
```diff
@@ -84,25 +90,52 @@ def apps(self) -> list[OpenStackApplication]:
...
-        :rtype: list[HypervisorGroup]
+        :return: dictionary with key as az name and value as HypervisorGroup
+        :rtype: dict[str, HypervisorGroup]
+        :raises ApplicationError: if there is unit without az defined
         """
-        raise NotImplementedError
+        azs = {}
+        for app in self.apps:
+            for unit in app.units.values():
+                az = unit.machine.az
```
rgildein, line 121:
```suggestion
                az = unit.machine.az or ""
```

rgildein: Like this, we can remove exception below.

tests/unit/steps/test_hypervisor.py
```diff
@@ -0,0 +1,111 @@
...
+        "app1/5": COUUnit("app1/5", machines["5"], ""),
+        # app2
+        "app2/0": COUUnit("app2/0", machines["0"], ""),
+        "app2/1": COUUnit("app2/1", machines["2"], ""),
+        "app2/2": COUUnit("app2/2", machines["4"], ""),
+    }
+
+    app1 = MagicMock(spec_set=COUApplication)()
+    app1.name = "app1"
+    app1.units = {name: app for name, app in units.items() if name.startswith("app1")}
```
gabrielcocenza, line 96:
```suggestion
    app1.units = {name: unit for name, unit in units.items() if name.startswith("app1")}
```

tests/unit/steps/test_hypervisor.py
```diff
@@ -0,0 +1,111 @@
...
+    app2 = MagicMock(spec_set=COUApplication)()
+    app2.name = "app2"
+    app1.units = {name: app for name, app in units.items() if name.startswith("app2")}
```
gabrielcocenza, line 99: same as above

gabrielcocenza (changes requested):

cou/steps/hypervisor.py
agileshaw: I don't understand what you mean here: with my approach the empty list is already added in for app in `azs[az] = HypervisorGroup(name=az, apps={app.name: []})`. Let's discuss it tomorrow in the sync.

cou/steps/hypervisor.py
```diff
@@ -68,10 +96,12 @@ class HypervisorUpgradePlanner:
...
         """Initialize the Hypervisor class.
 
-        :param apps: list of OpenStack applications
+        The application should be sorted by upgrade order.
+
+        :param apps: sorted list of OpenStack applications
         :type apps: list[OpenStackApplication]
         """
-        verify_apps(apps)
+        # verify_apps(apps)
```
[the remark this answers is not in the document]
agileshaw: A TODO comment could be a more clear placeholder

cou/steps/hypervisor.py
agileshaw: Ah I see that you are using defaultdict instead, which looks better.

cou/steps/hypervisor.py
```diff
@@ -29,23 +31,49 @@
...
+
+class AZs(defaultdict):
+    """AZs dictionary object with default value HypervisorGroup."""
+
+    def __init__(self) -> None:
+        """Initialize the Hypervisor class.
+
+        The AZs represent default dict, with predefined HypervisorGroup as default value.
+        """
+        super().__init__(default_factory=None)
```
agileshaw, line 66: This is not needed since we are overriding `__missing__` method (where `default_factory` was normally used)

cou/steps/hypervisor.py
```diff
@@ -114,20 +159,33 @@ def _generate_pre_upgrade_plan(self) -> PreUpgradeStep:
...
+                logger.debug(
+                    "skipping application %s because it is not part of group %s",
+                    app.name,
+                    group.name,
+                )
+                continue
+
+            units = group.apps[app.name]
+            logger.info("generating upgrade steps for %s app and %s units", app.name, units)
+            plan.sub_steps = app.upgrade_steps(target, units)  # type: ignore[call-arg, assignment]
```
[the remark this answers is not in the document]
agileshaw: I missed this in my first review. The current implementation will override `plan.substeps` when iterating through apps. If we are sure that we are doing the upgrade app by app under the same HypervisorGroup, I think we should:
```python
app_plan = ApplicationUpgradePlan(description=f"Upgrade plan for `{app.name}` in '{group.name}' to {target}")
app_plan.sub_steps = app.upgrade_steps(target, units)
plan.add_step(app_plan)
```

agileshaw (changes requested):

cou/steps/hypervisor.py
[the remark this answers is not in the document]
rgildein: I will delete it, since I think I will not use it.

cou/steps/hypervisor.py
```diff
@@ -84,25 +114,40 @@ def apps(self) -> list[OpenStackApplication]:
...
+        :raises ApplicationError: if there is unit without az defined
         """
-        raise NotImplementedError
+        azs = AZs()
+        for app in self.apps:
+            for unit in app.units.values():
+                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
+                #                 belong to a single group.
+                az = unit.machine.az or ""
+                azs[az].apps[app.name].append(unit)
```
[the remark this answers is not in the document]
rgildein: This will be called only once, so I do not see any harm doing it like this. I'll not add any post_init functionality, since that usually cause only troubles and it's not useful here.

cou/steps/hypervisor.py
```diff
@@ -29,23 +31,49 @@
...
-@dataclass(frozen=True)
-class HypervisorMachine:
-    """Hypervisor machine containing units for multiple applications."""
-
-    name: str
-    units: list[COUUnit]
-
-
-@dataclass(frozen=True)
+@dataclass
```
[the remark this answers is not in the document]
rgildein: Because we are adding apps to the group, it's not done at once.

cou/steps/hypervisor.py
```diff
@@ -114,20 +159,33 @@ def _generate_pre_upgrade_plan(self) -> PreUpgradeStep:
...
-        raise NotImplementedError
+        plan = HypervisorUpgradePlan(description=f"Upgrade plan for '{group.name}' to {target}")
+        for app in self.apps:
+            if app.name not in group.apps:
+                logger.debug(
+                    "skipping application %s because it is not part of group %s",
+                    app.name,
+                    group.name,
+                )
+                continue
```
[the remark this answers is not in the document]
rgildein: Sorry I do not understand this question, the `HypervisorGroup` has information about apps, but this is about app not being in group.

cou/steps/hypervisor.py
rgildein: It is because without this you can do `AZs(default_factory=MyFactory)`, which will be ignored.

gabrielcocenza:
/canonical/self-hosted-runners/run-workflows 08c296b

cou/steps/hypervisor.py
```diff
@@ -29,23 +31,49 @@
...
+
+        If key is missing in the AZs dict, the default value
+        HypervisorGroup(name=key, apps=defaultdict(list) will be used. Like this we can always
+        access the az["my-az"].name == "my-az" or az["my-az"].apps["my-app"].
+        """
+        self[key] = HypervisorGroup(name=key, app_units=defaultdict(list))
+        return self[key]
 
 
 def verify_apps(apps: list[OpenStackApplication]) -> None:
```
gabrielcocenza, line 79: I guess we can remove this and add if necessary

cou/steps/hypervisor.py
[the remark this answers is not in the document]
gabrielcocenza: `HypervisorUpgradePlanner` it's not a dataclass so far and there is a `__init__` method. Moving it there it's more appropriate IMO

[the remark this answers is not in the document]
gabrielcocenza: This is not a blocker, so I'm solving it

cou/steps/hypervisor.py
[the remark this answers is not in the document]
gabrielcocenza: As we discussed, the order is important and in rare scenarios a app can be part of an group and not from another.

cou/steps/hypervisor.py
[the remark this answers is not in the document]
gabrielcocenza: I didn't know this behavior. I think it worth adding a simple unit test showing this behavior on the step module as it's the first time that someone is using like this in the project

gabrielcocenza (approved):
Small things to improve, but LGTM

cou/steps/hypervisor.py
```diff
@@ -29,23 +31,49 @@
...
+        :return: True if equal False if different.
+        :rtype: bool
+        """
+        if not isinstance(other, HypervisorGroup):
+            return NotImplemented
+
+        return other.name == self.name
+
+
+class AZs(defaultdict):
```
gabrielcocenza, line 58: I think we should change this class name to `UpgradeGroups`. reviewing other PRs I realized that this name it's confusing

cou/steps/hypervisor.py
```diff
@@ -84,25 +113,40 @@ def apps(self) -> list[OpenStackApplication]:
...
-        raise NotImplementedError
+        azs = AZs()
+        for app in self.apps:
+            for unit in app.units.values():
+                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
+                #                 belong to a single group.
+                az = unit.machine.az or ""
+                azs[az].app_units[app.name].append(unit)
+
+        return azs
```
gabrielcocenza, line 149:
```suggestion
        az_groups = UpgradeGroups()
        for app in self.apps:
            for unit in app.units.values():
                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
                #                 belong to a single group.
                az = unit.machine.az or ""
                az_groups[az].app_units[app.name].append(unit)

        return az_groups
```

With something like this I think it's more agnostic to what is the logic behind the grouping. If in the future it needs to group with something different from azs it would still use the `UpgradeGroups` class, create a new property and populate the keys with whatever is needed.

agileshaw (approved):
LGTM

cou/steps/hypervisor.py
rgildein: We are doing dev branch, so no point to remove it here, since it's not related. Resolving this comment

cou/steps/hypervisor.py
rgildein: We can rename it any time, resolving this comment to unblock my work

tests/unit/steps/test_hypervisor.py
```diff
@@ -0,0 +1,111 @@
...
+    azs = AZs()
+    test_unit = COUUnit("my-unit", MagicMock(spec_set=COUMachine)(), "")
+
+    # test accessing parts of AZs
+    assert azs["my-app"].name == "my-app"
+    assert azs["my-app"].app_units["my-app"] == []
+
+    # append unit to the HypervisorGroup
+    azs["my-app"].app_units["my-app"].append(test_unit)
+    assert azs["my-app"].app_units["my-app"] == [test_unit]
```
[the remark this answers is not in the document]
rgildein: I'll change this latter.

PR #247, opened 2024-02-16
Introducing the usage of `HypervisorUpgradePlanner` in planning stage.

rgildein:
Introduce base structure of the `cou.steps.plan.generate_plan` generating the upgrade plan for data plane applications, right now it will generate only plan using `HypervisorUpgradePlanner` for all application deployed on same machine as Nova-compute.


cou/steps/plan.py
```diff
@@ -317,6 +318,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+
+    subordinate_upgrade_plan = create_upgrade_group(
+        apps=apps,
+        description="Control Plane subordinate(s) upgrade plan",
+        target=target,
+        force=force,
+        filter_function=lambda app: isinstance(app, SubordinateBaseClass),
+    )
+
+    logger.debug("generating of control plan upgrade plan finished")
```
gabrielcocenza, line 351:
```suggestion
    logger.debug("generating of control plan upgrade plane finished")
```

cou/steps/plan.py
```diff
@@ -354,28 +424,16 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
-        apps=analysis_result.apps_control_plane,
-        description="Control Plane subordinate(s) upgrade plan",
-        target=target,
-        force=args.force,
-        filter_function=lambda app: isinstance(app, SubordinateBaseClass),
+    plan.sub_steps.extend(
+        _generate_control_plane_plan(target, analysis_result.apps_control_plane, args.force)
     )
-    plan.add_step(control_plane_subordinate_upgrade_plan)
+    # plan.sub_steps.extend(
```
gabrielcocenza, line 434: I guess you can remove this comment and the `# pragma: no cover`

cou/steps/plan.py
```diff
@@ -317,6 +318,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+        for app in apps
+        for unit in app.units.values()
+        if app.charm == "nova-compute"
+    ]
+    hypervisor_apps = []
+    for app in apps:
+        for machine in app.machines:
+            if machine in nova_compute_machines:
+                hypervisor_apps.append(app)
+                break  # exiting machine for loop, since
```
gabrielcocenza, line 380: This is another reason why I think that the machine should have knowledge about if there is nova-compute on it or not. I think #248 should be merged first than this PR. With that I think this logic will be changed.

cou/steps/plan.py
```diff
@@ -317,6 +318,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+
+    hypervisor_planner = HypervisorUpgradePlanner(hypervisor_apps)
+    hypervisor_plans = hypervisor_planner.generate_upgrade_plan(target, force)
+    logger.debug("generating of hypervisors upgrade plan finished")
+
+    # generating rest
```
gabrielcocenza, line 386: didn't understand this comment

cou/steps/plan.py
rgildein: I think you want to suggest to change first plan to plane.

cou/steps/plan.py
rgildein: I cannot, since I do not add any tests for this.

cou/steps/plan.py
rgildein: I agree.

cou/steps/plan.py
rgildein: For example, the ceph-osd will be there.

gabrielcocenza:
/canonical/self-hosted-runners/run-workflows eadbc4d

gabrielcocenza:
/canonical/self-hosted-runners/run-workflows eadbc4d1e45027e21389746e688d5255d3b5c1fc

agileshaw:
/canonical/self-hosted-runners/run-workflows 99bfc316ab68d24684fececc6f3db83f988db884

cou/steps/plan.py
```diff
@@ -293,6 +294,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+        for app in apps
+        for unit in app.units.values()
+        if app.charm == "nova-compute"
+    ]
+    hypervisor_apps = []
+    for app in apps:
+        for machine in app.machines:
+            if machine in nova_compute_machines:
+                hypervisor_apps.append(app)
+                break  # exiting machine for loop, since
```
aieri, line 356: truncated comment

cou/steps/plan.py
```diff
@@ -293,6 +294,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+
+    hypervisor_planner = HypervisorUpgradePlanner(hypervisor_apps)
+    hypervisor_plans = hypervisor_planner.generate_upgrade_plan(target, force)
+    logger.debug("generating of hypervisors upgrade plan finished")
```
aieri, line 360:
```suggestion
    logger.debug("Generation of the hypervisors upgrade plan complete")
```

aieri (approved):
Verbiage nits, otherwise +1

cou/steps/plan.py
gabrielcocenza: We can let this to be changed later, so solving it.

cou/steps/plan.py
```diff
@@ -293,6 +294,75 @@ def _get_os_release_and_series(analysis_result: Analysis) -> tuple[OpenStackRele
...
+        unit.machine.machine_id
+        for app in apps
+        for unit in app.units.values()
+        if app.charm == "nova-compute"
+    ]
+    hypervisor_apps = []
+    for app in apps:
+        for machine in app.machines:
+            if machine in nova_compute_machines:
+                hypervisor_apps.append(app)
```
gabrielcocenza, line 355: I think we should keep ceph-osd out of the hypervisors_apps. This is also including subordinate charms. I guess it's not a issue now, but we need to keep an eye on it.

cou/steps/plan.py
gabrielcocenza:
```suggestion
    # generate the plan for the remaining data-plane apps
```
See my comment above of removing ceph-osd from hypervisor_apps.

cou/steps/hypervisor.py
```diff
@@ -198,13 +198,15 @@ def _generate_post_upgrade_plan(self) -> PostUpgradeStep:
         raise NotImplementedError
 
     # pylint: disable=unused-argument
-    def generate_upgrade_plan(self, target: OpenStackRelease) -> UpgradePlan:
+    def generate_upgrade_plan(self, target: OpenStackRelease, force: bool) -> UpgradePlan:
```
gabrielcocenza, line 201: the planning so far is filtering the hypervisors machines. I think it makes sense accept this parameter in this function. If you think it should be part of another PR I'm fine with it, but we are already adding force so it might be an opportunity.

gabrielcocenza (commented):
Just some comments to be discussed, but non-blockers

cou/steps/plan.py
rgildein: You know that this comment is just for you now and will be removed after we actually do ceph-osd.

cou/steps/hypervisor.py
rgildein: Yes, and with that I need to add `__eq__` to COUMachie, since you cannot do `machine_id in machines`, where machines are `list[COUMachine]`.

gabrielcocenza: I'm assuming that won't be part of this PR, so solving it

gabrielcocenza (approved):
This PR has some conflicts with the one that I'm proposing #255 to separate control_plane from data_plane plan generation. However, I don't see a problem merging this and deal with the detail on this other PR

agileshaw (approved):
LGTM.

PR #251, opened 2024-02-20
Add pre- and post-upgrade steps to `HypervisorUpgradePlanner`

rgildein:
Add pre- and post-upgrade steps to `HypervisorUpgradePlanner`, where post-upgrade steps are generated in reverse order. I think it will be safer at least for waiting step, where we should wait for subordinate app to be in idle, than colocated and at the last the nova-compute.

### DIFF
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -14,7 +14,9 @@
 
 """Hypervisor planner."""
 import logging
+from collections import defaultdict
 from dataclasses import dataclass
+from typing import Any
 
 from cou.apps.base import OpenStackApplication
 from cou.steps import (
@@ -22,6 +24,7 @@ from cou.steps import (
     PostUpgradeStep,
     PreUpgradeStep,
     UpgradePlan,
+    UpgradeStep,
 )
 from cou.utils.juju_utils import COUUnit
 from cou.utils.openstack import OpenStackRelease
@@ -29,15 +32,7 @@ from cou.utils.openstack import OpenStackRelease
 logger = logging.getLogger(__name__)
 
 
-@dataclass(frozen=True)
-class HypervisorMachine:
-    """Hypervisor machine containing units for multiple applications."""
-
-    name: str
-    units: list[COUUnit]
-
-
-@dataclass(frozen=True)
+@dataclass
 class HypervisorGroup:
     """Group of hypervisors.
 
@@ -45,7 +40,41 @@ class HypervisorGroup:
     """
 
     name: str
-    hypervisors: list[HypervisorMachine]
+    app_units: dict[str, list[COUUnit]]
+
+    def __eq__(self, other: Any) -> bool:
+        """Equal magic method for HypervisorGroup.
+
+        :param other: HypervisorGroup object to compare.
+        :type other: Any
+        :return: True if equal False if different.
+        :rtype: bool
+        """
+        if not isinstance(other, HypervisorGroup):
+            return NotImplemented
+
+        return other.name == self.name
+
+
+class AZs(defaultdict):
+    """AZs dictionary object with default value HypervisorGroup."""
+
+    def __init__(self) -> None:
+        """Initialize the Hypervisor class.
+
+        The AZs represent default dict, with predefined HypervisorGroup as default value.
+        """
+        super().__init__(default_factory=None)
+
+    def __missing__(self, key: str) -> HypervisorGroup:
+        """Handle missing key in AZs.
+
+        If key is missing in the AZs dict, the default value
+        HypervisorGroup(name=key, apps=defaultdict(list) will be used. Like this we can always
+        access the az["my-az"].name == "my-az" or az["my-az"].apps["my-app"].
+        """
+        self[key] = HypervisorGroup(name=key, app_units=defaultdict(list))
+        return self[key]
 
 
 def verify_apps(apps: list[OpenStackApplication]) -> None:
@@ -68,7 +97,9 @@ class HypervisorUpgradePlanner:
     def __init__(self, apps: list[OpenStackApplication]) -> None:
         """Initialize the Hypervisor class.
 
-        :param apps: list of OpenStack applications
+        The application should be sorted by upgrade order.
+
+        :param apps: sorted list of OpenStack applications
         :type apps: list[OpenStackApplication]
         """
         self._apps = apps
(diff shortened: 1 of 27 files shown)


gabrielcocenza:
/canonical/self-hosted-runners/run-workflows fca5a9c2a1136e0b7f5eb681b1cfc626f3d2f9b8

cou/steps/hypervisor.py
```diff
@@ -104,15 +103,22 @@ def azs(self) -> list[HypervisorGroup]:
...
+        :param target: OpenStack codename to upgrade.
+        :type target: OpenStackRelease
+        :return: List of pre-upgrade steps.
+        :rtype: list[PreUpgradeStep]
         """
-        raise NotImplementedError
+        steps = []
+        for app in self.apps:
+            steps.extend(app.pre_upgrade_steps(target, units=None))
+            logger.info("generating pre-upgrade steps for %s app", app.name)
```
gabrielcocenza, line 119: I don't think this will work passing no units. There are some pre-upgrade steps, like refresh the current packages that needs to be targeted with the specific units in the group that it's going to be upgraded

cou/steps/hypervisor.py
```diff
@@ -129,16 +135,25 @@ def _generate_hypervisor_upgrade_plan(
...
+        :type target: OpenStackRelease
+        :return: List of post-upgrade steps.
+        :rtype: list[PostUpgradeStep]
         """
-        raise NotImplementedError
+        steps = []
+        # NOTE(rgildein): Using the reverse order of post-upgrade steps, so these steps starts from
+        #                 subordinate or colocated apps and not nova-compute.
+        for app in self.apps[::-1]:
+            steps.extend(app.post_upgrade_steps(target, units=None))
```
gabrielcocenza, line 153: same here. We need to pass specific units

cou/steps/hypervisor.py
aieri, line 1: it looks like the idea here is for a hypervisor upgrade to occur as follows:
1. pre upgrade steps for all units on the hypervisor, ordered according to the docs
2. upgrade of all units on the hypervisor, ordered according to the docs
3. post upgrade steps for all units on the hypervisor, in reverse order

I was imagining it would have rather been:
1. pre upgrade steps for the whole hypervisor (if any)
2. for every unit on the hypervisor, ordered according to the docs:
  1. pre upgrade
  2. upgrade
  3. post upgrade
4. post upgrade steps for the whole hypervisor (if any)

Could we see as an example the unpacking of an hypothetical nova-compute + cinder hypervisor to validate which approach is more appropriate?

rgildein: I added unit test and here is an example of hypervisor plan.
```
Upgrading all applications deployed on machines with hypervisor.
        Upgrade plan for 'az-0' to victoria
                Upgrade software packages of 'cinder' from the current APT repositories
                        Upgrade software packages on unit cinder/0
                Refresh 'cinder' to the latest revision of 'ussuri/stable'
                Upgrade software packages of 'nova-compute' from the current APT repositories
                        Upgrade software packages on unit nova-compute/0
                Refresh 'nova-compute' to the latest revision of 'ussuri/stable'
                Upgrade 'cinder' to the new channel: 'victoria/stable'
                Change charm config of 'cinder' 'openstack-origin' to 'cloud:focal-victoria'
                Change charm config of 'nova-compute' 'action-managed-upgrade' to True.
                Upgrade 'nova-compute' to the new channel: 'victoria/stable'
                Change charm config of 'nova-compute' 'source' to 'cloud:focal-victoria'
                Wait 300s for app nova-compute to reach the idle state.
                Check if the workload of 'nova-compute' has been upgraded on units: nova-compute/0
                Wait 300s for app cinder to reach the idle state.
                Check if the workload of 'cinder' has been upgraded on units: cinder/0
        Upgrade plan for 'az-1' to victoria
                Upgrade software packages of 'nova-compute' from the current APT repositories
                        Upgrade software packages on unit nova-compute/1
                Refresh 'nova-compute' to the latest revision of 'ussuri/stable'
                Change charm config of 'nova-compute' 'action-managed-upgrade' to True.
                Upgrade 'nova-compute' to the new channel: 'victoria/stable'
                Change charm config of 'nova-compute' 'source' to 'cloud:focal-victoria'
                Wait 300s for app nova-compute to reach the idle state.
                Check if the workload of 'nova-compute' has been upgraded on units: nova-compute/1
        Upgrade plan for 'az-2' to victoria
                Upgrade software packages of 'nova-compute' from the current APT repositories
                        Upgrade software packages on unit nova-compute/2
                Refresh 'nova-compute' to the latest revision of 'ussuri/stable'
                Change charm config of 'nova-compute' 'action-managed-upgrade' to True.
                Upgrade 'nova-compute' to the new channel: 'victoria/stable'
                Change charm config of 'nova-compute' 'source' to 'cloud:focal-victoria'
                Wait 300s for app nova-compute to reach the idle state.
                Check if the workload of 'nova-compute' has been upgraded on units: nova-compute/2
```

gabrielcocenza: I'm confused why it's not showing the enable / disable, instances counts , run openstack upgrade and etc. I think I ended up not generating a unit test checking the whole plan for nova-compute so I'll investigate it tomorrow

aieri: ok this is good. I think your approach is the right one. There are steps missing from your printout but we can discuss those separately.

aieri (approved):

aieri (approved):
looks like nothing changed since my last review, so re-approving

cou/steps/hypervisor.py
```diff
@@ -148,29 +150,51 @@ def azs(self) -> AZs:
...
+                    app.name,
+                    group.name,
+                )
+                continue
 
-    def _generate_hypervisor_group_upgrade_plan(
-        self, target: OpenStackRelease, group: HypervisorGroup
-    ) -> HypervisorUpgradePlan:
+            units = group.app_units[app.name]
+            logger.info("generating pre-upgrade steps for %s app and %s units", app.name, units)
```
agileshaw, line 178: The current phrase sounds like the app and units are not necessary related. I suggest updating it a little:
```suggestion
            logger.info("generating pre-upgrade steps for %s units of %s app", app.name, units)
```

cou/steps/hypervisor.py
```diff
@@ -148,29 +150,51 @@ def azs(self) -> AZs:
...
+            steps.extend(app.pre_upgrade_steps(target, units))
+
+        return steps
+
+    def _generate_upgrade_steps(
+        self, target: OpenStackRelease, force: bool, group: HypervisorGroup
+    ) -> list[UpgradeStep]:
```
agileshaw, line 185: I forgot the reason why having `force` is needed for upgrade steps. Is it to further filter machine if there are vms spinned up after plan stage but before the actual upgrade operation? If so, why we don't want to also check it in the pre-upgrade stage?

cou/steps/hypervisor.py
```diff
@@ -182,22 +206,43 @@ def _generate_hypervisor_group_upgrade_plan(
 
             units = group.app_units[app.name]
             logger.info("generating upgrade steps for %s app and %s units", app.name, units)
```
agileshaw, line 208: same as above

cou/steps/hypervisor.py
```diff
@@ -182,22 +206,43 @@ def _generate_hypervisor_group_upgrade_plan(
...
+            if app.name not in group.app_units:
+                logger.debug(
+                    "skipping application %s because it is not part of group %s",
+                    app.name,
+                    group.name,
+                )
+                continue
+
+            units = group.app_units[app.name]
+            logger.info("generating post-upgrade steps for %s app and %s units", app.name, units)
```
agileshaw, line 241: also here

cou/steps/hypervisor.py
```diff
@@ -211,18 +256,21 @@ def generate_upgrade_plan(self, target: OpenStackRelease, force: bool) -> Upgrad
...
-        logger.debug("generating pre upgrade steps")
-        plan.add_step(self._generate_pre_upgrade_plan())
+            # pre upgrade steps
+            logger.debug("generating pre upgrade steps")
+            hypervisor_plan.sub_steps.extend(self._generate_pre_upgrade_steps(target, group))
 
-        # upgrade steps
-        for az, group in self.azs.items():
+            # upgrade steps
             logger.debug("generating upgrade steps for %s AZ", az)
```
agileshaw, line 267: Please keep the debug message consistent across different stages of upgrades

agileshaw (changes requested):
Some questions and logging message suggestions

cou/steps/hypervisor.py
```diff
@@ -63,7 +64,8 @@ def __init__(self) -> None:
 
         The AZs represent default dict, with predefined HypervisorGroup as default value.
         """
-        super().__init__(default_factory=None)
+        super().__init__()
+        self.default_factory = None
```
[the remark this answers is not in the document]
rgildein: Without this, the test failed due `default_factory` being part of the dict itself.  So during:
```python
for az, group in self.azs.items():
```
the first az was `default_dict` and group was `None`.

cou/steps/hypervisor.py
rgildein: What do you mean?

cou/steps/hypervisor.py
rgildein: To be able to run steps with force, yes for example for nova-comput it means that we can upgrade non empty hypervisors. 

> If so, why we don't want to also check it in the pre-upgrade stage?

I do not understand, why would you need it in pre-upgrade stage?

cou/steps/hypervisor.py
```diff
@@ -181,23 +205,44 @@ def _generate_hypervisor_group_upgrade_plan(
...
+                continue
+
+            units = group.app_units[app.name]
+            logger.info("generating post-upgrade steps for %s units of %s app", app.name, units)
+            steps.extend(app.post_upgrade_steps(target, units=units))
+
+        return steps
 
-    # pylint: disable=unused-argument
     def generate_upgrade_plan(self, target: OpenStackRelease, force: bool) -> UpgradePlan:
```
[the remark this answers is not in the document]
rgildein: This pull request is about pre- and post-upgrade steps, but yes we will implement it later passing units, machines or AZs, which will be used to target only certain AZ with certain machines.

cou/steps/hypervisor.py
agileshaw: You are not logging AZ name in pre-upgrade and post_upgarde steps. It would be better to align them. This was meant to be a nit comment though.

cou/steps/hypervisor.py
agileshaw: > To be able to run steps with force, yes for example for nova-comput it means that we can upgrade non empty hypervisors.

Maybe I'm missing something. My understanding is that if `force` is recorded, we will not filter out machines that are non-empty, so upgrade steps will generate steps for them even without accepting a `force` parameter.

But if it is really needed, I believe we should also add it for pre-upgrade stage because to not by default upgrading packages if there are vms running. Unless that's not risky at all?

rgildein: Yes, that is one part. But we are passing force to the upgrade plan to not checking if empty hypervisor again (ignoring it).
We are checking it during planning and also part of upgrade phase, as upgrade step (e.g. [line](https://github.com/canonical/charmed-openstack-upgrader/blob/70f3814f31821dc0abef67048038fdbb33e59409/tests/unit/steps/test_hypervisor.py#L215)).

cou/steps/hypervisor.py
rgildein: Ah now I see, yes because originally it was not part of for loop. I'll change it.

agileshaw (approved):

PR #252, opened 2024-02-20
drop `--hostname` from CLI

rgildein:
After discussion, we agreed that this option is not needed, since it can be confusing which hostname the user meant.

### DIFF
--- a/tests/unit/test_commands.py
+++ b/tests/unit/test_commands.py
@@ -146,7 +146,6 @@ def test_parse_args_quiet_verbose_exclusive(args):
                 backup=True,
                 force=False,
                 machines=None,
-                hostnames=None,
                 availability_zones=None,
                 **{"upgrade_group": "data-plane"}
             ),
@@ -161,7 +160,6 @@ def test_parse_args_quiet_verbose_exclusive(args):
                 backup=True,
                 force=True,
                 machines=None,
-                hostnames=None,
                 availability_zones=None,
                 **{"upgrade_group": "data-plane"}
             ),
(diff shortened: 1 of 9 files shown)


tests/unit/test_commands.py
```diff
@@ -233,57 +228,10 @@ def test_parse_args_quiet_verbose_exclusive(args):
...
-                force=True,
-                machines=None,
-                hostnames={"1", "2", "3"},
-                availability_zones=None,
-                **{"upgrade_group": "data-plane"}
-            ),
-        ),
-        (
-            # repetitive machine 3
-            ["plan", "data-plane", "--hostname=1,2,3", "--force", "-n=3,4"],
```
gabrielcocenza, line 273: I think we can keep this test and just change `hostname`to `machine`. This test is also checking that no repetitive elements appears after the parding

tests/unit/test_commands.py
```diff
@@ -554,115 +494,10 @@ def test_parse_args_plan(args, expected_CLIargs):
...
-                backup=True,
-                force=True,
-                machines=None,
-                hostnames={"1", "2", "3"},
-                availability_zones=None,
-                **{"upgrade_group": "data-plane"}
-            ),
-        ),
-        (
-            # repetitive machine 3
```
gabrielcocenza, line 643: Same here

gabrielcocenza (commented):
Just two small comments, but LGTM

gabrielcocenza:
/canonical/self-hosted-runners/run-workflows 56f977c0aa975c3b9ff32f2b5d43281da3e6420d

agileshaw (commented):
+1 on Gabriel's suggestion of adding the repetitive tests back. Otherwise LGTM

gabrielcocenza (approved):

agileshaw (approved):
LGTM

PR #255, opened 2024-02-21
Separate control-plane and data-plane plans

gabrielcocenza:
- depending on the cli args some plans might need to be skipped

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -387,6 +387,15 @@ class CLIargs:
         """
         return self.upgrade_group == DATA_PLANE
 
+    @property
+    def is_control_plane_command(self) -> bool:
+        """Whether if the command passed is control-plane related.
+
+        :return: True if is control-plane, false otherwise.
+        :rtype: bool
+        """
+        return self.upgrade_group == CONTROL_PLANE or not self.is_data_plane_command
+
 
 def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-statements
     """Parse cli arguments.
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 """Command line arguments parsing for 'charmed-openstack-upgrader'."""
+from __future__ import annotations
+
 import argparse
 import logging
 from dataclasses import dataclass
+from enum import Enum
 from typing import Any, Iterable, Optional
 
 import pkg_resources
@@ -408,46 +411,31 @@ class CLIargs:
         return not self.auto_approve
 
     @property
-    def is_hypervisors_command(self) -> bool:
-        """Whether if the command passed is specific to hypervisors.
+    def scope(self) -> UpgradeScope:
+        """Define which is the scope to upgrade.
 
-        :return: True if is hypervisors, false otherwise.
-        :rtype: bool
+        Scopes can be control-plane, data-plane or the whole cloud.
+
+        :return: Scope to upgrade
+        :rtype: UpgradeScope
         """
-        return self.upgrade_group == HYPERVISORS
+        return UpgradeScope(self.upgrade_group)
 
-    @property
-    def is_data_plane_command(self) -> bool:
-        """Whether if the command passed is data-plane related.
 
-        :return: True if is data-plane, false otherwise.
-        :rtype: bool
-        """
-        return self.is_hypervisors_command or self.upgrade_group == DATA_PLANE
+class UpgradeScope(Enum):
+    """Possible upgrade scopes."""
 
-    @property
-    def is_control_plane_command(self) -> bool:
-        """Whether if the command passed is control-plane related.
+    CONTROL_PLANE = CONTROL_PLANE
+    DATA_PLANE = DATA_PLANE, HYPERVISORS
+    WHOLE_CLOUD = None
 
-        :return: True if is control-plane, false otherwise.
-        :rtype: bool
-        """
-        return self.upgrade_group == CONTROL_PLANE
-
-    @property
-    def is_generic_command(self) -> bool:
-        """Whether the command has no specific group for upgrading.
-
-        When this happens both control-plane and data-plane should generate
-        plans.
-        E.g of generic commands:
-            - cou upgrade
-            - cou plan
-
-        :return: True if upgrade_group is None, false otherwise.
-        :rtype: bool
-        """
-        return self.upgrade_group is None
+    def __new__(cls, *values: Any) -> UpgradeScope:
+        obj = object.__new__(cls)
+        # first value is canonical value
+        obj._value_ = values[0]
+        for other_value in values[1:]:
+            cls._value2member_map_[other_value] = obj
+        return obj
 
 
 def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-statements
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -394,7 +394,22 @@ class CLIargs:
         :return: True if is control-plane, false otherwise.
         :rtype: bool
         """
-        return self.upgrade_group == CONTROL_PLANE or not self.is_data_plane_command
+        return self.upgrade_group == CONTROL_PLANE
+
+    @property
+    def is_generic_command(self) -> bool:
+        """Whether the command has no specific group for upgrading.
+
+        When this happens both control-plane and data-plane should generate
+        plans.
+        E.g of generic commands:
+            - cou upgrade
+            - cou plan
+
+        :return: True if upgrade_group is None, false otherwise.
+        :rtype: bool
+        """
+        return self.upgrade_group is None
 
 
 def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-statements
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 import argparse
 import logging
 from dataclasses import dataclass
-from enum import Enum
 from typing import Any, Iterable, Optional
 
 import pkg_resources
@@ -399,33 +398,6 @@ class CLIargs:
         :rtype: bool
         """
         return not self.auto_approve
-
-    @property
-    def scope(self) -> UpgradeScope:
-        """Define which is the scope to upgrade.
-
-        Scopes can be control-plane, data-plane or the whole cloud.
-
-        :return: Scope to upgrade
-        :rtype: UpgradeScope
-        """
-        return UpgradeScope(self.upgrade_group)
-
-
-class UpgradeScope(Enum):
-    """Possible upgrade scopes."""
-
-    CONTROL_PLANE = CONTROL_PLANE
-    DATA_PLANE = DATA_PLANE, HYPERVISORS
-    WHOLE_CLOUD = None
-
-    def __new__(cls, *values: Any) -> UpgradeScope:
-        obj = object.__new__(cls)
-        # first value is canonical value
-        obj._value_ = values[0]
-        for other_value in values[1:]:
-            cls._value2member_map_[other_value] = obj
-        return obj
 
 
 def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-statements
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -13,8 +13,6 @@
 # limitations under the License.
 
 """Command line arguments parsing for 'charmed-openstack-upgrader'."""
-from __future__ import annotations
-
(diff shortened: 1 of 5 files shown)


cou/commands.py
```diff
@@ -387,6 +387,15 @@ def is_data_plane_command(self) -> bool:
...
         return self.upgrade_group == DATA_PLANE
 
+    @property
+    def is_control_plane_command(self) -> bool:
+        """Whether if the command passed is control-plane related.
+
+        :return: True if is control-plane, false otherwise.
+        :rtype: bool
+        """
+        return self.upgrade_group == CONTROL_PLANE or not self.is_data_plane_command
```
rgildein, line 397: Why we need to have or?

cou/steps/plan.py
```diff
@@ -304,10 +304,34 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
-    logger.info("Hypervisors selected: %s", hypervisors)
     target = determine_upgrade_target(analysis_result)
 
+    plan = generate_common_plan(analysis_result, args, target)
+
+    control_plane_plan = await generate_control_plane_plan(analysis_result, args, target)
+    data_plane_plan = await generate_data_plane_plan(analysis_result, args)
+
+    for group in control_plane_plan + data_plane_plan:
+        if group:
```
rgildein, line 315: We have function to generate empty plan an here we are separeting filtering that, so like this we have two if instead of one and I do not see any benefit of it.

cou/steps/plan.py
```diff
@@ -304,10 +304,34 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     :rtype: UpgradePlan
     """
     pre_plan_sanity_checks(args, analysis_result)
-    hypervisors = await filter_hypervisors_machines(args, analysis_result)
-    logger.info("Hypervisors selected: %s", hypervisors)
     target = determine_upgrade_target(analysis_result)
 
+    plan = generate_common_plan(analysis_result, args, target)
+
+    control_plane_plan = await generate_control_plane_plan(analysis_result, args, target)
```
rgildein, line 311: Why not simply do if is data-plane add it as sub-step? Name of function is generate and not get, so it should generate and not return empty step or None.

tests/unit/steps/test_steps_plan.py
```diff
@@ -758,3 +758,60 @@ async def test_get_upgradable_hypervisors_machines(
     assert {
         hypervisor.machine_id for hypervisor in hypervisors_possible_to_upgrade
     } == expected_result
+
+
+@pytest.mark.parametrize("backup", [True, False])
+@patch("cou.steps.plan.UpgradePlan")
```
rgildein, line 764: You do not test what kind of step are in common, only chceling that there are 2 or one step.

tests/unit/steps/test_steps_plan.py
```diff
@@ -758,3 +758,60 @@ async def test_get_upgradable_hypervisors_machines(
...
+@patch("cou.steps.plan.PreUpgradeStep")
```
rgildein, line 765: Why this is needed?

tests/unit/test_commands.py
```diff
@@ -18,7 +18,16 @@
 import pytest
 
 from cou import commands
-from cou.commands import CLIargs
+from cou.commands import CONTROL_PLANE, DATA_PLANE, CLIargs
+
+
+@pytest.mark.parametrize(
+    "upgrade_group, expected_result",
+    [(CONTROL_PLANE, True), (DATA_PLANE, False), ("foo", True)],
```
rgildein, line 26: Hmm, "foo" is control plane? How we will use this?

rgildein (changes requested):

tests/unit/steps/test_steps_plan.py
gabrielcocenza: I've changed a little bit. See if the approach now is better

tests/unit/test_commands.py
gabrielcocenza: I changed the logic so we are not using it anymore

cou/steps/plan.py
rgildein, line 311:
```python
if args.is_controle_plane:
  plan.add_substep(_generate...)
```

cou/steps/plan.py
aieri: yeah I find it awkward as well. One possible way around it would be to have `add_step` be able to handle noop plans, so you never have to worry about what you're sending through

cou/steps/plan.py
```diff
@@ -334,26 +358,59 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
+    """_summary_.
+
+    :param analysis_result: _description_
+    :type analysis_result: Analysis
+    :param args: _description_
+    :type args: CLIargs
+    :param target: _description_
+    :type target: OpenStackRelease
+    :return: _description_
+    :rtype: tuple[UpgradePlan, UpgradePlan]
```
aieri, line 376: non-blocking: for consistency, I think all the `generate_*` functions should have the same rtype. Instead of a tuple you could return a single plan (with principal and subordinate as substeps)

aieri (changes requested):
marking as 'request changes' because I agree with Robert. My additional comments are non-blocking.

cou/steps/plan.py
rgildein: It's ok to have empty plan/step since it is not printed and not executed.

cou/steps/plan.py
rgildein: Yes, I agree. If the name of function is `get_*` ok, it can return empty object or None, but not `generate_*`.

rgildein:
```python
    plan = generate_common_plan(analysis_result, args, target)

    if args.is_control_plane_command:
        plan.add_step(await generate_control_plane_plan(analysis_result, args, target))

    if args.is_data_plane_command:
        plan.add_step(await generate_data_plane_plan(analysis_result, args))

```
It is cleaner and much obvious what we are trying to do, instead of having to generate returning empty step and then have to loop with  `control_plane_plan + data_plane_plan` and `if group`.

tests/unit/steps/test_steps_plan.py
```diff
@@ -761,57 +761,69 @@ async def test_get_upgradable_hypervisors_machines(
...
+    plan = cou_plan.generate_common_plan(mock_analysis_result, cli_args, target)
 
     if backup:
-        assert mock_PreUpgradeStep.call_count == 2
+        assert len(plan.sub_steps) == 2
+        assert plan.sub_steps[1].description == "Backup mysql databases"
     else:
-        mock_PreUpgradeStep.call_count == 1
+        assert len(plan.sub_steps) == 1
+        assert plan.sub_steps[0].description != "Backup mysql databases"
```
rgildein, line 785: Like this you are testing two things at once, I do not understand what we want to test. I really think we should define exp_steps even in parametrize.

rgildein (changes requested):
We should move the `if args.is_data_plane_command or args.is_generic_command` out from generating a plan.

cou/steps/plan.py
gabrielcocenza: I agree with the consistency and now they are returning `list[UpgradePlan]`. I think returning a simple plan with principal and subordinates as substeps could change the indentation of the plan and break some unit tests.

cou/commands.py
```diff
@@ -387,6 +387,30 @@ def is_data_plane_command(self) -> bool:
         """
         return self.upgrade_group == DATA_PLANE
 
+    @property
+    def is_control_plane_command(self) -> bool:
```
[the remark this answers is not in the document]
agileshaw: I'm also not a fan of these checks. I don't really see the problem of passing in `upgrade_group` as-is to the plan generation functions. And if the distinction between `data-plane` and `hypervisor` are not needed in other parts of the code (I'm not seeing them used differently yet), we can have a setter function to merge them into a single value.

cou/commands.py
```diff
@@ -408,46 +411,31 @@ def prompt(self) -> bool:
...
-            - cou plan
+    CONTROL_PLANE = CONTROL_PLANE
+    DATA_PLANE = DATA_PLANE, HYPERVISORS
+    WHOLE_CLOUD = None
 
-        :return: True if upgrade_group is None, false otherwise.
-        :rtype: bool
-        """
-        return self.upgrade_group is None
+    def __new__(cls, *values: Any) -> UpgradeScope:
```
rgildein, line 432: What is this? What is canonical value?

cou/steps/plan.py
```diff
@@ -116,7 +116,7 @@ def verify_data_plane_ready_to_upgrade(args: CLIargs, analysis_result: Analysis)
     :type analysis_result: Analysis
     :raises DataPlaneCannotUpgrade: When data-plane is not ready to upgrade.
     """
-    if args.is_data_plane_command:
+    if args.scope is UpgradeScope.DATA_PLANE:
```
rgildein, line 119: @aieri we mention enums yesterday, but this do not make sense to me. Why not using `args.upgrade_group == DATA_PLANE`?

gabrielcocenza: Because there is also the `hypervisors` that is considered as data-plane

cou/commands.py
gabrielcocenza: This is a tweak to have multi-value enum. With this is possible to have something like:

```
class A(Enum):
  B = 1, "foo"
  C = 2, "bar"
```

This means that A(1) == A("foo") == A.B and A(2) == A("bar") == A.C

> What is canonical value?
It's the value as the Enum implementation normally works

cou/steps/plan.py
aieri: yeah that's the problem with having unit tests that use text representation as a substitute for internal structure... :/

Unit tests should not dictate how we write the rest of the code, btw

rgildein (approved):

tests/unit/steps/test_steps_plan.py
```diff
@@ -701,3 +689,89 @@ async def test_get_upgradable_hypervisors_machines(
...
+        mock_control_plane.assert_called_once()
+        mock_data_plane.assert_called_once()
+
+    if upgrade_group == CONTROL_PLANE:
+        mock_control_plane.assert_called_once()
+        mock_data_plane.assert_not_called()
+
+    if upgrade_group in {DATA_PLANE, HYPERVISORS}:
+        mock_control_plane.assert_not_called()
+        mock_data_plane.assert_called_once()
```
agileshaw, line 777: non-blocking: I'm not a fan of parameterize with conditional checks. They are better separated into different test functions.

tests/unit/steps/test_steps_plan.py
```diff
@@ -497,37 +500,28 @@ def test_create_upgrade_plan_failed(force):
...
-    if is_hypervisors_command:
+    if machines:
         mock_cli_machines.assert_called_once_with(cli_args.machines, analysis_result)
+        mock_cli_azs.assert_not_called()
+    elif azs:
+        mock_cli_machines.assert_not_called()
         mock_cli_azs.assert_called_once_with(cli_args.availability_zones, analysis_result)
     else:
         mock_cli_machines.assert_not_called()
         mock_cli_azs.assert_not_called()
```
agileshaw, line 522: non-blocking: same here.

tests/unit/steps/test_steps_plan.py
```diff
@@ -344,16 +347,16 @@ def test_verify_data_plane_ready_to_upgrade_error(
...
     mock_analysis_result = MagicMock(spec=Analysis)()
     mock_analysis_result.min_os_version_data_plane = OpenStackRelease("ussuri")
-    cli_args.is_data_plane_command = is_data_plane_command
+    cli_args.upgrade_group = upgrade_group
     cou_plan.verify_data_plane_ready_to_upgrade(cli_args, mock_analysis_result)
-    if is_data_plane_command:
+    if upgrade_group in {DATA_PLANE, HYPERVISORS}:
         mock_control_plane_upgraded.assert_called_once_with(mock_analysis_result)
     else:
         mock_control_plane_upgraded.assert_not_called()
```
agileshaw, line 362: non-blocking: and here.

agileshaw (approved):
LGTM. Some suggestions on tests inline, but they are not blockers

aieri (approved):

rgildein (approved):

agileshaw (approved):
LGTM. Thanks for splitting the tests.

PR #257, opened 2024-02-22
fix zaza.model.run_on_unit

rgildein:
Fixing the functional tests running command on unit. 

Zaza add waiting for action result in [commit](https://github.com/openstack-charmers/zaza/commit/3e7b2a5ec74b2ada7a2c2286ef57b31f5c0eac53), which cause waiting forever. I reported an issue on zaza side, since that function previous supported running commands and now it supports only running actions.

### DIFF
--- a/tests/functional/tests/model.py
+++ b/tests/functional/tests/model.py
@@ -58,7 +58,8 @@ class COUModelTest(unittest.TestCase):
         test_file = "test.txt"
         path = f"/tmp/{test_file}"
         exp_path = os.path.join(tmp_dir, test_file)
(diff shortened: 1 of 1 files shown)


gabrielcocenza:
/canonical/self-hosted-runners/run-workflows 49effb0e89cadeb2b13e07749883d89ff97d9186

gabrielcocenza (approved):
LGTM. Thanks for the fix

rgildein:
zaza issue: [#649](https://github.com/openstack-charmers/zaza/issues/649)

agileshaw (approved):
For our purses, using libjuju is a reasonable workaround. So I'm +1 on this change

aieri (approved):

Issue #140, opened 2023-10-25
manually_upgrade_data_plane should be removed when cou start supporting data plane upgrade


gabrielcocenza:
Changes introduced on #132 should be removed once we starting supporting data plane upgrades

PR #259, opened 2024-02-27
remove message to manually upgrade data-plane

rgildein:
Since the dev/data-plane is introducing upgrade for data-plane there is no reason to left this message here.

fixes: #140

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -105,7 +105,7 @@ async def continue_upgrade() -> bool:
     return False
 
 
-async def analyze_and_plan(args: CLIargs) -> tuple[Analysis, UpgradePlan]:
+async def analyze_and_plan(args: CLIargs) -> UpgradePlan:
     """Analyze cloud and generate the upgrade plan with steps.
 
     :param args: CLI arguments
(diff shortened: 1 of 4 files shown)


cou/cli.py
```diff
@@ -137,9 +137,8 @@ async def get_upgrade_plan(args: CLIargs) -> None:
     :param args: CLI arguments
     :type args: CLIargs
     """
-    analysis_result, upgrade_plan = await analyze_and_plan(args)
+    _, upgrade_plan = await analyze_and_plan(args)
```
agileshaw, line 140: I don't think there's any other place needs the analysis_result output from `analyze_and_plan` function. We can change that function to only return `upgrade_plan` and you don't need to ignore an value here.

agileshaw (changes requested):

cou/cli.py
rgildein: True, I do not think that this is really related with desire goal of this fix, but I removed it.

aieri (approved):

agileshaw (changes requested):
Could you also update the docstring of `analyze_and_plan` function?

rgildein:
> `analyze_and_plan`

I do not like this redundancy (type hinting in docstring and real one).

gabrielcocenza (approved):

PR #263, opened 2024-02-27
Add Cinder OpenStackApplication

rgildein:
Draft PR, based on #264

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -340,10 +340,7 @@ class OpenStackApplication(COUApplication):
         ]
 
     def upgrade_steps(
-        self,
-        target: OpenStackRelease,
-        units: Optional[list[COUUnit]],
-        force: bool,
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]], force: bool
     ) -> list[UpgradeStep]:
         """Upgrade steps planning.
 
@@ -370,6 +367,7 @@ class OpenStackApplication(COUApplication):
             self._set_action_managed_upgrade(enable=bool(units)),
             self._get_upgrade_charm_step(target),
             self._get_change_install_repository_step(target),
+            self._get_units_upgrade_steps(units, force),
         ]
 
     def post_upgrade_steps(
@@ -423,6 +421,44 @@ class OpenStackApplication(COUApplication):
                 upgrade_steps.add_step(step)
         return upgrade_steps
 
+    def _get_unit_upgrade_steps(self, unit: COUUnit, force: bool) -> UnitUpgradeStep:
+        """Get the upgrade steps for single unit.
+
+        :param unit: Unit to generate upgrade steps
+        :type unit: COUUnit
+        :param force: Whether the unit step generation should be forced
+        :type force: bool
+        :return: Unit upgrade step
+        :rtype: UnitUpgradeStep
+        """
+        # pylint: disable=unused-argument
+        unit_plan = UnitUpgradeStep(description=f"Upgrade plan for unit: {unit.name}")
+        unit_plan.add_step(self._get_pause_unit_step(unit))
+        unit_plan.add_step(self._get_openstack_upgrade_step(unit))
+        unit_plan.add_step(self._get_resume_unit_step(unit))
+        return unit_plan
+
+    def _get_units_upgrade_steps(self, units: Optional[list[COUUnit]], force: bool) -> UpgradeStep:
+        """Get the upgrade steps for the units.
+
+        :param units: Units to generate upgrade steps
+        :type units: list[COUUnit]
+        :param force: Whether the plan generation should be forced
+        :type force: bool
+        :return: Upgrade step
+        :rtype: UpgradeStep
+        """
+        if not units:
+            logger.debug("units were not provided, skipping")
+            return UpgradeStep()
+
+        units_plan = UpgradeStep(
+            description=f"Upgrade plan for units: {', '.join([unit.name for unit in units])}",
+            parallel=True,
+        )
+        units_plan.sub_steps = [self._get_unit_upgrade_steps(unit, force) for unit in units]
+        return units_plan
+
     def _get_upgrade_current_release_packages_step(
         self, units: Optional[list[COUUnit]]
     ) -> PreUpgradeStep:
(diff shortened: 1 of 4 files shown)


cou/apps/base.py
```diff
@@ -526,41 +522,27 @@ def _get_upgrade_charm_step(self, target: OpenStackRelease) -> UpgradeStep:
...
-        :return: Step to disable action-managed-upgrade
-        :rtype: UpgradeStep
+        :param enable: enable or disable option
+        :type enable: bool
+        :return: Step to change action-managed-upgrade config option, if option exist.
+        :rtype: UnitUpgradeStep
         """
-        if self.config.get("action-managed-upgrade", {}).get("value", False):
+        if "action-managed-upgrade" not in self.config:
+            logger.debug("%s application has not action-managed-upgrade config option", self.name)
```
aieri, line 534:
```suggestion
            logger.debug("%s application doesn't have an action-managed-upgrade config option", self.name)
```

tests/unit/apps/test_core.py
```diff
@@ -342,7 +342,7 @@ def test_upgrade_plan_ussuri_to_victoria(model):
             coro=model.upgrade_charm(app.name, "ussuri/stable", switch=None),
         ),
         UpgradeStep(
-            description=f"Change charm config of '{app.name}' 'action-managed-upgrade' to False.",
+            description="Change 'action-managed-upgrade' config option to False.",
```
aieri, line 345: I would slightly prefer to keep the `app.name` in the description for consistency with the other steps, but it's not a blocker

tests/unit/apps/test_core.py
```diff
@@ -877,3 +877,101 @@ def _generate_nova_compute_app(model):
...
+        Upgrade software packages of 'cinder' from the current APT repositories
+            Upgrade software packages on unit cinder/0
+            Upgrade software packages on unit cinder/1
+            Upgrade software packages on unit cinder/2
+        Refresh 'cinder' to the latest revision of 'ussuri/stable'
+        Change 'action-managed-upgrade' config option to False.
+        Upgrade 'cinder' to the new channel: 'victoria/stable'
+        Change charm config of 'cinder' 'openstack-origin' to 'cloud:focal-victoria'
+        Wait 300s for app cinder to reach the idle state.
+        Check if the workload of 'cinder' has been upgraded on units: cinder/0, cinder/1, cinder/2
```
aieri, line 897: you're missing running the upgrade action; if these are the steps, the workload check will fail as no upgrade will have occurred

cou/apps/core.py
aieri, line 1: also not a blocker, but why is adding the Cinder class in scope for this PR? Couldn't the two things (Cinder class and improving the action-managed-upgrade functions) be kept separate?

tests/unit/apps/test_core.py
rgildein: I was thinking that we could remove name from all steps.
Current plan string:
```bash
Upgrade plan for 'nova-compute' to victoria
    Upgrade software packages of 'nova-compute' from the current APT repositories
        Upgrade software packages on unit nova-compute/0
        Upgrade software packages on unit nova-compute/1
        Upgrade software packages on unit nova-compute/2
    Refresh 'nova-compute' to the latest revision of 'ussuri/stable'
    Change charm config of 'nova-compute' 'action-managed-upgrade' to True.
    Upgrade 'nova-compute' to the new channel: 'victoria/stable'
    Change charm config of 'nova-compute' 'source' to 'cloud:focal-victoria'
```
what I think we should have (shorter lines, which will not creat line break):
```bash
Upgrade plan for 'nova-compute' to victoria
    Upgrade software packages from the current APT repositories
        Upgrade software packages on unit nova-compute/0
        Upgrade software packages on unit nova-compute/1
        Upgrade software packages on unit nova-compute/2
    Refresh to the latest revision of 'ussuri/stable' channel
    Change 'action-managed-upgrade' config option to True.
    Upgrade channel to the 'victoria/stable'
    Change 'source' config option to 'cloud:focal-victoria'
```
What do you say?

tests/unit/apps/test_core.py
rgildein: Yes, because I did not add any mores steps to Cinder, yet. Converting to draft + move fix for set `action-managed-upgrade` to #263

tests/unit/apps/test_core.py
rgildein: Now I realized that it will cause an issue in hypervisor upgrade plan, where we do not have the app placeholder, e.g. `Upgrade plan for 'nova-compute' to victoria`. 
Maybe we can add such placeholder to hypervisor planner and remove app name from steps.

tests/unit/apps/test_core.py
gabrielcocenza: I don't see the need of creating a cinder class right now. I think passing units to upgrade an OpenStackApplication should be able to run the pause, upgrade and resume actions of the units to upgrade.

tests/unit/apps/test_core.py
aieri: either is fine as long we're sure we're always printing the first line (`Upgrade plan for 'nova-compute' to victoria`)

cou/apps/base.py
```diff
@@ -423,6 +421,44 @@ def generate_upgrade_plan(
                 upgrade_steps.add_step(step)
         return upgrade_steps
 
+    def _get_unit_upgrade_steps(self, unit: COUUnit, force: bool) -> UnitUpgradeStep:
+        """Get the upgrade steps for single unit.
```
gabrielcocenza, line 425:
```suggestion
        """Get the upgrade steps for a single unit.
```

cou/apps/core.py
```diff
@@ -78,40 +75,31 @@ def upgrade_steps(
...
 
-        app_steps = super().upgrade_steps(target, units, force)
-        unit_steps = self._get_units_upgrade_steps(units, force)
-        return app_steps + unit_steps
+        return super().upgrade_steps(target, units, force)
 
-    def _get_units_upgrade_steps(self, units: list[COUUnit], force: bool) -> list[UpgradeStep]:
-        """Get the upgrade steps for the units.
+    def _get_unit_upgrade_steps(self, unit: COUUnit, force: bool) -> UnitUpgradeStep:
+        """Get the upgrade steps for single unit.
```
gabrielcocenza, line 81:
```suggestion
        """Get the upgrade steps for a single unit.
```

cou/apps/core.py
```diff
@@ -157,3 +145,8 @@ def _get_disable_scheduler_step(self, unit: COUUnit) -> UnitUpgradeStep:
                 unit_name=unit.name, action_name="disable", raise_on_failure=True
             ),
         )
+
+
+@AppFactory.register_application(["cinder"])
+class Cinder(OpenStackApplication):
+    """Cinder application."""
```
gabrielcocenza, line 152: I'm not seeing the need to create the Cinder class. I think we can remove it

gabrielcocenza (commented):
Generally LGTM. I think there is no need to create a cinder class so far

agileshaw (commented):
Mostly LGTM. I'm +1 on only adding a Cinder application class when it is really necessary. This PR can just focus on unit-by-unit upgrade

gabrielcocenza (approved):

aieri (approved):

PR #264, opened 2024-02-28
 Improve 'action-managed-upgrade' config handling

rgildein:
Merge logic of `_get_disable_action_managed_step` and `_get_enable_action_managed_step` to single function and fix not calling it if config option does not exist.

Previous we return the step to change 'action-managed-upgrade' even if option does not exist and we want
to enable it.


gabrielcocenza (approved):
LGTM

agileshaw (approved):
LGTM

PR #265, opened 2024-02-28
Test/bseng 2084/create snap for testing

rgildein:
THIS PR IS ONLY FOR TESTING PURPOSE.


rgildein:
We will create another one.

PR #262, opened 2024-02-27
Add specific test for NovaCompute upgrade plan

rgildein:
In unit tests of #251 we found out that generating upgrade plan for Nova-compute is missing some required steps, however I found out that I accidentally use OpenStackAPplication instead of NovaCompute application to generate those upgrade plans.

Current upgrade plan for NovaCompute application with 3 units:
```bash
Upgrade plan for 'nova-compute' to victoria
    Upgrade software packages of 'nova-compute' from the current APT repositories
        Upgrade software packages on unit nova-compute/0
        Upgrade software packages on unit nova-compute/1
        Upgrade software packages on unit nova-compute/2
    Refresh 'nova-compute' to the latest revision of 'ussuri/stable'
    Change charm config of 'nova-compute' 'action-managed-upgrade' to True.
    Upgrade 'nova-compute' to the new channel: 'victoria/stable'
    Change charm config of 'nova-compute' 'source' to 'cloud:focal-victoria'
    Upgrade plan for units: nova-compute/0, nova-compute/1, nova-compute/2
        Upgrade plan for unit: nova-compute/0
            Disable nova-compute scheduler from unit: 'nova-compute/0'.
            Check if unit nova-compute/0 has no VMs running before upgrading.
            ├── Pause the unit: 'nova-compute/0'.
            ├── Upgrade the unit: 'nova-compute/0'.
            ├── Resume the unit: 'nova-compute/0'.
            Enable nova-compute scheduler from unit: 'nova-compute/0'.
        Upgrade plan for unit: nova-compute/1
            Disable nova-compute scheduler from unit: 'nova-compute/1'.
            Check if unit nova-compute/1 has no VMs running before upgrading.
            ├── Pause the unit: 'nova-compute/1'.
            ├── Upgrade the unit: 'nova-compute/1'.
            ├── Resume the unit: 'nova-compute/1'.
            Enable nova-compute scheduler from unit: 'nova-compute/1'.
        Upgrade plan for unit: nova-compute/2
            Disable nova-compute scheduler from unit: 'nova-compute/2'.
            Check if unit nova-compute/2 has no VMs running before upgrading.
            ├── Pause the unit: 'nova-compute/2'.
            ├── Upgrade the unit: 'nova-compute/2'.
            ├── Resume the unit: 'nova-compute/2'.
            Enable nova-compute scheduler from unit: 'nova-compute/2'.
    Wait 1800s for model test_model to reach the idle state.
[...]

PR #266, opened 2024-02-29
Chore/no ref/improve os app

rgildein:
Fix unitests, which was broken after merging #262 and add better string representation of OpenStackApplication. This is used currently only in logging, but it's important for to be able to reproduce on what cloud COU was runned.


aieri (approved):

agileshaw (approved):
LGTM

PR #270, opened 2024-03-01
Limit the plan generation of Hypervisor to the machines passed.

gabrielcocenza:
- currently the hypervisor is generating the plan for all hypervisor machines not considering the cli filter that can be specific to some machines or azs
- changed the eq method of HypervisorGroup to also compare the app_units

### DIFF
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -157,10 +157,12 @@ class HypervisorUpgradePlanner:
             for unit in app.units.values():
                 # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
                 #                 belong to a single group.
-                machine = unit.machine
-                if machine in self.machines:
-                    az = machine.az or ""
-                    azs[az].app_units[app.name].append(unit)
+                if unit.machine not in self.machines:
+                    logger.debug("skipping machine %s", unit.machine.machine_id)
+                    continue
+
+                az = unit.machine.az or ""
+                azs[az].app_units[app.name].append(unit)
 
         return azs
 
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -155,12 +155,12 @@ class HypervisorUpgradePlanner:
         azs = AZs()
         for app in self.apps:
             for unit in app.units.values():
-                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
-                #                 belong to a single group.
                 if unit.machine not in self.machines:
                     logger.debug("skipping machine %s", unit.machine.machine_id)
                     continue
 
+                # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
+                #                 belong to a single group.
                 az = unit.machine.az or ""
                 azs[az].app_units[app.name].append(unit)
 
--- a/cou/steps/hypervisor.py
+++ b/cou/steps/hypervisor.py
@@ -26,7 +26,7 @@ from cou.steps import (
     UpgradePlan,
     UpgradeStep,
 )
-from cou.utils.juju_utils import COUUnit
+from cou.utils.juju_utils import COUMachine, COUUnit
 from cou.utils.openstack import OpenStackRelease
 
 logger = logging.getLogger(__name__)
@@ -53,7 +53,7 @@ class HypervisorGroup:
         if not isinstance(other, HypervisorGroup):
             return NotImplemented
 
-        return other.name == self.name
+        return other.name == self.name and other.app_units == self.app_units
 
 
 class AZs(defaultdict):
@@ -95,15 +95,18 @@ class HypervisorUpgradePlanner:
     This planner is meant to be used to upgrade machines contains the nova-compute application.
     """
 
-    def __init__(self, apps: list[OpenStackApplication]) -> None:
+    def __init__(self, apps: list[OpenStackApplication], machines: list[COUMachine]) -> None:
         """Initialize the Hypervisor class.
 
         The application should be sorted by upgrade order.
 
         :param apps: sorted list of OpenStack applications
         :type apps: list[OpenStackApplication]
+        :param machines: Hypervisor machines to generate upgrade plan.
+        :type machines: list[COUMachine]
         """
         self._apps = apps
+        self._machines = machines
 
     @property
     def apps(self) -> list[OpenStackApplication]:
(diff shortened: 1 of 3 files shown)


cou/steps/hypervisor.py
```diff
@@ -145,8 +157,10 @@ def azs(self) -> AZs:
             for unit in app.units.values():
                 # NOTE(rgildein): If there is no AZ, we will use empty string and all units will
                 #                 belong to a single group.
-                az = unit.machine.az or ""
-                azs[az].app_units[app.name].append(unit)
+                machine = unit.machine
+                if machine in self.machines:
```
rgildein, line 161: not-blocking: we are not using machine anywhere else, so I think we should use `if unit.machine in self.machines:`.

cou/steps/hypervisor.py
```diff
@@ -145,8 +157,10 @@ def azs(self) -> AZs:
...
+                    az = machine.az or ""
```
rgildein, line 162: The note above should be moved to this part, since it belong to it.

cou/steps/hypervisor.py
rgildein, line 161: I think we should log skipped machine, so I'm maybe doing something like this:

```python
for unit in app.units.values():
    if unit.machine not in self.machines:
        logger.debug("skipping machine %s", unit.machine)
        continue

    az = ...
```
What do you say?

cou/steps/plan.py
```diff
@@ -333,45 +333,38 @@ def _generate_control_plane_plan(
...
+    for app in analysis_result.apps_data_plane:
         if app.charm == "ceph-osd":
             # ceph-osd will not be handled by hypervisor planner
             continue
 
         for machine in app.machines:
-            if machine in nova_compute_machines:
+            if machine in hypervisors_machines:
                 hypervisor_apps.append(app)
                 break  # exiting machine for loop
```
rgildein, line 364: I think we do not need this anymore, since we are already going through `analysis_result.apps_data_plane`, or not? We only need to drop `ceph-osd` form those apps?

tests/unit/steps/test_hypervisor.py
```diff
@@ -53,10 +54,11 @@ def test_generate_post_upgrade_steps():
     """Test generating of post-upgrade steps."""
     target = OpenStackRelease("victoria")
     units = ["1", "2", "3"]
+    machines = [COUMachine(f"{i}", (), f"az{i//2}") for i in range(3)]
```
rgildein, line 57:
```suggestion
    machines = [COUMachine(f"{i}", (), f"az{i}") for i in range(3)]
```

tests/unit/steps/test_hypervisor.py
```diff
@@ -37,11 +37,12 @@ def test_generate_pre_upgrade_steps():
     """Test generating of pre-upgrade steps."""
     target = OpenStackRelease("victoria")
     units = ["1", "2", "3"]
+    machines = [COUMachine(f"{i}", (), f"az{i//2}") for i in range(3)]
```
rgildein, line 40:
```suggestion
    machines = [COUMachine(f"{i}", (), f"az{i}") for i in range(3)]
```

tests/unit/steps/test_hypervisor.py
```diff
@@ -69,7 +71,7 @@ def test_hypervisor_group():
     group1 = HypervisorGroup("test", {"app1": []})
     group2 = HypervisorGroup("test", {"app2": []})
 
-    assert group1 == group2
+    assert group1 != group2
```
rgildein, line 74: Can you create group3, which will be same as group1? And compare them.

tests/unit/steps/test_hypervisor.py
```diff
@@ -268,7 +286,95 @@ def test_hypervisor_upgrade_plan(model):
...
     )
 
-    planner = HypervisorUpgradePlanner([cinder, nova_compute])
+    planner = HypervisorUpgradePlanner([cinder, nova_compute], list(machines.values()))
+    plan = planner.generate_upgrade_plan(target, False)
+
+    assert str(plan) == exp_plan
+
+
+def test_hypervisor_upgrade_plan_filtered(model):
```
rgildein, line 295: not-blocking: `test_hypervisor_upgrade_plan_filtered` -> `test_hypervisor_upgrade_plan_single_machine` or maybe even drop `test_hypervisor_plan_single_unit` (Meaning we need to do same for test above).

rgildein (changes requested):
Overall it's looks good, only some small changes required.

cou/steps/plan.py
gabrielcocenza: Maybe in the future we'll also need to drop subordinates, but I guess it's ok drooping now :+1:

cou/steps/hypervisor.py
```diff
@@ -157,10 +157,12 @@ def azs(self) -> AZs:
...
                 #                 belong to a single group.
-                machine = unit.machine
-                if machine in self.machines:
-                    az = machine.az or ""
-                    azs[az].app_units[app.name].append(unit)
+                if unit.machine not in self.machines:
+                    logger.debug("skipping machine %s", unit.machine.machine_id)
+                    continue
+
+                az = unit.machine.az or ""
```
rgildein, line 164: not-blocking: lines 158-159 should be moved before this line

cou/steps/plan.py
rgildein: yes, we will need to do that, but now it's fine

rgildein (approved):

rgildein (approved):

cou/steps/plan.py
```diff
@@ -333,45 +333,35 @@ def _generate_control_plane_plan(
...
+    for app in analysis_result.apps_data_plane:
         if app.charm == "ceph-osd":
             # ceph-osd will not be handled by hypervisor planner
             continue
 
-        for machine in app.machines:
-            if machine in nova_compute_machines:
-                hypervisor_apps.append(app)
-                break  # exiting machine for loop
+        hypervisor_apps.append(app)
```
agileshaw, line 361: This can simply be:
```
hypervisor_apps = [app for app in analysis_result.apps_data_plane if app.charm != "ceph-osd"]
```

agileshaw (changes requested):
Overall looks good, just one suggestion.

cou/steps/plan.py
rgildein: There is a huge chance that we will need to filter more than just ceph-osd, so I do not think that this should be blocked by this. Resolving this conversation.

agileshaw: The suggested approach can handle additional filtering as well. But I'm ok to unblock this PR with the current approach.

agileshaw (approved):

rgildein (approved):

PR #261, opened 2024-02-27
Add check for enable-auto-restarts config option

rgildein:
Add sanity checks for generating upgrade plan for application, where we can check all relevant pre-checks, before generating upgrade plan. Add _check_auto_restarts to verify that the `enable-auto-restarts` config option is disabled.

PR #274, opened 2024-03-04
Moved MismatchedOpenStackVersions to Application

gabrielcocenza:
- Applications give the lowest OpenStack version on units
- verify that control-plane apps units are running the same version


cou/apps/base.py
```diff
@@ -243,33 +239,26 @@ def target_channel(self, target: OpenStackRelease) -> str:
         return f"{target.codename}/stable"
 
     @property
-    def current_os_release(self) -> OpenStackRelease:
-        """Current OpenStack Release of the application.
+    def os_release_units(self) -> defaultdict[OpenStackRelease, list[str]]:
```
rgildein, line 242: This should not return default dict, but rather just dict. It does not make sense if I call `os_release_units.os_release_units["no-real-unit/99"]` and it works. I think that maybe even tuple of tuple (e.g. `(("unit/0", "1.1"), ("unit/1", "1.2"), )`) will be better.

cou/steps/plan.py
```diff
@@ -84,6 +86,33 @@ def verify_supported_series(analysis_result: Analysis) -> None:
         )
 
 
+def verify_os_versions_control_plane(apps: list[OpenStackApplication]) -> None:
```
rgildein, line 89: This does not make any sense here, it should be part of the application and called by `upgrade_plan_sanity_checks` introduced in #261.

Like this we need to integrate over apps (we are doing that when we call generate upgrade plan), check that for some app it should be skipped (for subordinate we are skipping it).

rgildein (changes requested):

cou/apps/base.py
```diff
@@ -793,3 +791,28 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
                 f"Application '{self.name}' already configured for release equal to or greater "
                 f"than {target}. Ignoring."
             )
+
+    def _check_mismatched_versions(self, units: Optional[list[COUUnit]]) -> None:
+        """Check that there are no mismatched versions on app units.
+
+        If units are passed that means that the application will upgrade unit by unit,
+        so there are no mismatch.
```
rgildein, line 799: I thought that we had another discussion yesterday, if we have unit by unit upgrade we will check those units and not raise an exception, but just warning.

Now I realized that the warning is maybe not good idea, since NovaCompute will be always upgrade unit by unit, so we should check all units provided. I know that, like this, a user can hit an issue when trying to upgrade AZ-xyz where one unit is already upgraded. The question is what should happen if I do
```bash
cou upgrade hypervisors --az az-xyz
```
and 1 of 3 nova-compute unit is in different version. This can happen when a user previously targeted only a single machine or when other has no-empty hypervisors.

gabrielcocenza: IMO, the responsibility of selecting a unit or not depending on the target passed should happen in the [filter_hypervisors_machines](https://github.com/canonical/charmed-openstack-upgrader/blob/dev/data-plane/cou/steps/plan.py#L432). There we can have a logging saying that unit/machine x was not selected because it's already upgraded. What do you think?

rgildein: What exactly you want to move there? The logic of skipping nova-compute with different version? What about cinder?

gabrielcocenza: When you upgrade one unit the other that is collocated also gets upgraded, but I see your point...
Whatever method we implement to not generate plan for units that are already upgraded, I think it's a separate thing from this mismatched check. I don't see value on checking this on applications that upgrade unit by unit because it's an expected thing to happen.

rgildein (approved):

cou/apps/base.py
agileshaw, line 799: I don't think it is correct to say there's no mismatch in unit-by-unit upgrade. I suggest reword this to:
```suggestion
        If units are passed, the application will upgrade in unit-by-unit fashion,
        and mismatch is not checked.
```

agileshaw (commented):
Changes LGTM, just one suggestion on rewording docstring

agileshaw (approved):
LGTM, thanks :)

PR #275, opened 2024-03-05
Add cache property for applications to COUModel

rgildein:
Like this it will be possible to access all application in model, without calling get_applications. This will be used in planning stage, where we are not running async functions and for any app, which require using other applications. e.g. ceph-osd need to check nova-compute status.

With this feature, we can easily access all colocated nova-compute and ceph-mon during pre- and post-upgrade steps in ceph-osd.

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -201,6 +201,22 @@ class COUModel:
         self._juju_data = FileJujuData()
         self._model = Model(max_frame_size=JUJU_MAX_FRAME_SIZE, jujudata=self.juju_data)
         self._name = name
+        self._applications: dict[str, COUApplication] | None = None
+
+    @property
+    def applications(self) -> dict[str, COUApplication]:
+        """Return cached applications.
+
+        This property represents the cached output of the get_applications function.
+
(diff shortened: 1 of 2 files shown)


cou/utils/juju_utils.py
```diff
@@ -201,6 +201,22 @@ def __init__(self, name: Optional[str]):
...
+        """Return cached applications.
+
+        This property represents the cached output of the get_applications function.
+
+        :return: dictionary of COUApplication
+        :rtype: dict[str, COUApplication]
+        :raises ValueError: When get_applications has not yet been called.
+        """
+        if self._applications is None:
+            raise ValueError("The get_applications has not yet been called.")
```
gabrielcocenza, line 217: Why not call `get_applications` here instead of raising an error?

rgildein: async?

gabrielcocenza: Forgot about that :sweat_smile:

gabrielcocenza (approved):

rgildein:
Please check #280

agileshaw (approved):
LGTM

PR #276, opened 2024-03-05
Add relations parameter to COUApplication

rgildein:
The relations obtained from application status can be easily used to get name of related applications. For example, we will use it in ceph-osd to get related ceph-mon app.
e.g. `ceph_mon.relations == {'mon': ['ceph-mon']}`


rgildein:
We abandon this PR, after discussion about ceph-osd and require-osd-release, which will be part of the cloud upgrade plan instead of the ceph-osd application.

PR #279, opened 2024-03-06
Add post-upgrade steps for whole upgrade plan

rgildein:
Adding ceph-mon post-upgrade step to ensure require-osd-release option matches with ceph-osd version. Add also add_steps function to BaseStep so we can easily add more steps at once.

### DIFF
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -251,6 +251,15 @@ class BaseStep:
 
         self._sub_steps.append(step)
 
+    def add_steps(self, steps: Iterable[BaseStep]) -> None:
+        """Add multiple steps.
+
+        :param steps: Sequence of BaseStep to be added as sub step.
+        :type steps: Iterable[BaseStep]
+        """
+        for step in steps:
+            self.add_step(step)
+
     def cancel(self, safe: bool = True) -> None:
         """Cancel step and all its sub steps.
 
--- a/cou/steps/__init__.py
+++ b/cou/steps/__init__.py
@@ -254,7 +254,7 @@ class BaseStep:
     def add_steps(self, steps: Iterable[BaseStep]) -> None:
         """Add multiple steps.
 
-        :param steps: Sequence of BaseStep to be added as sub step.
+        :param steps: Sequence of BaseStep to be added as sub steps.
         :type steps: Iterable[BaseStep]
         """
         for step in steps:
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -41,12 +41,14 @@ from cou.exceptions import (
     HaltUpgradePlanGeneration,
     HighestReleaseAchieved,
     NoTargetError,
+    NotSupported,
     OutOfSupportRange,
 )
-from cou.steps import PreUpgradeStep, UpgradePlan
+from cou.steps import PostUpgradeStep, PreUpgradeStep, UpgradePlan
 from cou.steps.analyze import Analysis
 from cou.steps.backup import backup
 from cou.steps.hypervisor import HypervisorUpgradePlanner
+from cou.utils.app_utils import set_require_osd_release_option
 from cou.utils.juju_utils import DEFAULT_TIMEOUT, COUMachine
 from cou.utils.nova_compute import get_empty_hypervisors
 from cou.utils.openstack import LTS_TO_OS_RELEASE, OpenStackRelease
@@ -304,7 +306,10 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     pre_plan_sanity_checks(args, analysis_result)
     target = determine_upgrade_target(analysis_result)
 
-    plan = generate_common_plan(target, analysis_result, args)
+    plan = UpgradePlan(
+        f"Upgrade cloud from '{analysis_result.current_cloud_os_release}' to '{target}'"
+    )
+    plan.add_steps(_get_pre_upgrade_steps(analysis_result, args))
 
     # NOTE (gabrielcocenza) upgrade group as None means that the user wants to upgrade
     #  the whole cloud.
@@ -315,27 +320,22 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     if args.upgrade_group in {DATA_PLANE, HYPERVISORS, None}:
         plan.sub_steps.extend(await _generate_data_plane_plan(target, analysis_result, args))
 
+    plan.add_steps(_get_post_upgrade_steps(analysis_result, args))
+
     return plan
 
 
-def generate_common_plan(
-    target: OpenStackRelease, analysis_result: Analysis, args: CLIargs
-) -> UpgradePlan:
-    """Generate the common upgrade plan.
+def _get_pre_upgrade_steps(analysis_result: Analysis, args: CLIargs) -> list[PreUpgradeStep]:
+    """Get the pre-upgrade steps.
 
-    :param target: Target OpenStack release
-    :type target: OpenStackRelease
     :param analysis_result: Analysis result
     :type analysis_result: Analysis
     :param args: CLI arguments
     :type args: CLIargs
-    :return: Common upgrade plan
-    :rtype: UpgradePlan
+    :return: List of pre-upgrade steps.
+    :rtype: list[PreUpgradeStep]
     """
-    plan = UpgradePlan(
-        f"Upgrade cloud from '{analysis_result.current_cloud_os_release}' to '{target}'"
-    )
-    plan.add_step(
+    steps = [
         PreUpgradeStep(
             description="Verify that all OpenStack applications are in idle state",
             parallel=False,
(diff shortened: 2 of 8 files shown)


cou/steps/__init__.py
```diff
@@ -251,6 +251,15 @@ def add_step(self, step: BaseStep) -> None:
 
         self._sub_steps.append(step)
 
+    def add_steps(self, steps: Iterable[BaseStep]) -> None:
```
rgildein, line 254: Using `plan.sub_steps.extend` is working fine, but that will not skip an empty step in list of steps.

cou/steps/plan.py
```diff
@@ -349,15 +349,61 @@ def generate_common_plan(
...
+    ]
+    match num_ceph_mon_apps := len(ceph_mons_apps):
+        case 0:
+            return PostUpgradeStep()
+        case 1:
+            app = ceph_mons_apps[0]  # getting first and only one ceph-mon app
+            unit = list(app.units.values())[0]  # getting first unit, sice we do not care which one
+            return PostUpgradeStep(
+                "Ensure require-osd-release option matches with ceph-osd version",
+                coro=set_require_osd_release_option(unit.name, app.model),
```
rgildein, line 401: I originally used `_get_required_osd_release` (command `ceph osd dump -f json`) and then command `f"ceph osd require-osd-release {target}` to set it. My question is behavior of `set_require_osd_release_option` is same or not, I think yes.
^ @agileshaw

agileshaw: `set_require_osd_release_option` also checks the actual release value of osds as well as find mismatches between the osds. The value of `require-osd-release` will only be changed if all osds are on the same release which differs from `require-osd-release`. 

I think we should use `set_require_osd_release_option` which is more fault-proof.

gabrielcocenza, line 401: As we discussed on COU sync, we can `set_require_osd_release_option` for all ceph-mon applications

tests/unit/steps/test_steps.py
```diff
@@ -251,6 +251,17 @@ def test_step_add_step_failed():
         plan.add_step(MagicMock())
 
 
+def test_step_add_steps():
+    """Test BaseStep adding sub steps at once."""
```
gabrielcocenza, line 255: I think it would be nice to have a text showing that an empty step is not added

tests/unit/steps/test_steps_plan.py
```diff
@@ -743,19 +743,122 @@ def test_generate_common_plan(cli_backup, cli_args, model):
...
             )
         )
 
-    assert common_plan == expected_plan
+    pre_upgrade_steps = cou_plan._get_pre_upgrade_steps(mock_analysis_result, cli_args)
+
+    assert pre_upgrade_steps == expected_steps
+
+
+@pytest.mark.parametrize("upgrade_group", ["control-plane", "hypervisor"])
```
gabrielcocenza, line 760: Please use the global variables instead of those control-plane and hypervisors strings

cou/steps/plan.py
```diff
@@ -349,15 +349,61 @@ def generate_common_plan(
...
+    :rtype: list[PreUpgradeStep]
+    """
+    steps = []
+    if args.upgrade_group in {DATA_PLANE, None}:
+        steps.append(_get_ceph_mon_post_upgrade_step(analysis_result))
+
+    return steps
+
+
+def _get_ceph_mon_post_upgrade_step(analysis_result: Analysis) -> PostUpgradeStep:
```
gabrielcocenza, line 381: nit: if we were using `analysis_result` to access more things I think it would make sense to pass as parameter. In this case I think it can be control_plane_apps directly

tests/unit/steps/test_steps_plan.py
```diff
@@ -743,19 +743,122 @@ def test_generate_common_plan(cli_backup, cli_args, model):
...
+    args = MagicMock(spec_set=CLIargs)()
+    args.upgrade_group = upgrade_group
+
+    pre_upgrade_steps = cou_plan._get_post_upgrade_steps(MagicMock(), args)
+
+    assert pre_upgrade_steps == []
+    mock_get_ceph_mon_post_upgrade_step.assert_not_called()
+
+
+@pytest.mark.parametrize("upgrade_group", ["data-plane", None])
```
gabrielcocenza, line 773: same here. Better use the global variables

gabrielcocenza (changes requested):

tests/unit/steps/test_steps.py
rgildein: What do you mean by that? Like in docstring of test? I'm not testing that empty step is not added.

cou/steps/plan.py
```diff
@@ -349,15 +349,62 @@ def generate_common_plan(
...
+    :raises NotSupported: When two ceph-mon are found at the model.
+    """
+    ceph_mons_apps = [app for app in apps if isinstance(app, CephMon)]
+    if num_ceph_mon_apps := len(ceph_mons_apps) > 1:
+        raise NotSupported(
+            f"Deployment with {num_ceph_mon_apps} ceph-mon applications is not supported."
+        )
+
+    steps = []
+    for app in ceph_mons_apps:
```
agileshaw, line 398: Making steps a list and also iterating through `ceph_mons_apps` doesn't look like a good approach because it gives user a false impression that more than one ceph-mon apps in deployment can be handled and will generate multiple steps, where in fact steps can only be empty or with 1 step added.

agileshaw (changes requested):

cou/steps/plan.py
```diff
@@ -349,15 +349,62 @@ def generate_common_plan(
...
+
+
+def _get_ceph_mon_post_upgrade_steps(apps: list[OpenStackApplication]) -> list[PostUpgradeStep]:
+    """Get the post-upgrade step for ceph-mon, where we check the require-osd-release option.
+
+    :param apps: List of OpenStackApplication.
+    :type apps: list[OpenStackApplication]
+    :return: The post-upgrade step.
+    :rtype: list[PreUpgradeStep]
+    :raises NotSupported: When two ceph-mon are found at the model.
```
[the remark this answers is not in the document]
rgildein: Actually this should be removed.

cou/steps/plan.py
```diff
@@ -349,15 +349,62 @@ def generate_common_plan(
...
+        raise NotSupported(
+            f"Deployment with {num_ceph_mon_apps} ceph-mon applications is not supported."
+        )
+
+    steps = []
+    for app in ceph_mons_apps:
+        unit = list(app.units.values())[0]  # getting first unit, sice we do not care which one
+        steps.append(
+            PostUpgradeStep(
+                "Ensure require-osd-release option matches with ceph-osd version",
```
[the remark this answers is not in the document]
rgildein: This is copied from CephMon application, so I will change it there also. We are suing description with and without dot at the end, and I think we should aligned that.

cou/steps/plan.py
rgildein: Actually we will handle multiple ceph-mon, I just forgot to remove that raise.

tests/unit/steps/test_steps_plan.py
```diff
@@ -743,19 +742,93 @@ def test_generate_common_plan(cli_backup, cli_args, model):
...
+        series="focal",
+        subordinate_to=[],
+        units=units,
+        workload_version="17.0.1",
+    )
+
+    exp_steps = 2 * [
+        PostUpgradeStep(
+            "Ensure the 'require-osd-release' option matches the 'ceph-osd' version",
+            coro=app_utils.set_require_osd_release_option("cpeh-mon/0", model),
```
gabrielcocenza, line 825: I'm confused why the unit test didn't fail with this typo. Maybe we have a bug on `compare_step_coroutines`
```suggestion
            coro=app_utils.set_require_osd_release_option("ceph-mon/0", model),
```

gabrielcocenza (changes requested):

tests/unit/steps/test_steps_plan.py
```diff
@@ -743,19 +742,93 @@ def test_generate_common_plan(cli_backup, cli_args, model):
...
+    pre_upgrade_steps = cou_plan._get_post_upgrade_steps(MagicMock(), args)
+
+    assert pre_upgrade_steps == []
+    mock_get_ceph_mon_post_upgrade_steps.assert_not_called()
+
+
+@pytest.mark.parametrize("upgrade_group", [DATA_PLANE, None])
+@patch("cou.steps.plan._get_ceph_mon_post_upgrade_steps")
+def test_get_post_upgrade_steps_ceph_mon(mock_get_ceph_mon_post_upgrade_steps, upgrade_group):
+    """Test get post upgrad steps including ceph-mon."""
```
agileshaw, line 775:
```suggestion
    """Test get post upgrade steps including ceph-mon."""
```

agileshaw (commented):
One more typo

agileshaw (approved):
LGTM, thanks

rgildein:
> Please fix the CI before merging :)

Those were not failing before, and it seems as issue not related with this PR.

PR #280, opened 2024-03-06
Add pre-upgrade step to verify nova-compute

rgildein:
Add pre-upgrade step for CephOsd, which will check nova-compute upgrade status.

raleted: #275

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -25,6 +25,7 @@ from cou.utils.juju_utils import COUUnit
 from cou.utils.openstack import (
     OPENSTACK_TO_TRACK_MAPPING,
     TRACK_TO_OPENSTACK_MAPPING,
+    OpenStackCodenameLookup,
     OpenStackRelease,
 )
 
@@ -204,10 +205,74 @@ class MysqlInnodbCluster(AuxiliaryApplication):
     wait_timeout = LONG_IDLE_TIMEOUT
 
 
-# NOTE (gabrielcocenza): Although CephOSD class is empty now, it will be
-# necessary to add post upgrade plan to set require-osd-release. Registering on
-# AuxiliaryApplication can be easily forgot and ceph-osd can't be instantiated
-# as a normal OpenStackApplication.
 @AppFactory.register_application(["ceph-osd"])
 class CephOSD(AuxiliaryApplication):
     """Application for ceph-osd."""
+
+    def _are_collocated(self, units: list[COUUnit], charm: str) -> bool:
+        """Check if ceph-osd units are collocated with charm.
+
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[COUUnit]]
+        :param charm: Name of charm
+        :type charm: str
+        :return: Boolean if ceph-osd is colocated with charm.
+        :rtype: bool
+        """
+        return any(
+            # pylint: disable=no-member
+            self.model.applications[app].charm == charm  # type: ignore[attr-defined]
+            for unit in units
+            for app in unit.machine.apps
+        )
+
+    def pre_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]]
+    ) -> list[PreUpgradeStep]:
+        """Pre Upgrade steps planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[COUUnit]]
+        :return: List of pre upgrade steps.
+        :rtype: list[PreUpgradeStep]
+        """
+        steps = super().pre_upgrade_steps(target, units)
+        if units is not None and self._are_collocated(units, "nova-compute"):
+            steps.append(
+                PreUpgradeStep(
+                    description="Check if colocated nova-compute units had been upgraded.",
+                    coro=self._verify_nova_compute(target, units),
+                )
+            )
+        return steps
+
+    async def _verify_nova_compute(self, target: OpenStackRelease, units: list[COUUnit]) -> None:
+        """Check if a nova-compute application has upgraded its workload version.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to check if got upgraded
+        :type units: list[COUUnit]
+        :raises ApplicationError: When collocated nova-compute apps workload version isn't reached.
+        """
+        units_not_upgraded = []
+        apps = await self.model.get_applications()  # reload all applications information
+
+        for unit in units:
+            for name in unit.machine.apps:
+                app = apps[name]
+                if app.charm == "nova-compute":
+                    for nova_unit in app.units:
+                        compatible_os_versions = OpenStackCodenameLookup.find_compatible_versions(
+                            app.charm, nova_unit.workload_version
+                        )
+
+                        if target not in compatible_os_versions:
+                            units_not_upgraded.append(unit.name)
+
+        if units_not_upgraded:
+            raise ApplicationError(
+                f"Units '{', '.join(units_not_upgraded)}' did not reach {target}."
+            )
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -206,7 +206,7 @@ class MysqlInnodbCluster(AuxiliaryApplication):
 
 
 @AppFactory.register_application(["ceph-osd"])
-class CephOSD(AuxiliaryApplication):
+class CephOsd(AuxiliaryApplication):
     """Application for ceph-osd."""
 
     def pre_upgrade_steps(
@@ -223,7 +223,7 @@ class CephOSD(AuxiliaryApplication):
         """
         steps = [
             PreUpgradeStep(
-                description="Check if colocated nova-compute units had been upgraded.",
+                description="Check if all nova-compute units had been upgraded",
                 coro=self._verify_nova_compute(target),
             )
         ]
@@ -245,7 +245,7 @@ class CephOSD(AuxiliaryApplication):
                 logger.debug("skipping application %s", app.name)
                 continue
 
-            for unit in app.units:
+            for unit in app.units.values():
                 compatible_os_versions = OpenStackCodenameLookup.find_compatible_versions(
                     app.charm, unit.workload_version
                 )
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -209,23 +209,6 @@ class MysqlInnodbCluster(AuxiliaryApplication):
 class CephOSD(AuxiliaryApplication):
     """Application for ceph-osd."""
 
-    def _are_collocated(self, units: list[COUUnit], charm: str) -> bool:
-        """Check if ceph-osd units are collocated with charm.
-
-        :param units: Units to generate upgrade plan
-        :type units: Optional[list[COUUnit]]
-        :param charm: Name of charm
-        :type charm: str
-        :return: Boolean if ceph-osd is colocated with charm.
-        :rtype: bool
-        """
-        return any(
-            # pylint: disable=no-member
-            self.model.applications[app].charm == charm  # type: ignore[attr-defined]
-            for unit in units
-            for app in unit.machine.apps
-        )
-
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[COUUnit]]
     ) -> list[PreUpgradeStep]:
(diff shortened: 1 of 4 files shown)


cou/apps/auxiliary.py
```diff
@@ -204,10 +205,74 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+                            app.charm, nova_unit.workload_version
+                        )
+
+                        if target not in compatible_os_versions:
+                            units_not_upgraded.append(unit.name)
+
+        if units_not_upgraded:
+            raise ApplicationError(
+                f"Units '{', '.join(units_not_upgraded)}' did not reach {target}."
+            )
```
gabrielcocenza, line 278: On theory, I think this shouldn't be necessary because every application is responsible to check if it got upgraded or not. E.g: we don't check on keystone if ceph-mon units got upgraded.

However, there is more complexity in this case because the user can run something like `cou upgrade` and some hypervisors are not empty so it doesn't get upgraded. I see some possible solutions for this case:

1. We could check if the hypervisors_machines selected are enough to finish the nova-compute upgrade and then generate ceph-osd or not depending on this. This might be more complex to implement, but I think it will generate a cleaner plan
2. Have this check as post upgrade step on nova-compute class. Generate this post-upgrade depending on the command. E.g: `cou upgrade hypervisors` should not have it
3. Add on ceph-osd as pre-strep as you are doing.

I prefer option `1`, but if we keep the current approach `3` I don't see the need to check if it's colocated or not. We can have a simple check to see if nova-compute units got upgraded. It's indifferent if they are together in the same machine in the upgrade sequence. Moreover, I think this step should be the first one and the others should show a dependency from it.

gabrielcocenza (changes requested):

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,74 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
 
-# NOTE (gabrielcocenza): Although CephOSD class is empty now, it will be
-# necessary to add post upgrade plan to set require-osd-release. Registering on
-# AuxiliaryApplication can be easily forgot and ceph-osd can't be instantiated
-# as a normal OpenStackApplication.
 @AppFactory.register_application(["ceph-osd"])
 class CephOSD(AuxiliaryApplication):
     """Application for ceph-osd."""
+
+    def _are_collocated(self, units: list[COUUnit], charm: str) -> bool:
```
aieri, line 212:
```suggestion
    def _are_colocated(self, units: list[COUUnit], charm: str) -> bool:
```
and elsewhere

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+                logger.debug("skipping application %s", app.name)
+                continue
+
+            for unit in app.units:
+                compatible_os_versions = OpenStackCodenameLookup.find_compatible_versions(
+                    app.charm, unit.workload_version
+                )
+
+                if target not in compatible_os_versions:
+                    units_not_upgraded.append(unit.name)
```
gabrielcocenza, line 254:
```suggestion
            if len(self.os_release_units.keys()) > 1:
                units_not_upgraded = [units for os_version, units in  self.os_release_units.values if os_version != target]
                raise ApplicationError(
                f"Units '{', '.join(units_not_upgraded)}' did not reach {target}."
            )
            
```

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
-# NOTE (gabrielcocenza): Although CephOSD class is empty now, it will be
-# necessary to add post upgrade plan to set require-osd-release. Registering on
-# AuxiliaryApplication can be easily forgot and ceph-osd can't be instantiated
-# as a normal OpenStackApplication.
 @AppFactory.register_application(["ceph-osd"])
 class CephOSD(AuxiliaryApplication):
     """Application for ceph-osd."""
+
+    def pre_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]]
```
gabrielcocenza, line 213:
```suggestion
        self, target: OpenStackRelease, units: Optional[list[COUUnit]] = None
```

gabrielcocenza (changes requested):
I'm ok with the current approach and the suggestions are not critical.

I think it would be nice to add a custom `upgrade_plan_sanity_checks` that checks if units were passed and in that case raise an Exception  saying that ceph-osd doesn't accept being upgraded unit-by-unit

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+                    app.charm, unit.workload_version
+                )
+
+                if target not in compatible_os_versions:
+                    units_not_upgraded.append(unit.name)
+
+        if units_not_upgraded:
+            raise ApplicationError(
+                f"Units '{', '.join(units_not_upgraded)}' did not reach {target}."
+            )
```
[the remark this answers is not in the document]
gabrielcocenza: In terms of usability I think it's nice for the user knowing which units need to be upgraded, so I'm in favor of keeping it

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+            if app.charm != "nova-compute":
+                logger.debug("skipping application %s", app.name)
+                continue
+
+            for unit in app.units:
+                compatible_os_versions = OpenStackCodenameLookup.find_compatible_versions(
+                    app.charm, unit.workload_version
+                )
+
+                if target not in compatible_os_versions:
```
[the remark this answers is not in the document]
gabrielcocenza: Yes, the correct form would be using the max, although it wouldn't make difference on nova-compute. However, `os_release_units` property already has all this information compiled in the right form and I think it's easier to use it.

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[COUUnit]]
+        :return: List of pre upgrade steps.
+        :rtype: list[PreUpgradeStep]
+        """
+        steps = [
+            PreUpgradeStep(
+                description="Check if colocated nova-compute units had been upgraded.",
```
rgildein, line 226:
```suggestion
                description="Check if all nova-compute units had been upgraded.",
```

cou/apps/auxiliary.py
gabrielcocenza: Actually I think it would be nicer to have a function like this on nova-compute:

```python
    def units_to_upgrade(self, target) -> list[COUUnit]:
        """Units that still need to upgrade to reach a target.

        :type target: OpenStack release.
        :return: list of units to upgrade
        :rtype: list[COUUnit]
        """
        return [
            units for units, os_version in self.os_release_units.items() if os_version < target
        ]
```

And then we could use that here like:
```python
units_not_upgraded = nova_compute.units_to_upgrade(target)
if units_not_upgraded:
    raise ApplicationError(...)
```

cou/apps/auxiliary.py
```diff
@@ -204,10 +205,55 @@ class MysqlInnodbCluster(AuxiliaryApplication):
...
+        :type target: OpenStackRelease
+        :raises ApplicationError: When any nova-compute app workload version isn't reached.
+        """
+        units_not_upgraded = []
+        apps = await self.model.get_applications()
+
+        for app in apps:
+            if app.charm != "nova-compute":
+                logger.debug("skipping application %s", app.name)
+                continue
```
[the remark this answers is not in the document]
rgildein: I'm using if and continue, to avoid having to many nested functionality. So I want to avoid having
```python
for ...
    if ...
        for ...
```
but rather
```python
for ...
    if ...
        continue

    for ...
```

cou/apps/auxiliary.py
rgildein: Base has no default value, so why adding it here?

cou/apps/auxiliary.py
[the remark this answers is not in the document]
rgildein: I use the `_verify_workload_upgrade` as inspiration for this, so should we also change that?

cou/apps/auxiliary.py
rgildein: Like this we will raise exception with not all information, if there are multiple nova-compute, we will check it only for single one.

cou/apps/auxiliary.py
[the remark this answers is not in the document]
rgildein: User fix one, then run it again a get another error and ...
I think it makes sense to collect all information and then raising it, we are doing it almost everywhere.

cou/apps/auxiliary.py
gabrielcocenza: Because ceph-osd it's not suppose to receive units, but this task should be covered on the `upgrade_plan_sanity_checks`, so I'm ok leaving this way. Solving it.

cou/apps/auxiliary.py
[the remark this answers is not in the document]
gabrielcocenza: They are kind of different things depending on the point of view .

> so should we also change that?

We could refactor a little bit, but the general idea is correct.

`_verify_workload_upgrade` is to be sure that an application got upgraded and in some apps, like rabbitmq, there are no changes on workload version for several OpenStack releases. In that case makes sense to see if the target is in a range of possible OpenStack releases. 

On nova-compute is a 1:1 map between workload_version and OpenStack release so we can safely compare with the target without the need to know what are the possible releases for a workload.

cou/apps/auxiliary.py
gabrielcocenza: It's just the general idea. You could append units on `units_not_upgraded` and then use the same way.

gabrielcocenza (commented):
> Since we add logic of units to all OpenStackApplication was introduced to all OpenStack applications, and we spoke that it's ok, I do not see why we need to raise an exception here. If you want to upgrade a single unit, it still means that all nova-computes needs to be upgraded

OpenStack Applications have actions to upgrade unit by unit. Ceph can't do that because it doesn't have the actions necessary to do it. 

Actually, I think all  `AuxiliaryApplication` can't upgrade unit by unit, so this should be target in a different PR

cou/apps/auxiliary.py
rgildein: You are forgetting, that the app is COUApplication and not NovaCompute.

gabrielcocenza: True.. I forgot it :sweat_smile: . Solving it

cou/apps/auxiliary.py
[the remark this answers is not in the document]
gabrielcocenza: I think it's easier to understand the code flow without this continue, but it's just personal taste

gabrielcocenza (commented):
I'm ok with the approach

gabrielcocenza (commented):
Just some small suggestions

tests/unit/apps/test_auxiliary.py
```diff
@@ -1020,3 +1027,127 @@ def test_mysql_innodb_cluster_upgrade(model):
...
+
+@pytest.mark.asyncio
+@patch("cou.utils.openstack.OpenStackCodenameLookup.find_compatible_versions")
+async def test_ceph_osd_verify_nova_compute_pass(mock_lookup, model):
+    """Test Ceph-osd verifying all nova computes."""
+    target = OpenStackRelease("victoria")
+    mock_lookup.return_value = [target]
+    nova_compute = MagicMock(spec_set=NovaCompute)()
+    nova_compute.charm = "nova-compute"
+    nova_compute.units = {"nova-compute/0": COUUnit("nova-compute/0", None, "21.0.0")}
```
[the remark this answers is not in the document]
rgildein: The definition of unit test is to test the function itself and not outside functions. I will add test for generating whole plan for Ceph-osd.

rgildein:
Unit tests are failing due #286.

tests/unit/apps/test_auxiliary.py
```diff
@@ -1074,3 +1134,107 @@ def test_auxiliary_upgrade_by_unit(mock_super, model):
...
+
+@pytest.mark.asyncio
+@patch("cou.utils.openstack.OpenStackCodenameLookup.find_compatible_versions")
+async def test_ceph_osd_verify_nova_compute_fail(mock_lookup, model):
+    """Test Ceph-osd verifying all nova computes."""
+    mock_lookup.return_value = [OpenStackRelease("ussuri")]
+    target = OpenStackRelease("victoria")
+    nova_compute = MagicMock(spec_set=NovaCompute)()
+    nova_compute.charm = "nova-compute"
+    nova_compute.units = {"nova-compute/0": COUUnit("nova-compute/0", None, "22.0.0")}
```
gabrielcocenza, line 1178: nit: I know that mocking `find_compatible_versions` the workload_version is irrelevant, but changing to the ussuri workload_version is preferable.
```suggestion
    nova_compute.units = {"nova-compute/0": COUUnit("nova-compute/0", None, "21.0.0")}
```

gabrielcocenza (approved):

tests/unit/apps/test_auxiliary.py
rgildein: Yes, I accidentally change it base on this [comment](https://github.com/canonical/charmed-openstack-upgrader/pull/280#discussion_r1518108139), if there will be another change-request I will change it. Thanks

PR #282, opened 2024-03-07
Auxiliary upgrade using all-in-one also when units are passed

gabrielcocenza:
- auxiliary apps cannot upgrade unit-by-unit. If units are passed, a warning message will show saying that will upgrade using all- in-one method

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -19,7 +19,7 @@ from cou.apps import LONG_IDLE_TIMEOUT
 from cou.apps.base import OpenStackApplication
 from cou.apps.factory import AppFactory
 from cou.exceptions import ApplicationError
-from cou.steps import PreUpgradeStep
+from cou.steps import ApplicationUpgradePlan, PreUpgradeStep
 from cou.utils.app_utils import set_require_osd_release_option, validate_ovn_support
 from cou.utils.juju_utils import COUUnit
 from cou.utils.openstack import (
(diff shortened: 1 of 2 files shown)


tests/unit/apps/test_auxiliary.py
```diff
@@ -1020,3 +1026,97 @@ def test_mysql_innodb_cluster_upgrade(model):
     upgrade_plan = app.generate_upgrade_plan(target, False)
 
     assert_steps(upgrade_plan, expected_plan)
+
+
+def test_auxiliary_upgrade_by_unit(model):
```
rgildein, line 1031: It would not be enough to check that `generate_upgrade_plan` was run with None, even we provided list of units?

tests/unit/apps/test_auxiliary.py
```diff
@@ -1020,3 +1026,97 @@ def test_mysql_innodb_cluster_upgrade(model):
...
+        PostUpgradeStep(
+            description=(
+                f"Check if the workload of '{app.name}' has been upgraded on units: "
+                f"{', '.join([unit for unit in app.units.keys()])}"
+            ),
+            parallel=False,
+            coro=app._verify_workload_upgrade(target, app.units.values()),
+        ),
+    ]
+    add_steps(expected_plan, upgrade_steps)
```
rgildein, line 1120: You can use `expected_plan.add_steps` instead.

rgildein (commented):
I think it's fine, but that unit test unnecessary.

tests/unit/apps/test_auxiliary.py
gabrielcocenza: Yeah, I think that is enough

tests/unit/apps/test_auxiliary.py
```diff
@@ -1020,3 +1026,51 @@ def test_mysql_innodb_cluster_upgrade(model):
...
+        can_upgrade_to="1.7/stable",
+        charm=charm,
+        channel="1.7/stable",
+        config={"source": {"value": "distro"}},
+        machines=machines,
+        model=model,
+        origin="ch",
+        series="focal",
+        subordinate_to=[],
+        units={
```
rgildein, line 1052: not-blocking: I think it save not define units or machines for this app.

cou/apps/auxiliary.py
```diff
@@ -122,6 +122,31 @@ def channel_codename(self) -> OpenStackRelease:
...
+        :param force: Whether the plan generation should be forced
+        :type force: bool
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[COUUnit]], optional
+        :return: Full upgrade plan if the Application is able to generate it.
+        :rtype: ApplicationUpgradePlan
+        """
+        if units:
+            logger.warning(
+                "%s cannot upgrade by units. The upgrade will use all-in-one method.", self.name
```
aieri, line 146:
```suggestion
                "%s cannot be upgraded using the single-unit method. The upgrade will proceed using the all-in-one method.", self.name
```

aieri (changes requested):
LGTM, just a rewording suggestion

PR #283, opened 2024-03-08
fix issue found by new version of mypy

rgildein:
Mypy found that we were using wrong type hinting.


agileshaw (approved):

aieri (approved):

aieri:
merging right away since the change doesn't affect functional tests and linting and unit tests have passed

PR #285, opened 2024-03-08
COUUnit able to evaluate its current OpenStackRelease

gabrielcocenza:
- move the responsability to evaluate OpenStackRelease from units to the COUUnit instead of the Application
- adapt unit tests

### DIFF
--- a/cou/apps/channel_based.py
+++ b/cou/apps/channel_based.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Channel based application class."""
 import logging
+from collections import defaultdict
 from typing import Iterable, Optional
 
 from cou.apps.base import OpenStackApplication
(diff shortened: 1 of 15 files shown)


gabrielcocenza:
With those changes it will be easier to not include units of `nova-compute` and colocated OpenStack apps that already got upgraded for a certain target

cou/apps/channel_based.py
```diff
@@ -28,17 +29,17 @@
 class ChannelBasedApplication(OpenStackApplication):
     """Application for charms that are channel based."""
 
-    def _get_latest_os_version(self, unit: COUUnit) -> OpenStackRelease:
-        """Get the latest compatible OpenStack release based on the channel.
+    @property
```
rgildein, line 32: Why we need this here? It's not same version as we have in base.py?

rgildein (commented):
I do not see the benefit from this approach, can you explain it more?
Like this we are mixing logic of OpenStackApplication and COUApplication, let say we have model with some apps (e.g. telegraf), which are not related with OpenStack anyhow, and we do `app.os_release`. We will get an error saying that there is no compatible version, which is confusing. I know we should never run `os_release` on not OpenStackApplication.

cou/apps/channel_based.py
gabrielcocenza: > It's not same version as we have in base.py?
No. Channel base apps doesn't have a workload version. We decide the OpenStackRelease by what the charm channel says

rgildein: I see, `os_versions[unit.os_release].append(unit.name)` and `os_versions[self.channel_codename].append(unit.name)`.

gabrielcocenza:
> Like this we are mixing logic of OpenStackApplication and COUApplication, let say we have model with some apps (e.g. telegraf), which are not related with OpenStack anyhow, and we do app.os_release. We will get an error saying that there is no compatible version, which is confusing. I know we should never run os_release on not OpenStackApplication.

Good point, I didn't think about this issue of mixing. What I'm currently missing is to easily access the `OpenStackRelease` just having the Units objects and not the Application itself. Maybe we can "overwrite"  the units in the factory once we know that the application is supported.

rgildein:
I do not think that this will be needed, so I'm closing it. Feel free to re-open it, if you come with another approach.

Issue #281, opened 2024-03-07
compare_step_coroutines are not working properly and need to be fixed


gabrielcocenza:
As it's possible to see at this [comment](https://github.com/canonical/charmed-openstack-upgrader/pull/279/files/ace4baa1014fc624acbb37883b6646f9cd69f12c#r1516816037) a typo was not identified.

PR #287, opened 2024-03-11
Fix comparing of coroutines in BaseStep

rgildein:
Fix comparing of coroutines in BaseStep, where we assumed that each coroutine has `args` or `kwargs` and completely ignored any other arguments. e.g. `model=...` was never checked. 

fixes: #281


tests/unit/steps/test_steps.py
```diff
@@ -32,21 +32,36 @@
...
-        (mock_coro(), None, False),
-        (mock_coro(), mock_coro(1, 2, 3), False),
-        (mock_coro(), mock_coro(arg1=True), False),
-        (mock_coro(), mock_coro(), True),
-        (mock_coro(1, 2, 3, kwarg1=True), mock_coro(1, 2, 3, kwarg1=True), True),
-        (mock_coro(1, 2, kwarg1=3, kwarg2=True), mock_coro(1, 2, 3, True), True),
+        (None, _coro("test/0"), False),
+        (_coro("test/0"), None, False),
+        (_coro("ceph-mon/1"), _coro("cpeh-mon/1"), False),
+        (_coro("test/0"), _coro(name="test/0"), True),
```
gabrielcocenza, line 50: This function passing got me surprised because this was the problem that I found to introduce the changes on #231

Now I discovered that the issue is comparing class methods that have async. E.g:

```python
class A:
    async def foo(self, time: int, msg:str):
        await asyncio.sleep(time)
        return msg

@pytest.mark.parametrize(
    "coro1, coro2, exp_result",
    [
        (A().foo(10, msg="hello"), A().foo(10, "hello"), True), # this unit test will fail

``` 

Not sure if we should spend time trying to fix this. I guess a unit test showing that this is not expected and a message on __eq__ method saying that it's not able to compare this situation on classes is enough to me.

gabrielcocenza (changes requested):

tests/unit/steps/test_steps.py
gabrielcocenza: Actually my example had a problem and this also works:

```
class A:
    async def foo(self, time: int, msg:str):
        await asyncio.sleep(time)
        return msg

a = A()

@pytest.mark.parametrize(
    "coro1, coro2, exp_result",
    [
        (a.foo(10, "hello"), a.foo(10, msg="hello"), True),
```

So I don't know how to reproduce the error that I was finding....
In this case I'll approve it

gabrielcocenza (approved):

tests/unit/steps/test_steps.py
gabrielcocenza: I was able to reproduce it....
If you go to `test_get_pause_unit_step` https://github.com/canonical/charmed-openstack-upgrader/blob/9e3a89a70c4f11518da459191a58488be42ebbfb/tests/unit/apps/test_base.py#L142-L157

and remove the kwargs the test will fail. Probably this is because of the model fixture that creates a `AsyncMock()` to `run_action` that might behave differently and it's not related to `compare_step_coroutines` itself.

tests/unit/steps/test_hypervisor.py
```diff
@@ -306,7 +306,7 @@ def test_hypervisor_upgrade_plan_single_machine(model):
     """
     target = OpenStackRelease("victoria")
     exp_plan = dedent_plan(
-        """
+        """\
```
agileshaw, line 309: Why is the `\` symbol needed here?

tests/unit/steps/test_steps.py
```diff
@@ -32,21 +32,36 @@
     compare_step_coroutines,
     is_unit_upgrade_step,
 )
+from cou.utils import app_utils
 
 
 async def mock_coro(*args, **kwargs): ...
 
 
+async def _coro(name: str, force: bool = False, units: list | None = None): ...
```
agileshaw, line 41: Could you add some comments here explaining why the two types of macked coros are define. It would be very confusing to look at from a future traveler's perspective.

agileshaw (changes requested):
Overall LGTM with two small comments inline.

tests/unit/steps/test_steps.py
rgildein: I will use one coro for all tests, and that will solve this issue.

gabrielcocenza (approved):

agileshaw (approved):
LGTM

PR #291, opened 2024-03-12
Preparation for OS app being frozen

rgildein:
- remove post channel settings
- drop post init charm_origin settings
- drop os_origin and origin_setting from post_init Configure these two values as properties, instead of getting them in post_init.


gabrielcocenza (commented):
LGTM. Please run all the CI checks so we can be sure that is alright

gabrielcocenza (approved):

agileshaw (approved):

PR #293, opened 2024-03-12
Add env for idle timeout

rgildein:
Add env variables for configuring the idle period of step waiting for app or model to be in idle state.

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import logging
+import os
 from collections import defaultdict
 from dataclasses import dataclass, field
 from io import StringIO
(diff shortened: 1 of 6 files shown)


agileshaw (changes requested):
Could you please also add the env variable description to README?

agileshaw (approved):
LGTM, thanks

cou/apps/base.py
```diff
@@ -46,7 +47,10 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_WAITING_TIMEOUT = 5 * 60  # 5 min
+STANDARD_IDLE_TIMEOUT: int = int(
```
gabrielcocenza, line 50: Is there a reason why those variables are not on apps/__init__.py like on dev-data/plane?

I'm afraid when we merge dev into main this will be duplicated

agileshaw: I missed this when doing review. It's not a blocker but I'm also in favor of putting them into the `__init__.py` since they are not only being used in one single file.

rgildein: We are importing from `base.py`, so "since they are not only being used in one single file." is not valid argument.
But I will move them.

rgildein: Actually, while I was doing it I realized that like this we can create circular imports. Let say we have
```python
import os
from cou.apps.base import OpenStackApplication

__all__ = ["OpenStackApplication"]
STANDARD_IDLE_TIMEOU = ...
```
and the circular import will happen.

gabrielcocenza: As we discussed on COU sync this is not a blocker. However we should pay attention if the dev/data-plane branch is merged on main. Resolving

gabrielcocenza (approved):

PR #297, opened 2024-03-12
Data-plane handle already upgraded units

gabrielcocenza:
- skip units that are already upgraded for a certain target.
- changed the name from _get_latest_os_version to get_latest_os_version
- azs start having target parameter

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -116,7 +116,7 @@ class OpenStackApplication(COUApplication):
                         "name": unit.name,
                         "machine": unit.machine.machine_id,
                         "workload_version": unit.workload_version,
-                        "os_version": str(self._get_latest_os_version(unit)),
+                        "os_version": str(self.unit_max_os_version(unit)),
                     }
                     for unit in self.units.values()
                 },
@@ -191,7 +191,7 @@ class OpenStackApplication(COUApplication):
         except ValueError:
             return self.is_from_charm_store
 
-    def _get_latest_os_version(self, unit: COUUnit) -> OpenStackRelease:
+    def unit_max_os_version(self, unit: COUUnit) -> OpenStackRelease:
         """Get the latest compatible OpenStack release based on the unit workload version.
 
         :param unit: Application Unit
@@ -251,7 +251,7 @@ class OpenStackApplication(COUApplication):
         """
         os_versions = defaultdict(list)
         for unit in self.units.values():
-            os_version = self._get_latest_os_version(unit)
+            os_version = self.unit_max_os_version(unit)
             os_versions[os_version].append(unit.name)
         return dict(os_versions)
 
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -116,7 +116,7 @@ class OpenStackApplication(COUApplication):
                         "name": unit.name,
                         "machine": unit.machine.machine_id,
                         "workload_version": unit.workload_version,
-                        "os_version": str(self.unit_max_os_version(unit)),
+                        "os_version": str(self.get_latest_os_version(unit)),
                     }
                     for unit in self.units.values()
                 },
@@ -191,7 +191,7 @@ class OpenStackApplication(COUApplication):
         except ValueError:
             return self.is_from_charm_store
 
-    def unit_max_os_version(self, unit: COUUnit) -> OpenStackRelease:
+    def get_latest_os_version(self, unit: COUUnit) -> OpenStackRelease:
         """Get the latest compatible OpenStack release based on the unit workload version.
 
         :param unit: Application Unit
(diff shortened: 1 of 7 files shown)


cou/steps/hypervisor.py
```diff
@@ -126,9 +126,8 @@ def machines(self) -> list[COUMachine]:
         """
         return self._machines
 
-    @property
-    def azs(self) -> AZs:
-        """Returns a list of AZs defined in individual applications.
+    def azs(self, target: OpenStackRelease) -> AZs:
```
rgildein, line 129: Without being property, I will suggest renaming it to `get_azs`. What do you say?

cou/apps/base.py
```diff
@@ -191,7 +191,7 @@ def is_valid_track(self, charm_channel: str) -> bool:
         except ValueError:
             return self.is_from_charm_store
 
-    def _get_latest_os_version(self, unit: COUUnit) -> OpenStackRelease:
+    def unit_max_os_version(self, unit: COUUnit) -> OpenStackRelease:
```
rgildein, line 194: I think that previous name was better, we can make it without `_`.

cou/steps/hypervisor.py
```diff
@@ -159,6 +159,13 @@ def azs(self) -> AZs:
                     logger.debug("skipping machine %s", unit.machine.machine_id)
                     continue
 
+                unit_os_release = app.unit_max_os_version(unit)
+                if unit_os_release >= target:
+                    logger.debug(
+                        "skipping unit %s is already on %s", unit.name, str(unit_os_release)
```
rgildein, line 165: Is `str(...)` really necessary?
```suggestion
                        "skipping unit %s is already on %s", unit.name, unit_os_release
```

tests/unit/steps/test_hypervisor.py
```diff
@@ -390,3 +496,119 @@ def test_hypervisor_upgrade_plan_single_machine(model):
...
 
     assert str(plan) == exp_plan
+
+
+def test_hypervisor_upgrade_plan_some_units_upgraded(model):
+    """Testing generating hypervisors upgrade plan partially upgraded."""
+    target = OpenStackRelease("victoria")
+    exp_plan = dedent_plan(
+        """
+    Upgrading all applications deployed on machines with hypervisor.
```
rgildein, line 506: Please use `\` at the beginning, so we can remove the part removing new line from `dedent_plan` helper function. More you can see [here](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

cou/steps/hypervisor.py
```diff
@@ -159,6 +159,13 @@ def azs(self) -> AZs:
                     logger.debug("skipping machine %s", unit.machine.machine_id)
                     continue
 
+                unit_os_release = app.unit_max_os_version(unit)
+                if unit_os_release >= target:
+                    logger.debug(
```
rgildein, line 164: I think we should use info here, instead of debug. It's really important information.

tests/unit/steps/test_hypervisor.py
```diff
@@ -157,22 +161,124 @@ def test_hypervisor_azs_grouping():
...
     # passing machine 1 to the HypervisorUpgradePlanner
     exp_azs_1 = AZs()
     exp_azs_1["az0"].app_units["app1"] = [units["app1/1"]]
 
     hypervisor_planner_machine_1 = HypervisorUpgradePlanner([app1, app2], [machines["1"]])
-    assert dict(hypervisor_planner_machine_1.azs) == exp_azs_1
+    assert dict(hypervisor_planner_machine_1.azs(target)) == exp_azs_1
+
+
+def test_hypervisor_azs_grouping_units_different_os_release():
```
rgildein, line 182: Is this really necessary test? It's too complex testing just simple if condition in the `azs` function.

rgildein (changes requested):

tests/unit/steps/test_hypervisor.py
gabrielcocenza: Personally I think it's good to keep this test. Unfortunately is too long to create the conditions but it shows some important behaviors like:
- apps in the same machine with different OpenStack releases (one app already upgraded), the not upgraded one will appear in the AZ
- apps already upgraded doesn't show up
- AZs already upgraded doesn't show up

rgildein: I see, but still for me, it's not good, since it's relying on multiple methods / objects, so it can fail without relevant changes in the `get_azs` function. Resolving this for now.

rgildein (approved):

agileshaw (approved):

rgildein (approved):

PR #301, opened 2024-03-13
add global and func test concurrency

rgildein:
Right now we are using global concurrency for workflow name and like this we are not running even lint and unit tests, which are run on GitHub runner instead of self-hosted one. That's why I do not see reason to block them to be run.

We also decided to use official GitHub runner for functional tests instead of self-hosted runner.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -1,10 +1,6 @@
 
 name: Check workflow running linter, unit and functional tests
 
-concurrency:
-  group: ${{ github.workflow }}
-
-
 on:
   workflow_call:
   workflow_dispatch:
@@ -14,6 +10,11 @@ on:
     paths-ignore:
       - '**.md'
       - '**.rst'
+
+concurrency:
+      # We do not want to run multiple jobs for single PR.
+      group: ${{ github.head_ref || github.ref_name }}
+      cancel-in-progress: true
 
 jobs:
   lint-unit:
@@ -61,6 +62,8 @@ jobs:
   func:
     name: Functional tests
     needs: snap-build
+    concurrency:
+      group: ci-func-tests
     runs-on: [self-hosted, large]
     timeout-minutes: 120
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -15,6 +11,11 @@ on:
       - '**.md'
       - '**.rst'
 
+concurrency:
+      # We do not want to run multiple jobs for single PR.
+      group: ${{ github.head_ref || github.ref_name }}
```
agileshaw, line 16: This would cancel other workflows, e.g. documentation check, on the same branch, which is not correct. We  can use something like [this](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) instead.

agileshaw (changes requested):

.github/workflows/check.yaml
rgildein: You can check the `head_ref` [here](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context).

rgildein: We can use `${{ github.head_ref || github.ref_name }}-lint-unit-func`, but there is no other workflow with same group, so that's why I think it's ok like this.

gabrielcocenza (approved):

.github/workflows/check.yaml
agileshaw: I understand `head_ref` is for each PR only, but within a PR we could have more workflows triggered. I still prefer using `${{ github.workflow }}-${{ github.ref }}` for each workflow to make sure the groups are tied to a single workflow definition to avoid future overhead. And I think it would be a good idea to have a team-wide template/standard for this kind of things.

agileshaw: `${{ github.workflow }}-${{ github.head_ref || github.ref }}` should be good, thanks.

agileshaw (approved):

PR #304, opened 2024-03-13
Add sanity checks to app generate upgrade plan

rgildein:
Like this we can create app without these error beeing raised and thay will be raised during generation of upgrade plan.

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -86,9 +86,6 @@ class OpenStackApplication:
     :type units: list[ApplicationUnit]
     :raises ApplicationError: When there are no compatible OpenStack release for the
         workload version.
-    :raises MismatchedOpenStackVersions: When units part of this application are running mismatched
-        OpenStack versions.
-    :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation.
     :raises CommandRunFailed: When a command fails to run.
     :raises RunUpgradeError: When an upgrade fails.
     """
@@ -226,6 +223,20 @@ class OpenStackApplication:
         """
         return self.charm_origin == "cs"
 
+    @property
+    def os_release_units(self) -> dict[OpenStackRelease, list[str]]:
+        """Get the OpenStack release versions from the units.
+
+        :return: OpenStack release versions from the units.
+        :rtype: defaultdict[OpenStackRelease, list[str]]
+        """
+        os_versions = defaultdict(list)
+        for unit in self.units:
+            os_version = self._get_latest_os_version(unit)
+            os_versions[os_version].append(unit.name)
+
+        return dict(os_versions)
+
     def is_valid_track(self, charm_channel: str) -> bool:
         """Check if the channel track is valid.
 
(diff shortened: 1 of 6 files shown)


rgildein:
I did not add the sanity check for `_check_mismatched_versions`, since that will be required too much work now without COUUnit.  I will add it after COUUnit (now only Unit) will be introduced, so here or in different PR.

gabrielcocenza (commented):
> I did not add the sanity check for _check_mismatched_versions, since that will be required too much work now without COUUnit.

It looks like that you are adding `_check_mismatched_versions` but not the unit tests. Wouldn't worth adding it later with the unit tests once the `COUUnit` is implemented on main?

rgildein:
> > I did not add the sanity check for _check_mismatched_versions, since that will be required too much work now without COUUnit.
> 
> It looks like that you are adding `_check_mismatched_versions` but not the unit tests. Wouldn't worth adding it later with the unit tests once the `COUUnit` is implemented on main?

I'll add it later with PR with `COUUnit` implementation, or by another PR, or if `COUUnit` land on main before this is merged, I will add here. But now I could not add `COUUnit`, so many PRs are blocking it. 
I thought that I was clear with my comment there, but it seems that not.

gabrielcocenza (approved):

cou/apps/base.py
```diff
@@ -678,3 +684,65 @@ def _get_wait_step(self) -> PostUpgradeStep:
...
+
+        if self.config["enable-auto-restarts"].get("value") is False:
+            raise ApplicationError(
+                "COU does not currently support upgrading applications that disable service "
+                "restarts. Please enable charm option enable-auto-restart and rerun COU to "
+                f"upgrade the {self.name} application."
+            )
+
+    def _check_application_target(self, target: OpenStackRelease) -> None:
+        """Check if application release is not lower than or equal to target.
```
aieri, line 709: non-blocking: this is accurate, but could be more clearly expressed flipping the negation
```suggestion
        """Check if the application is already upgraded.
```

cou/apps/subordinate.py
```diff
@@ -25,6 +26,23 @@
 class SubordinateBase(OpenStackApplication):
     """Subordinate base class."""
 
+    def _check_application_target(self, target: OpenStackRelease) -> None:
+        """Check if application release is not lower than or equal to target.
```
aieri, line 30: as above

cou/apps/subordinate.py
```diff
@@ -25,6 +26,23 @@
...
+
+        Subordinate application are not comparing apt_source_codename with target.
```
aieri, line 32:
```suggestion
        Subordinate applications use the apt source of the related principal and don't have an origin/openstack-origin config option.
```

aieri (approved):
lgtm, I just have rewording suggestions. Feel free to move them in a separate PR if you prefer.

cou/apps/base.py
```diff
@@ -83,14 +83,11 @@ class OpenStackApplication:
     :type units: list[ApplicationUnit]
     :raises ApplicationError: When there are no compatible OpenStack release for the
         workload version.
-    :raises MismatchedOpenStackVersions: When units part of this application are running mismatched
-        OpenStack versions.
-    :raises HaltUpgradePlanGeneration: When the class halts the upgrade plan generation.
```
agileshaw, line 88: non-blocking: why are we removing these exceptions docstrings? I'm curious about the general expectation of what and what not to include in the class docstring.

tests/unit/apps/test_base.py
```diff
@@ -41,3 +43,93 @@ def test_application_get_latest_os_version_failed(
...
+
+
+@patch("cou.apps.base.OpenStackApplication.channel", new_callable=PropertyMock)
+@patch("cou.apps.base.OpenStackApplication._verify_channel")
+@patch("cou.apps.base.OpenStackApplication.apt_source_codename", new_callable=PropertyMock)
+@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
+@patch(
+    "cou.apps.base.OpenStackApplication.can_upgrade_current_channel",
+    new_callable=PropertyMock(return_value=False),
+)
```
agileshaw, line 104: non-blocking: maybe its better to include a test case with `can_upgrade_current_channel.return_value=True` (and other conditions as `false`)

agileshaw (approved):
LGTM with two non-blocking suggestions

gabrielcocenza (approved):

cou/apps/base.py
rgildein: Anything, what __init__, __new__ or __post_init__ can raised.

tests/unit/apps/test_base.py
rgildein: You are right, but I will add it in another PR, since this one is blocking me.

PR #306, opened 2024-03-13
Generate upgrade plan for ceph-osd, principals and data-plane subordinates

gabrielcocenza:
- separate data-plane apps into hypervisor apps and non-hypervisors
- generate hypervisors plan depending on cli input
- generate ceph-osd, principals and subordinates plan depending on cli input
- raise DataPlaneCannotUpgrade if an unknown data-plane app is passed

### DIFF
--- a/cou/steps/plan.py
+++ b/cou/steps/plan.py
@@ -23,6 +23,7 @@ from typing import Callable
 from cou.apps.auxiliary import (  # noqa: F401
     AuxiliaryApplication,
     CephMon,
+    CephOsd,
     OvnPrincipal,
     RabbitMQServer,
 )
@@ -313,11 +314,22 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
     # NOTE (gabrielcocenza) upgrade group as None means that the user wants to upgrade
     #  the whole cloud.
     if args.upgrade_group in {CONTROL_PLANE, None}:
-        plan.sub_steps.extend(
+        plan.add_steps(
             _generate_control_plane_plan(target, analysis_result.apps_control_plane, args.force)
         )
+
+    hypervisor_apps, non_hypervisors_apps = _separate_hypervisors_apps(
+        analysis_result.apps_data_plane
+    )
     if args.upgrade_group in {DATA_PLANE, HYPERVISORS, None}:
-        plan.sub_steps.extend(await _generate_data_plane_plan(target, analysis_result, args))
+        plan.add_step(
+            await _generate_hypervisors_plan(args, analysis_result, target, hypervisor_apps)
+        )
+
+    if args.upgrade_group in {DATA_PLANE, None}:
+        plan.add_steps(
+            _generate_ceph_osd_and_subordinates_plan(target, non_hypervisors_apps, args.force)
+        )
 
     plan.add_steps(_get_post_upgrade_steps(analysis_result, args))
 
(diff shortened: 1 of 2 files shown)


cou/steps/plan.py
```diff
@@ -435,39 +447,84 @@ def _generate_control_plane_plan(
...
+    :rtype: tuple[list[OpenStackApplication], list[OpenStackApplication]]
+    """
+    hypervisor_apps = []
+    non_hypervisors_apps = []
+    for app in apps_data_plane:
+        if app.charm == "ceph-osd" or app.is_subordinate:
+            non_hypervisors_apps.append(app)
+            continue
+
+        hypervisor_apps.append(app)
```
rgildein, line 467: not-blocking: In this case I think it cleaner to use if else.
```suggestion
        if app.charm == "ceph-osd" or app.is_subordinate:
            non_hypervisors_apps.append(app)
        else:
            hypervisor_apps.append(app)
```

cou/steps/plan.py
```diff
@@ -435,39 +447,84 @@ def _generate_control_plane_plan(
...
+    return hypervisor_apps, non_hypervisors_apps
+
+
+async def _generate_hypervisors_plan(
     args: CLIargs,
-) -> list[UpgradePlan]:  # pragma: no cover
-    """Generate upgrade plan for data plane.
+    analysis_result: Analysis,
+    target: OpenStackRelease,
+    hypervisor_apps: list[OpenStackApplication],
```
rgildein, line 475: not-blocking: Since the name of function already contain the "hypervisors", is really important to use `hypervisor_apps` instead of `apps`? This naming seems redundant to me.

cou/steps/plan.py
```diff
@@ -435,39 +447,84 @@ def _generate_control_plane_plan(
...
+    :type target: OpenStackRelease
+    :param non_hypervisors_apps:  Non-hypervisor apps
+    :type non_hypervisors_apps: list[OpenStackApplication]
+    :param force: Whether the plan generation should be forced
+    :type force: bool
+    :return: list containing the upgrade plan for ceph-osd and subordinates
+    :rtype: list[UpgradePlan]
+    """
+    return [
+        create_upgrade_group(
```
rgildein, line 513: What is the benefit of using this helper function create_upgrade_group? Is that even right for ceph-osd?
We will have `Data Plane ceph-osd upgrade plan` step with only with sub steps `Upgrade plan for 'ceph-osd' to <target>`. 
Can we simply use
```python
for app in non_hypervisors_apps:
    app.generate_upgrade_plan
```

cou/steps/plan.py
```diff
@@ -435,39 +447,84 @@ def _generate_control_plane_plan(
...
-            # ceph-osd will not be handled by hypervisor planner
-            continue
-
-        hypervisor_apps.append(app)
 
-    hypervisor_planner = HypervisorUpgradePlanner(hypervisor_apps, hypervisors_machines)
-    hypervisor_plans = hypervisor_planner.generate_upgrade_plan(target, args.force)
-    logger.debug("Generation of the hypervisors upgrade plan complete")
+def _generate_ceph_osd_and_subordinates_plan(
+    target: OpenStackRelease, non_hypervisors_apps: list[OpenStackApplication], force: bool
```
rgildein, line 499: not-blocking: Same as for `hypervisor_apps`, can we use only `apps` here?

tests/unit/steps/test_steps_plan.py
```diff
@@ -920,5 +1025,44 @@ async def test_generate_plan_upgrade_group_data_plane(
...
+
+@patch("cou.steps.plan.create_upgrade_group")
+def test_generate_ceph_osd_and_subordinates_plan(mock_create_upgrade_group):
+    target = OpenStackRelease("victoria")
+    non_hypervisors_apps = [
+        MagicMock(spec_set=OpenStackApplication)(),
+        MagicMock(spec_set=OvnSubordinate)(),
+    ]
+    result = cou_plan._generate_ceph_osd_and_subordinates_plan(target, non_hypervisors_apps, False)
+    assert mock_create_upgrade_group.call_count == 2
```
rgildein, line 1064: We should check calls itself and not only checking that there were 2 calls. So we need to use `assert_has_calls` and define those calls.

rgildein (changes requested):

cou/steps/plan.py
gabrielcocenza: The benefit of using `create_upgrade_group` is that it handles `HaltUpgradePlanGeneration` and also some logs.

rgildein: It would not be better if the app itself managed the HaltUpgradePlanGeneration? The generate plan function cat catch that, log that and return empty plan. Like that it would be much cleaner, what do you say?

rgildein: I agree, this PR is not for it. I opened the issue #312.

tests/unit/steps/test_steps_plan.py
gabrielcocenza: I agree...
The problem was that we were passing lambda functions as arguments for `create_upgrade_group` and it's very hard to test it. If it's hard to test, it means that the code is not good, so I've changed a little bit in a way that now we can check what was called

cou/steps/plan.py
```diff
@@ -48,7 +48,7 @@
 from cou.steps.backup import backup
 from cou.steps.hypervisor import HypervisorUpgradePlanner
 from cou.utils.app_utils import set_require_osd_release_option
-from cou.utils.juju_utils import DEFAULT_TIMEOUT, COUMachine
+from cou.utils.juju_utils import DEFAULT_TIMEOUT, COUMachine, COUUnit
```
agileshaw, line 51: Not in the scope of this PR, but just want to point out `COUUnit` should be changed to `Unit`, like `Machine`, in a separate patch.

cou/steps/plan.py
```diff
@@ -416,58 +429,105 @@ def _generate_control_plane_plan(
...
+    hypervisor_apps = []
+    non_hypervisors_apps = []
+    _, nova_compute_machines = _get_nova_compute_units_and_machines(apps_data_plane)
+    for app in apps_data_plane:
+        if app.charm == "ceph-osd" or app.is_subordinate:
+            non_hypervisors_apps.append(app)
+        elif any(unit.machine in nova_compute_machines for unit in app.units.values()):
+            hypervisor_apps.append(app)
+        else:
+            raise DataPlaneCannotUpgrade(f"COU does not know how to upgrade '{app.name}'")
```
agileshaw, line 469: Is this really needed if we already have a supported list of data-plane apps defined? It's nice that we are being prudent, but I think it's also important to prune unnecessary code.

cou/steps/plan.py
```diff
@@ -313,11 +313,24 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
         )
+
+    hypervisor_apps, non_hypervisors_apps = _separate_hypervisors_apps(
+        analysis_result.apps_data_plane
+    )
     if args.upgrade_group in {DATA_PLANE, HYPERVISORS, None}:
-        plan.sub_steps.extend(await _generate_data_plane_plan(target, analysis_result, args))
+        plan.add_step(
+            await _generate_data_plane_hypervisors_plan(
+                args, analysis_result, target, hypervisor_apps
```
agileshaw, line 326: Nit: Can we make the order of arguments consistent for all the `_generate_*_plan` functions?

agileshaw (changes requested):

cou/steps/plan.py
rgildein: That will be changed on main, and then we merge main to dev/data-plane with all changes we made.

cou/steps/plan.py
```diff
@@ -313,11 +313,24 @@ async def generate_plan(analysis_result: Analysis, args: CLIargs) -> UpgradePlan
...
+            )
+        )
```
rgildein, line 328: not-blocking:
```suggestion
        step = await _generate_data_plane_hypervisors_plan(
            args, analysis_result, target, hypervisor_apps
        )
        plan.add_step(step)
```

cou/steps/plan.py
```diff
@@ -416,58 +429,105 @@ def _generate_control_plane_plan(
...
+        apps=[app for app in apps if app.is_subordinate is False],
         description="Control Plane principal(s) upgrade plan",
         target=target,
         force=force,
-        filter_function=lambda app: not isinstance(app, SubordinateBase),
     )
 
     subordinate_upgrade_plan = create_upgrade_group(
-        apps=apps,
+        apps=[app for app in apps if app.is_subordinate],
```
rgildein, line 439: It's not an real issue for us, but like this, we are iterating over apps multiple times.

cou/steps/plan.py
rgildein: I agree with removing this.

cou/steps/plan.py
rgildein: No need to change it in this PR, but we eventually should change it and remove / improve `create_upgrade_group`, which really unintuitive.

cou/steps/plan.py
```diff
@@ -505,44 +565,58 @@ async def _get_upgradable_hypervisors_machines(
...
+    )
+
+    if cli_force:
+        return nova_compute_machines
+
+    return await get_empty_hypervisors(nova_compute_units, analysis_result.model)
+
+
+def _get_nova_compute_units_and_machines(
+    apps_data_plane: list[OpenStackApplication],
```
rgildein, line 579: Why not simply just `apps`?

rgildein (changes requested):

tests/unit/steps/test_steps_plan.py
rgildein: I do not understand why it's hard to test it. Can you provide some example?

cou/steps/plan.py
gabrielcocenza: > It's not an real issue for us, but like this, we are iterating over apps multiple times.

Is it different from what `filter` in the end of the day does?

tests/unit/steps/test_steps_plan.py
rgildein: You can check the it with [`ANY`](https://docs.python.org/3/library/unittest.mock.html#any) is not perfect, but fine. I agree that the code is not the best. I really think we should avoid `create_upgrade_group`.

agileshaw (approved):
LGTM

cou/steps/plan.py
rgildein: Yes, if you pass a generator here and use that generator in function, it would be fine. But like this you are iterating over all apps and then again iterating only on filtered apps.

rgildein (approved):

PR #311, opened 2024-03-14
Add swift-proxy and swift-storage as data-plane charms

### LINKED ISSUE #310: Add swift-storage and swift-proxy as data-plane apps
Swift-storage and swift-proxy are not commonly deployed on clouds anymore, however it should be added as data-plane apps

gabrielcocenza:
Closes #310


gabrielcocenza:
After the COU sync today, it was said that those apps shouldn't upgrade using the all-in-one strategy, so we need to take more care before adding support. I'm closing this PR until we decide whether to support or not.

PR #316, opened 2024-03-18
Add logic for step to raise HaltUpgradeExecution

rgildein:
This feature was introduced in [1], however the logic was not handling in parallel execution.

After more discussion, we agreed on following execution behavior for

# Sequential:
If any step in sequence raise an exception the sequence will be stopped immediately, however if the exception is `HaltUpgradeExecution` the sequence will continue with skipping all dependent steps. 
For example, if the `Check if unit nova-compute/0 has no VMs running before upgrading.` in following example raise `HaltUpgradeExecution` all steps with prefix `├──` will be skipped and the `Enable nova-compute scheduler from unit: 'nova-compute/0'.` 
```bash
Upgrade plan for 'nova-compute' to victoria
        ...
        Upgrade plan for units: nova-compute/0, nova-compute/1, nova-compute/2
                Upgrade plan for unit: nova-compute/0
                        Disable nova-compute scheduler from unit: 'nova-compute/0'.
                        Check if unit nova-compute/0 has no VMs running before upgrading.
                        ├── Pause the unit: 'nova-compute/0'.
                        ├── Upgrade the unit: 'nova-compute/0'.
                        ├── Resume the unit: 'nova-compute/0'.
                        Enable nova-compute scheduler from unit: 'nova-compute/0'.
        ...
```
In other hand, if `├── Pause the unit: 'nova-compute/0'.` raise some exception, no further steps will be executed in `Upgrade plan for unit: nova-compute/0` and COU will exit.

# Parallel
In parallel execution, we are waiting for all steps to finished, even if any raise an exception. After all steps finished, the exception is raised with all collected exception form steps.
For example, if `Disable nova-compute scheduler from unit: 'nova-compute/0'.` raise an exception in the following example, the `Upgrade plan for unit: nova-compute/0` will be stopped, but `Upgrade plan for unit: nova-compute/1` and `Upgrade plan for unit: nova-compute/2` will continue until they finished. After each step in those the steps finished, the `RunUpgradeError` is raised and COU exit. It means that ` Wait 1800s for model test_model to reach the idle state.` will not be executed. 

```bash
Upgrade plan for 'nova-compute' to victoria
        ...
        Upgrade plan for units: nova-compute/0, nova-compute/1, nova-compute/2
                Upgrade plan for unit: nova-compute/0
                        Disable nova-compute scheduler from unit: 'nova-compute/0'.
                        ...
                Upgrade plan for unit: nova-compute/1
                        ...
                Upgrade plan for unit: nova-compute/2
                        ...
        Wait 1800s for model test_model to reach the idle state.
        Check if the workload of 'nova-compute' has been upgraded
```

---
[1]: https://github.com/canonical/charmed-openstack-upgrader/pull/239


tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+    mock_indicator.start.assert_called_once_with(upgrade_step.description)
+    upgrade_step.run.assert_awaited_once_with()
+    mock_run_sub_stpes_sequentially.assert_awaited_once_with(upgrade_step, False, True)
+    mock_indicator.succeed.assert_not_called()
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize("step", [UpgradeStep, PreUpgradeStep, PostUpgradeStep])
+@patch("cou.steps.execute.apply_step")
+@patch("cou.steps.execute._run_sub_stpes_sequentially")
```
aieri, line 169:
```suggestion
@patch("cou.steps.execute._run_sub_steps_sequentially")
```

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+    upgrade_step.run = AsyncMock()
+    upgrade_step.parallel = False
+
     await _run_step(upgrade_step, False)
-    mock_progress_indicator.start.assert_called_once()
-    mock_progress_indicator.succeed.assert_called_once()
+
+    mock_indicator.start.assert_not_called()
+    upgrade_step.run.assert_awaited_once_with()
+    mock_run_sub_stpes_sequentially.assert_awaited_once_with(upgrade_step, False, True)
```
aieri, line 184:
```suggestion
    mock_run_sub_steps_sequentially.assert_awaited_once_with(upgrade_step, False, True)
```

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+    upgrade_step.run = AsyncMock()
+    upgrade_step.parallel = True
+
     await _run_step(upgrade_step, False, True)
-    mock_progress_indicator.start.assert_called_once()
-    mock_progress_indicator.succeed.assert_not_called()
+
+    mock_indicator.start.assert_called_once_with(upgrade_step.description)
+    upgrade_step.run.assert_awaited_once_with()
+    mock_run_sub_stpes_in_parallel.assert_called_once_with(upgrade_step, False, True)
```
aieri, line 204:
```suggestion
    mock_run_sub_steps_in_parallel.assert_called_once_with(upgrade_step, False, True)
```

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+    upgrade_step = MagicMock(spec_set=UpgradeStep())
+    upgrade_step.run = AsyncMock()
+    upgrade_step.parallel = False
+
+    await _run_step(upgrade_step, False)
+
+    mock_indicator.start.assert_called_once_with(upgrade_step.description)
+    upgrade_step.run.assert_awaited_once_with()
+    mock_indicator.succeed.assert_called_once_with()
+    mock_run_sub_stpes_sequentially.assert_awaited_once_with(upgrade_step, False, False)
```
aieri, line 226:
```suggestion
    mock_run_sub_steps_sequentially.assert_awaited_once_with(upgrade_step, False, False)
```

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
-    upgrade_plan = AsyncMock(spec_set=plan("Test plan"))
-    mock_progress_indicator.spinner_id = None
-    await _run_step(upgrade_plan, False)
-    mock_progress_indicator.start.assert_not_called()
-    mock_progress_indicator.succeed.assert_not_called()
+async def test_run_step_sequentially_halt(
+    mock_indicator, mock_run_sub_stpes_sequentially, mock_apply_step
+):
+    """Test running upgrade step and all sub-steps sequentially raising HaltUpgradeExecution."""
+    mock_run_sub_stpes_sequentially.side_effect = HaltUpgradeExecution("halt")
```
aieri, line 216:
```suggestion
    mock_run_sub_steps_sequentially.side_effect = HaltUpgradeExecution("halt")
```

aieri (changes requested):
this seems generally fine (aside from the function name typo), although:
* when do we want to run substeps in parallel?
* unit tests are pretty hard to read; I tried thinking about possible parametrization or other ways to reduce cognitive complexity but couldn't come up with anything better

cou/steps/execute.py
```diff
@@ -25,6 +26,66 @@
...
+                               persist.
+    :type overwrite_progress: bool
+    :raises HaltUpgradeExecution: When any step raised it. All other steps will be completed.
+    """
+    logger.debug("running all sub-steps of %s step in parallel", step)
+    grouped_coroutines = (
+        apply_step(sub_step, prompt, overwrite_progress) for sub_step in step.sub_steps
+    )
+    try:
+        await asyncio.gather(*grouped_coroutines)
```
gabrielcocenza, line 49: nit: Is the default value, but I think this makes a little bit more explicit why the other awaitables won't be cancelled
```suggestion
        await asyncio.gather(*grouped_coroutines, return_exceptions=False)
```

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+    mock_apply_step.side_effect = [None, HaltUpgradeExecution("halt"), None]
+
+    with pytest.raises(HaltUpgradeExecution):
+        await _run_sub_stpes_sequentially(upgrade_step, False, False)
+
+    # Note(rgildein): Since the second step is raising HaltUpgradeExecution, apply_plan will be
+    #                 awaited only 2 times.
+    mock_apply_step.assert_has_awaits(
+        [call(sub_steps[0], False, False), call(sub_steps[1], False, False)]
+    )
```
gabrielcocenza, line 124: non-blocker:
I think that in that case it wold be nice to also test that the warning log is as expected

tests/unit/steps/test_steps_execute.py
```diff
@@ -16,85 +16,215 @@
...
+@patch("cou.steps.execute.apply_step")
+@patch("cou.steps.execute._run_sub_stpes_sequentially")
+@patch("cou.steps.execute.progress_indicator")
+async def test_run_step_sequentially_upgrade_step(
+    mock_indicator, mock_run_sub_stpes_sequentially, mock_apply_step
+):
+    """Test running upgrade step and all sub-steps sequentially."""
+    upgrade_step = MagicMock(spec_set=UpgradeStep())
+    upgrade_step.run = AsyncMock()
+    upgrade_step.parallel = False
```
gabrielcocenza, line 137: wouldn't make sense to add sub_steps here to match the description?

This and the next tests says about sub_steps but they don't have it.

cou/steps/execute.py
```diff
@@ -25,6 +26,66 @@
...
+            await apply_step(sub_step, prompt, overwrite_progress)
+        except HaltUpgradeExecution:
+            sub_step_index = step.sub_steps.index(sub_step)
+            logger.warning(
+                "Step: '%s' from '%s' failed to complete execution. The following steps will "
+                "be skipped:\n %s",
+                sub_step.description,
+                step.description,
+                "\n".join(
+                    sub_step.description for sub_step in step.sub_steps[sub_step_index + 1 :]
```
agileshaw, line 83: I know that step dependency is not implemented in `main` yet, but just want to put a reminder here to add that once step implementation is here ([ref](https://github.com/canonical/charmed-openstack-upgrader/pull/239/commits/611ff2838aa03a28928ab2805d3163717fbdf9a0#diff-391beb5540d6d993d384de19d144db9956db73d354d26a99ceef4a813fa42f33R70)).

gabrielcocenza (commented):
Generally LGTM. Left some non-blocker comments

cou/exceptions.py
```diff
@@ -59,6 +59,10 @@ class HaltUpgradePlanGeneration(COUException):
     """Exception to halt the application upgrade at any moment."""
```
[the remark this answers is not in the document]
rgildein: Not sure, why it was not there since I just copied and pasted it.

cou/steps/execute.py
rgildein: The dependent is already introduced in main, but here it does not make sense to use it, since the current logic is skipping any further steps and not only the dependent ones. So the logs "lied".

cou/steps/execute.py
rgildein: Even if `return_exceptions=True` others will not be canceled. So I do not see using default value here. 
If we use `return_exceptions=True` it will mean that exception will be returned and not raised, so we can collect result of any coroutine.

tests/unit/steps/test_steps_execute.py
rgildein: I do not see any benefit to checking logs, but since it's warning I see some benefit. So I will do it.

tests/unit/steps/test_steps_execute.py
rgildein: There is line testing that `mock_run_sub_steps_sequentially.assert_awaited_once_with(upgrade_step, False, False)`, and since we are not touching sub_steps in this test, it does not make sense add any.

cou/steps/execute.py
gabrielcocenza: Interesting....
I've played around to see the behavior and I think we should use `return_exceptions=True` to be sure that all tasks are completed. See the following example

```python

import asyncio

async def fn1(x):
    await asyncio.sleep(3)
    print(f"fn1 is called with x={x}")
    return f"Hello from fn1: {x}"
    
async def raise_error(y):
    await asyncio.sleep(2)  # raise error is faster than fn1
    print(f"raise_error is called with y={y}")
    raise asyncio.TimeoutError()
```

See that like this the first and third call does not print
```python
result1 = await asyncio.gather(fn1("First call"),raise_error("Second call"), fn1("Third call"), return_exceptions=False)

> raise_error is called with y=Second call
> TimeoutError                              Traceback (most recent call last)
```

If we to a try except and give more time the other tasks are completed:
```python
try:
    result1 = await asyncio.gather(fn1("First call"),raise_error("Second call"), fn1("Third call"), return_exceptions=False)

except (asyncio.TimeoutError):
    await asyncio.sleep(10)
    raise
    
> raise_error is called with y=Second call
> fn1 is called with x=First call
> fn1 is called with x=Third call
```
But adding this sleep is dirty, and when change `return_exceptions` it works as expected:
```
result1 = await asyncio.gather(fn1("First call"),raise_error("Second call"), fn1("Third call"), return_exceptions=True)

> result1
['Hello from fn1: First call',
 asyncio.exceptions.TimeoutError(),
 'Hello from fn1: Third call']
```

rgildein: You are right, I did not consider that task will be finished, but we're already handling the exception -> going to next task.
I think that at the end we will not to do `return_exceptions=True` and check if any return value is based on Exception. All value should be None only those raising exception will be some kind of exception, but like this we will not exit if another value is raised, e.g. `ValueError`. That's not perfect, either. Alternatively, we can wait for all steps to finished in the exception section and then raise the exception.
What do you think?

gabrielcocenza: > , but like this we will not exit if another value is raised, e.g. ValueError

I didn't understand this phrase. If more than one task raises an exception, I think it's a good thing to have it in the results and then compile to the user which tasks failed

> Alternatively, we can wait for all steps to finished in the exception section and then raise the exception.

What if another exception happens while waiting the others to finish it? I don't think its a good approach

cou/steps/execute.py
agileshaw: As I mentioned in the original PR, the problem with not using dependency is because including `Enable nova-compute scheduler from unit: 'nova-compute/2' step` in the skipped step gives the user a false impression that the scheduler is not re-enabled upon exiting.

This is not an ideal solution either because I think the correct way is to have `re-enabling scheduler` as a clean-up step (an actual step, not running inside check empty function) to always run. But with the log change here, at least we are not misleading the user.

agileshaw (approved):
LGTM, thanks

cou/steps/execute.py
```diff
@@ -25,6 +26,76 @@
...
+                               for all sub-steps. True to overwrite and False (the default) to
+                               persist.
+    :type overwrite_progress: bool
+    """
+    halt = False
+    logger.debug("running all sub-steps of %s step sequentially", step)
+    for sub_step in step.sub_steps:
+        if halt and sub_step.dependent:
+            logger.warning("skipping dependent step: %s", sub_step.description)
+            continue
```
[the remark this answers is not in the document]
rgildein: I agree, but I'm also not sure how we can track this.

gabrielcocenza (approved):

aieri (approved):

PR #318, opened 2024-03-19
rename COUModel to Model

rgildein:
We agreed that from all object from juju_utils we will remove prefix `COU`, so this is first part to make next PR little bit smaller.

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -5,7 +5,6 @@ Github
 frictionless
 COU
 cou
-Model
 SIGINT
 SIGINTs
 SIGTERM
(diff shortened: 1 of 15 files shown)


gabrielcocenza (approved):

docs/.custom_wordlist.txt
```diff
@@ -5,7 +5,7 @@ Github
 frictionless
 COU
 cou
-COUModel
+Model
```
agileshaw, line 8: I don't think this need to be part of the custom wordlist

agileshaw (changes requested):

docs/.custom_wordlist.txt
rgildein: Not sure, why global search did not found it. Thanks

rgildein: Ach, no I changed and it's not needed. Removing

agileshaw (approved):

gabrielcocenza (approved):

PR #324, opened 2024-03-21
Add Unit, Application and use them as base for OpenStackApplication object

rgildein:
This is implementation of multiple changes from dev/data-plane branch. Mostly from [1] and [2].

With these changes, we are using `juju_utils.Application` objects as base object for `OpenStackApplication`, and like this we moved usage of libjuju mostly to the juju_utils. The `juju_utils.Model` now include `get_applications` function, which returns all information needed for `OpenStackApplication` without returning any libjuju object.

---
[1]: https://github.com/canonical/charmed-openstack-upgrader/pull/240
[2]: https://github.com/canonical/charmed-openstack-upgrader/pull/242


gabrielcocenza (approved):

aieri (approved):
I've skimmed through the changeset but it's so huge that I cannot give a meaningful review. I also get weird deja-vu from having seen bits and pieces in other earlier PRs.
I guess that at this point... it's fine? There is no way to deal with this much cognitive complexity. Let's just not do something like this ever again...

PR #325, opened 2024-03-22
Verify Hypervisors CLI input just for hypervisors machines

gabrielcocenza:
- the validation of hypervisors machines could fail in a non-hyperconverged environment because the user could pass a machine id that had a data-plane charm. e.g: ceph-osd and not nova-compute


agileshaw (approved):
LGTM

agileshaw:
I think this should also be rebased to https://github.com/rgildein/charmed-openstack-upgrader/tree/refactor/BSENG-2203/move-testing-generate-plan and target `main` for merging

gabrielcocenza:
This PR was moved to #342 after the deletion of dev/data-plane

PR #328, opened 2024-03-25
Merging the dev/data-plane to main

rgildein:
This PR is containing all development work from dev/data-plane branch.

fixes: #315

PR #329, opened 2024-03-25
Moving upgrade plan generation to one place

rgildein:
Draft how I would like to move all steps generating the upgrade plan to single place, and we can do this for all application.
Please check only last commit, since this is based on #328.


tests/unit/apps_planning/test_auxiliary_subordinate_plans.py
```diff
@@ -0,0 +1,106 @@
...
+
+@pytest.mark.parametrize(
+    "app_name, channel, version",
+    [
+        ("hacluster", "2.4/stable", "2.4"),
+        ("mysql-router", "8.0/stable", "8.0"),
+        ("ceph-dashboard", "octopus/stable", "17.0.1"),
+    ],
+)
+def test_auxiliary_application_upgrade_plan(app_name, channel, version, model):
```
aieri, line 34: maybe `test_auxiliary_application_refresh_plan`, as this is only an in-channel refresh

Issue #330, opened 2024-03-25
The `validate_ovn_support` should be part of upgrade_plan_sanity_checks


rgildein:
Right now applications `"ovn-central", "ovn-dedicated-chassis", "ovn-chassis"` are using `validate_ovn_support` in `pre_upgrade_steps` instead of `upgrade_plan_sanity_checks`.

PR #331, opened 2024-03-25
Add `_check_ovn_support` for `ovn-chassis`, `ovn-central` and `ovn-dedicated-chassis`

rgildein:
Add `_check_ovn_support` sanity check for `ovn-chassis`, `ovn-central` and `ovn-dedicated-chassis`, instead of calling it directly in `pre_upgrade_steps`.

based on: #328 
fixes: #330


cou/apps/auxiliary.py
```diff
@@ -204,21 +204,28 @@ def _get_change_require_osd_release_step(self) -> PreUpgradeStep:
...
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
-        :param units: Units to generate upgrade plan
-        :type units: Optional[list[Unit]]
-        :return: List of pre upgrade steps.
-        :rtype: list[PreUpgradeStep]
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[Unit]], optional
+        :raises ApplicationError: When enable-auto-restarts is not enabled.
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation.
```
agileshaw, line 225: Can we also include `MismatchedOpenStackVersions`? (I'm also adding it to the [`base.py`](https://github.com/canonical/charmed-openstack-upgrader/pull/333/commits/3369569c3375fd064537872ada9b5314cb41f1b4#diff-cc23959811834c0e219a14e62eeb8f126146f7f465644751cae5ddb6f84ee318R363))

cou/apps/auxiliary_subordinate.py
```diff
@@ -43,17 +42,24 @@ def current_os_release(self) -> OpenStackRelease:
...
         :param target: OpenStack release as target to upgrade.
         :type target: OpenStackRelease
-        :param units: Units to generate upgrade plan
-        :type units: Optional[list[Unit]]
-        :return: List of pre upgrade steps.
-        :rtype: list[PreUpgradeStep]
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[Unit]], optional
+        :raises ApplicationError: When enable-auto-restarts is not enabled.
+        :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation.
```
agileshaw, line 62: same here

cou/apps/auxiliary.py
gabrielcocenza: As I said [here](https://github.com/canonical/charmed-openstack-upgrader/pull/338#discussion_r1542966999) I don't see much benefit on propagating the raises on high order functions. This docstring for example is saying about the `ApplicationError: When enable-auto-restarts is not enabled` but it's missing `When workload version is lower than 22.03.0.` It's almost impossible to keep update without a tool to help us

gabrielcocenza (approved):

cou/apps/auxiliary.py
agileshaw: This function is specifically for sanity checks, so I think we should include all possible exceptions to be thrown. But I agree it's not necessary to add for even higher level calling functions.

agileshaw (approved):
LGTM, thanks

PR #332, opened 2024-03-25
Sanity check for ovn charms to check enable-version-pinning

gabrielcocenza:
- enable-version-pinning should be set to False before upgrading [ovn charms](https://docs.openstack.org/charm-guide/latest/project/procedures/ovn-upgrade-2203.html#disable-version-pinning)

### DIFF
--- a/cou/apps/auxiliary_subordinate.py
+++ b/cou/apps/auxiliary_subordinate.py
@@ -18,7 +18,7 @@ from cou.apps.auxiliary import AuxiliaryApplication
 from cou.apps.factory import AppFactory
 from cou.apps.subordinate import SubordinateBase
 from cou.steps import PreUpgradeStep
-from cou.utils.app_utils import validate_ovn_support
+from cou.utils.app_utils import check_ovn_version_pinning, validate_ovn_support
 from cou.utils.juju_utils import COUUnit
 from cou.utils.openstack import AUXILIARY_SUBORDINATES, OpenStackRelease
 
@@ -42,6 +42,19 @@ class AuxiliarySubordinateApplication(SubordinateBase, AuxiliaryApplication):
 @AppFactory.register_application(["ovn-chassis"])
 class OvnSubordinate(AuxiliarySubordinateApplication):
     """Ovn subordinate application class."""
+
+    def upgrade_plan_sanity_checks(
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]]
+    ) -> None:
+        """Run sanity checks before generating upgrade plan.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[COUUnit]], optional
+        """
+        check_ovn_version_pinning(self.name, self.config["enable-version-pinning"].get("value"))
+        super().upgrade_plan_sanity_checks(target, units)
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[COUUnit]]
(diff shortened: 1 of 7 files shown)


cou/utils/app_utils.py
```diff
@@ -91,6 +91,22 @@ def validate_ovn_support(version: str) -> None:
         )
 
 
+def check_ovn_version_pinning(name: str, version_pinning_config: bool) -> None:
```
rgildein, line 94: not-blocking: Why not using this as part of apps itself?

```python
def _check_ovn_version_pinning(self):
    if self.config["enable-version-pinning"].get("value"):
        raise ApplicationError(f"'{self.name}' should set 'enable-version-pinning' before upgrading")
```

cou/apps/auxiliary_subordinate.py
```diff
@@ -43,6 +43,19 @@ def current_os_release(self) -> OpenStackRelease:
...
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]]
+    ) -> None:
+        """Run sanity checks before generating upgrade plan.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[COUUnit]], optional
+        """
+        check_ovn_version_pinning(self.name, self.config["enable-version-pinning"].get("value"))
```
rgildein, line 56: Should this check be after `upgrade_plan_sanity_checks`, meaning that first we run the default one?

tests/unit/apps/test_auxiliary.py
```diff
@@ -840,6 +840,37 @@ def test_ovn_workload_ver_lower_than_22_principal(model):
...
+                name=f"{charm}/0",
+                workload_version="22.03.2",
+                machine=machines["0"],
+            )
+        },
+        workload_version="22.03.2",
+    )
+
+    with pytest.raises(ApplicationError, match=exp_msg):
+        app.generate_upgrade_plan(target, False)
```
rgildein, line 871: I do not see the point of running the whole `generate_upgrade_plan` and not simply `upgrade_plan_sanity_checks`.

cou/utils/app_utils.py
gabrielcocenza: It's because we would have duplicate code from the principal ovn charm (ovn-central, ovn-dedicated-chassis) and the subordinate (ovn-chassis). That way we have a single place to share the same code

agileshaw (commented):
Overall LGTM. I will wait for a new patch addressing Robert's comment before approving

cou/utils/app_utils.py
rgildein: They are inheriting from the same place at the end, or we can define mixing.
IMO: having same 2 lines in two places would be better than having one.
Now we have:
```python
check_ovn_version_pinning(self.name, self.config["enable-version-pinning"].get("value"))
super().upgrade_plan_sanity_checks(target, units)
```
and we can have
```python
super().upgrade_plan_sanity_checks(target, units)
if self.config["enable-version-pinning"].get("value"):
    raise ApplicationError(f"'{self.name}' should set 'enable-version-pinning' before upgrading")
```

cou/apps/auxiliary_subordinate.py
rgildein: They are, but since they are raising an exception, I think it would be good to stick with the order.

rgildein:
We will not merge anything else to dev/data-plane, so this should be reopened against main after we merged #328.

gabrielcocenza:
This PR now is moved to #338 to target the main branch

PR #333, opened 2024-03-26
Bypassing COUException in application raised during plan generation sanity checks

agileshaw:
With this patch, COU generates and prints plan even if some in-scope applications failed their pre-plan sanity checks. These applications will be excluded from the plan and their warning log messages will be printed after plan generation.

Additionally, if any failure is catched this way during plan generation, COU blocks upgrades from running.

Here is an demonstration of what the output would be like if `rabbitmq-server` and `keystone` fail to pass their sanity checks when running `cou upgrade`
```
$ cou upgrade
Full execution log: '/home/ubuntu/.local/share/cou/log/cou-20231215211917.log'
Connected to 'test-model' ✔
Analyzing cloud... ✔
Generating upgrade plan... ✔
    Back up MySQL databases
    Upgrade plan for 'cinder' to 'victoria'
        Upgrade software packages of 'cinder' from the current APT repositories
            Upgrade software packages on unit cinder/0
        Refresh 'cinder' to the latest revision of 'ussuri/stable'
        Upgrade 'cinder' to the new channel: 'victoria/stable'
        Change charm config of 'cinder' 'openstack-origin' to 'cloud:focal-victoria'
        Wait for up to 300s for app 'cinder' to reach the idle state
        Verify that the workload of 'cinder' has been upgraded
    ... (other applications in the plan)
Error: Cannot generate plan for 'rabbitmq-server'.
    COU does not currently support upgrading applications that disable service restarts. Please enable charm option enable-auto-restart and rerun COU to upgrade the rabbitmq-server application.
Error: Cannot generate plan for 'keystone'. 
    Units of application keystone are running mismatched OpenStack versions: 'victoria': keystone/0, keystone/2, 'ussuri': keystone/1
Not possible to run upgrades. Please fix the errors before proceeding.
```


agileshaw:
Merged the latest `main` branch into this PR

cou/apps/base.py
```diff
@@ -360,6 +360,8 @@ def upgrade_plan_sanity_checks(
         :type units: Optional[list[Unit]], optional
         :raises ApplicationError: When enable-auto-restarts is not enabled.
         :raises HaltUpgradePlanGeneration: When the application halt the upgrade plan generation.
+        :raises MismatchedOpenStackVersions:  When the units of the app are running
+            different OpenStack versions
```
rgildein, line 364:
```suggestion
        :raises MismatchedOpenStackVersions: When the units of the app are running different
                                             OpenStack versions.
```

cou/cli.py
```diff
@@ -137,8 +138,18 @@ async def get_upgrade_plan(args: CLIargs) -> None:
...
+    upgrade_plan, warning_messages = await analyze_and_plan(args)
     print_and_debug(upgrade_plan)
+
+    if warning_messages:
+        for warning in warning_messages:
+            logger.warning(warning)
+        print(
+            "Running upgrades will not be possible until problems indicated in the warnings "
+            "are resolved."
+        )
```
rgildein, line 150: Should this be an error log message instead? Or maybe even better would be raising an exception here?

cou/cli.py
```diff
@@ -137,8 +138,18 @@ async def get_upgrade_plan(args: CLIargs) -> None:
     :param args: CLI arguments
     :type args: CLIargs
     """
-    upgrade_plan = await analyze_and_plan(args)
+    upgrade_plan, warning_messages = await analyze_and_plan(args)
     print_and_debug(upgrade_plan)
+
+    if warning_messages:
+        for warning in warning_messages:
+            logger.warning(warning)
```
rgildein, line 146:
```suggestion
        logger.warning("%s", "\n".join(warning_messages))
```

cou/cli.py
```diff
@@ -151,9 +162,18 @@ async def run_upgrade(args: CLIargs) -> None:
     :param args: CLI arguments
     :type args: CLIargs
     """
-    upgrade_plan = await analyze_and_plan(args)
+    upgrade_plan, warning_messages = await analyze_and_plan(args)
     print_and_debug(upgrade_plan)
 
+    if warning_messages:
+        for warning in warning_messages:
+            logger.warning(warning)
```
rgildein, line 170:
```suggestion
        logger.warning("%s", "\n".join(warning_messages))
```

cou/cli.py
```diff
@@ -151,9 +162,18 @@ async def run_upgrade(args: CLIargs) -> None:
...
+        print(
+            "Cannot run upgrades. Please resolve the problems indicated in the warnings "
+            "before proceeding."
+        )
```
rgildein, line 174: Same here.

tests/unit/steps/test_plan.py
```diff
@@ -295,8 +298,174 @@ async def test_generate_plan(mock_filter_hypervisors, model, cli_args):
...
 
-    upgrade_plan = await cou_plan.generate_plan(analysis_result, cli_args)
+    upgrade_plan, warning_messages = await cou_plan.generate_plan(analysis_result, cli_args)
     assert str(upgrade_plan) == exp_plan
+    assert len(warning_messages) == 0
+
+
+@pytest.mark.asyncio
+@patch("cou.steps.plan._filter_hypervisors_machines")
+async def test_generate_plan_with_warning_messages(mock_filter_hypervisors, model, cli_args):
```
rgildein, line 308: For me these test is testing so many other things, not just warning messages. I'm not sure if these is even usefull.

rgildein (changes requested):
Same small suggestions, but in general it's ok.
However, I have feelings that we should try something more sophisticate, because this seems to me too hacky, we are touching too many files and too many functions to return error messages. I have two proposal for discussion:
1) Each application can hold it's own error messages from any sanity check (it needs to become not frozen) and you can print them after planning + not run upgrade if any app `is_blocked == True`.
2) We use singleton object in planning + wrapper for any `app_upgrade_plan = app.generate_upgrade_plan(...)`. The wrapper will store warnings in singleton, and you do not need to pass them through so many functions. 

I think we already discussed the option 1, and I agreed that it may be too complicated, but second approach would be really easy and similar to what we have now, except the code will be much more readable.

cou/steps/plan.py
```diff
@@ -55,19 +56,21 @@
...
-    :return: Plan with all upgrade steps necessary based on the Analysis.
-    :rtype: UpgradePlan
+    :return: Tuple containing a plan with all upgrade steps necessary based on the Analysis,
+        and a list of recorded warning messages from plan generation.
+    :rtype: tuple[UpgradePlan, list[str]]
     """
     _pre_plan_sanity_checks(args, analysis_result)
     target = _determine_upgrade_target(analysis_result)
 
+    warning_messages = []
```
gabrielcocenza, line 73: Can we create a simple class that inherit from list just to wrap and have a more meaningful object ? E.g:

```python
class PlanWarning(list):
    ....
```
It feels more natural to me to have then a return type `tuple[UpgradePlan, PlanWarning]`

cou/cli.py
```diff
@@ -125,10 +126,10 @@ async def analyze_and_plan(args: CLIargs) -> UpgradePlan:
     progress_indicator.succeed()
 
     progress_indicator.start("Generating upgrade plan...")
-    upgrade_plan = await generate_plan(analysis_result, args)
+    upgrade_plan, warning_messages = await generate_plan(analysis_result, args)
     progress_indicator.succeed()
```
gabrielcocenza, line 130: I think we should put progress_indicator failed if there are warning messages

cou/cli.py
gabrielcocenza: I don't think we should raise an exception here, but I agree with the one in `run_upgrade` bellow

cou/steps/plan.py
```diff
@@ -653,22 +658,54 @@ def _create_upgrade_group(
...
+    :param force: Whether the plan generation should be forced
+    :type force: bool
+    :raises Exception: When cannot generate upgrade plan.
+    :return: Upgrade plan of an instance.
+    :rtype: Optional[UpgradePlan]:
+    """
+    instance_id = (
+        instance.name
+        if isinstance(instance, OpenStackApplication)
+        else "Hypervisors Groups:" + ", ".join(instance.get_azs(target).keys())
```
rgildein, line 694: Not related with this PR, but maybe we can add `name` property to the `HypervisorUpgradePlanner`. What do yo say?

cou/steps/plan.py
```diff
@@ -653,22 +658,54 @@ def _create_upgrade_group(
...
+        # for some known reason.
+        logger.debug("'%s' halted the upgrade planning generation: %s", instance_id, exc)
+    except COUException as exc:
+        logger.debug("Cannot generate plan for '%s'\n\t%s", instance_id, exc)
+        PlanWarnings().add_warning(f"Cannot generate plan for '{instance_id}'\n\t{exc}")
+    except Exception as exc:
+        logger.error("Cannot generate upgrade plan for '%s': %s", instance_id, exc)
+        raise
+
+    return None
```
rgildein, line 711: I think it would be better to return empty plan here.

cou/steps/plan.py
```diff
@@ -653,22 +658,54 @@ def _create_upgrade_group(
...
+    try:
+        instance_upgrade_plan = instance.generate_upgrade_plan(target, force)
+        return instance_upgrade_plan
+    except HaltUpgradePlanGeneration as exc:
+        # we do not care if applications halt the upgrade plan generation
+        # for some known reason.
+        logger.debug("'%s' halted the upgrade planning generation: %s", instance_id, exc)
+    except COUException as exc:
+        logger.debug("Cannot generate plan for '%s'\n\t%s", instance_id, exc)
+        PlanWarnings().add_warning(f"Cannot generate plan for '{instance_id}'\n\t{exc}")
```
rgildein, line 706: I thought we will return here the "place holder" plan. Did we decide to change it?
not-blocking: It would not be easier to define debug log message in `add_warning`?

cou/steps/plan.py
agileshaw: Yes, that'd reduce the number of conditions used.

cou/steps/plan.py
agileshaw: Actually I just read through the code for step class: if the step and none of its sub steps contain coroutine, then the step will be considered as `false` and therefore will not show up in the plan. So creating a `place-holder` plan would not work as we want. So I'm in favor of keeping it the way it is now.

cou/steps/plan.py
```diff
@@ -55,14 +58,51 @@
...
+        self._warnings.append(warning)
+
+    @property
+    def warnings(self) -> list[str]:
+        """Get all warnings stored in the class.
+
+        :return: The list of all warnings stored in the class.
+        :rtype: list[str]
+        """
+        return self._warnings
```
gabrielcocenza, line 95: I have some concerns with this class that I would to check:

I did some test and it's not behaving as I expect:

```terminal
In [117]: a = PlanWarnings()

In [118]: b = PlanWarnings()


In [119]: a is b
Out[119]: True

In [120]: a.add_warning("asdf")

In [122]: a.warnings
Out[122]: ['asdf']

In [123]: b.warnings
Out[123]: ['asdf']

In [125]: PlanWarnings().add_warning("1")

In [126]: b.warnings
Out[126]: ['1']

In [127]: PlanWarnings().add_warning("3")

In [128]: b.warnings
Out[128]: ['3']

In [129]: PlanWarnings().warnings
Out[129]: []
```
It seems that every time that PlanWarnings().warnings is called a new `_warnings` is created, so if we use this on other modules or in different executions of `_generate_instance_plan` we won't have a single warning messages

cou/cli.py
agileshaw: For planning, since generation of plan, the indicator should not be failed. For running upgrades, it is already being marked as failed.

cou/steps/plan.py
rgildein: Good catch, I was thinking about it seeing `__init__` implementation, because of `__init__` defining empty warnings, those should be defined in class instead of _instance?

agileshaw: `_instance` is also needed to to hold that single instance whenever the user tries to instantiate.

cou/steps/plan.py
```diff
@@ -55,14 +58,48 @@
...
+
+
```
rgildein, line 94:
```suggestion
class PlanWarnings:
    """Representation of a collection of warning messages from plan generation.

    This is a singleton class which holds all warning messages returned by applications
    when generating a plan.
    """

    messages: list[str] = []

    @classmethod
    def add_message(cls, message: str) -> None:
        """Add a new warning to the collection.

        :param message: A warning message to be stored.
        :type message: str
        """
        cls.messages.append(message)
```

cou/steps/plan.py
rgildein: We discuss this on sync meeting and I think it's ok. resolving. Thanks

cou/steps/plan.py
rgildein: We can remove `_instance` and basically whole `__new__`, since we are defining just list of strings, and I think we could go with classmethod so it's more easy to access those objects.

e.g.
```python
PlanWarnings.add_message("my message")
PlanWarnings.messages
# instead of
PlanWarnings().add_message("my message")
PlanWarnings().messages
```

agileshaw: Done. Thanks for the suggestions

cou/steps/plan.py
agileshaw: Resolving this thread now that we simply use class method

rgildein (approved):

gabrielcocenza (approved):
LGTM

PR #334, opened 2024-03-26
Add base structure for testing of statuses from real cloud

rgildein:
With these changes, we can simply add any result form real cloud to `tests/unit/sample_plans` directory in yaml file formatted like this:
```yaml
plan: |
  Upgrade plan defined as string

applications:
  keystone:
    can_upgrade_to: ussuri/stable
    charm: keystone
    channel: ussuri/stable
    config:
      openstack-origin:
        value: distro
      action-managed-upgrade:
        value: true
    origin: ch
    series: focal
    subordinate_to: []
    workload_version: 17.0.1
    units:
      keystone/0:
        name: keystone/0
        machine: '0'
        workload_version: 17.0.1
        os_version: ussuri
    machines:
      '0':
        id: '0'
        apps: !!python/tuple ['keystone-ldap']
        az: az-0
```

My only two concerns are:
~~- usage of `pytest-subtests`, which result could be strange if test failed, as you can see in current CI~~ After discussion I choose to got with only fixture returning dictionary with file as key, and tuple value with model and exp_plan. The model.get_applications returns applications defined in file.
~~- these test should be more part of functional tests instead of unit tests, or somehow exclude them from coverage report (e.g. 100% coverage is achieved without these tests)~~ After discussions we agreed that new tox env should be created, so I created `mocked-plans`. I also add new jobs to run this environment to our GitHub workflow.


tests/unit/utils.py
```diff
@@ -38,3 +42,74 @@ def dedent_plan(plan: str) -> str:
     result = dedent(plan)
     result = result.replace("    ", "\t")  # replace 4 spaces with tap
     return result
+
+
+def get_status():
```
agileshaw, line 47: Question: why do we also want to read json-formatted status? Is the yaml format not enough?

agileshaw (commented):
This is a partial review since I haven't finish going through all the changes yet.

gabrielcocenza (dismissed):

tests/unit/apps_planning/test_apps_plan.py
```diff
@@ -0,0 +1,34 @@
...
+
+import pytest
+
+from cou.commands import CLIargs
+from cou.steps.analyze import Analysis
+from cou.steps.plan import generate_plan
+
+
+@pytest.mark.asyncio
+@patch("cou.utils.nova_compute.get_instance_count", return_value=0)
```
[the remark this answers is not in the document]
aieri: I suspect keeping the input yamls as close to the production cloud exports would be quite adequate. Whether some hypervisors are or are not running instances is a "runtime" conditions and is not part of the model. Keeping this mocked return here allows us to more easily test what happens if none/some/all hypervisors are empty.

tests/unit/sample_plans/base.yaml
```diff
@@ -0,0 +1,155 @@
+plan: |
```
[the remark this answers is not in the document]
aieri: the way I understand this PR is:
* given a cloud, there is only one overall correct upgrade plan
* the overall plan can be segmented into sub-plans (e.g. control plane, data plane, some set of empty hypervisors)
* we should ensure that we are able to generate the correct overall plan for as many cloud models as possible
* our ability to segment the plan into sub-plans, or exclude units that cannot be upgraded for some reason, should be tested separately and independently from the input model

If the above is what we're trying to do, I think sticking to only testing the overall control plane + data plane (assuming all hypervisors are empty) upgrade plan is correct.

tests/unit/utils.py
rgildein: This is used in another tests and I simply moved it to utils from conftest, since it make much more sense.

tests/unit/apps_planning/test_apps_plan.py
```diff
@@ -0,0 +1,34 @@
...
+async def test_base_plan(_, model, sample_plans):
```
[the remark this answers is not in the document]
rgildein: Good point, I forgot to remove it.

tests/unit/apps_planning/test_apps_plan.py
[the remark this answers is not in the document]
rgildein: Currently, we are not storing this information directly with `Application` object, but it can be found in logs I think. However, I do see how it would be useful to define it in the yaml file, since it will be run multiple times (once in stage of planning and once during actual upgrade). 
It can be defined in yaml file just for planning stage, what do you say?

tests/unit/sample_plans/base.yaml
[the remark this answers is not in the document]
rgildein: In all these test I'm focused only on testing upgrade plan base on deployment and their configuration, so I did not try to do this tests for any sanity check or filtering not empty hypervisors or for testing different user inputs.

We can define more parts in the yaml, like CLIargs, number of VMs, etc. But those parts should be tested by unittests itself, since here we are focusing on real world deployment with lot of applications e.g. huge yaml file with all application definition.

tests/mocked_plans/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+from tests.mocked_plans.utils import get_sample_plan
+
+
+@pytest.fixture(scope="session")
+def sample_plans() -> dict[str, tuple[Model, str]]:
+    """Fixture that returns all sample plans in a directory.
+
+    This fixture returns a dictionary with filename as key and value as a
+    cou.utils.juju_utils.Model object whose get_applications function returns the applications
+    from the file and the expected plan.
```
agileshaw, line 29:
```suggestion
    This fixture returns a dictionary with the filename as the key and
    a tuple consisting of a cou.utils.juju_utils.Model object and the
    expected plan in string format as the value. The get_applications
    function of this Model object returns the applications read from a 
    YAML file, from which the expected plan is also parsed.
```

pyproject.toml
```diff
@@ -1,7 +1,3 @@
-# This is a template `pyproject.toml` file for snaps
-# This file is managed by bootstack-charms-spec and should not be modified
-# within individual snap repos. https://launchpad.net/bootstack-charms-spec
-
```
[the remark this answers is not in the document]
rgildein: Since this file was modified too many times, and we are not maintaining those templates anymore, I do not see reason to keep it there.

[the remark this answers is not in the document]
agileshaw: Sure, I'm Ok with that

agileshaw (approved):
LGTM, thanks

agileshaw (approved):

tests/mocked_plans/test_base_plan.py
```diff
@@ -0,0 +1,34 @@
...
+from cou.steps.analyze import Analysis
+from cou.steps.plan import generate_plan
+
+
+@pytest.mark.asyncio
+@patch("cou.utils.nova_compute.get_instance_count", return_value=0)
+async def test_base_plan(_, sample_plans):
+    """Testing the base plans."""
+    args = CLIargs("plan", auto_approve=True)
+    model, exp_plan = sample_plans["base.yaml"]
```
aieri, line 29: non-blocking: this ties a very generic test to knowledge of the individual filenames. Could we consider changing the fixture to be parametrized and return one sample plan at a time instead?

aieri: obviously this doesn't really matter right now since we only have one yaml

tests/mocked_plans/utils.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+from unittest.mock import AsyncMock, PropertyMock
+
+import yaml
+
+from cou.utils.juju_utils import Application, Machine, Model, Unit
+from tests.unit.utils import dedent_plan
+
+
+def get_sample_plan(source: Path) -> tuple[Model, str]:
```
aieri, line 24: uber-nit: if you call the yaml file a sample plan, "getting it" would only mean reading it, and this function could perhaps more accurately be renamed to `parse_sample_plan`

aieri (approved):
Thank you, I really like this work!

I have only left some non-blocking suggestions.

Additionally and as a sidenote, I don't like that we're tightly coupling what a plan _is_ to its string representation; we should be able to change how a plan is represented without having to alter the tests that verify that it is correct. But of course this cannot be solved in this PR.

tests/mocked_plans/test_base_plan.py
rgildein: I was thinking about using [Parametrizing fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#parametrizing-fixtures), however that will mean that your yaml file require working with this test configurations, e.g. CLIArgs and all hypervisors being empty. Like this, we could not create custom yaml file with let say only plan for one hypervisors (e.g. `CLIargs("hypervisors", machines=["0"], auto_approve=True)`) and on another hand like that all tests will need to work for each yaml file, since each test will run for each yaml file.

tests/mocked_plans/utils.py
rgildein: I like that, but I'll change it in another PR.

tests/mocked_plans/test_base_plan.py
rgildein: Let's discuss this further in our sync. I'm merging this PR for now, since we do not have more tests and yaml files for now, so it doesn't make much sense to implement it now.

tests/mocked_plans/utils.py
rgildein: reported #335

tests/mocked_plans/test_base_plan.py
rgildein: reported #356

PR #335, opened 2024-03-26
Fix enable and disable scheduler order on upgrade

gabrielcocenza:
- disable the nova-compute scheduler should happen before upgrading any colocated application. E.g: It's not good practice to start upgrading cinder, colocated with nova-compute,  without disabling the scheduler before
- enable the nova-compute scheduler should happen after all apps of the hypervisor are done. E.g: It's not good practice enable the nova-compute scheduler if cinder, colocated with nova-compute, didn't upgrade yet.

### DIFF
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -19,7 +19,7 @@ from typing import Iterable, Optional
 from cou.apps import LONG_IDLE_TIMEOUT
 from cou.apps.base import OpenStackApplication
 from cou.apps.factory import AppFactory
-from cou.steps import PostUpgradeStep, UnitUpgradeStep, UpgradeStep
+from cou.steps import PostUpgradeStep, PreUpgradeStep, UnitUpgradeStep, UpgradeStep
 from cou.utils.juju_utils import COUUnit
 from cou.utils.nova_compute import verify_empty_hypervisor_before_upgrade
 from cou.utils.openstack import OpenStackRelease
@@ -58,6 +58,20 @@ class NovaCompute(OpenStackApplication):
     wait_timeout = LONG_IDLE_TIMEOUT
     wait_for_model = True
     upgrade_units_running_vms = False
+
+    def pre_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[COUUnit]]
+    ) -> list[PreUpgradeStep]:
+        """Pre Upgrade steps planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[COUUnit]]
+        :return: List of pre upgrade steps.
+        :rtype: list[PreUpgradeStep]
+        """
+        return self._get_disable_scheduler_step(units) + super().pre_upgrade_steps(target, units)
 
     def upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[COUUnit]], force: bool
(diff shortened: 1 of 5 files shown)


cou/apps/core.py
```diff
@@ -117,20 +132,30 @@ def _get_empty_hypervisor_step(self, unit: COUUnit) -> UnitUpgradeStep:
...
+        for unit in units:
+            steps.append(
+                PostUpgradeStep(
+                    description=f"Enable nova-compute scheduler from unit: '{unit.name}'",
+                    coro=self.model.run_action(
+                        unit_name=unit.name, action_name="enable", raise_on_failure=True
+                    ),
+                )
+            )
+        return steps
```
rgildein, line 158: not-blocking: 
```suggestion
        if not units:
            units = list(self.units.values())

        return [
            PostUpgradeStep(
                description=f"Enable nova-compute scheduler from unit: '{unit.name}'",
                coro=self.model.run_action(
                    unit_name=unit.name, action_name="enable", raise_on_failure=True
                )
                for unit in units
            ]
```

cou/steps/hypervisor.py
```diff
@@ -228,7 +228,13 @@ def _generate_upgrade_steps(
...
             units = group.app_units[app.name]
             logger.info("generating upgrade steps for %s units of %s app", app.name, units)
-            steps.extend(app.upgrade_steps(target, units, force))
+
+            if app.charm == "nova-compute":
+                # NOTE(gabrielcocenza) it's important to disable the scheduler before upgrading
+                # any application, that is why we insert in the first position.
+                steps = app.upgrade_steps(target, units, force) + steps
+            else:
+                steps.extend(app.upgrade_steps(target, units, force))
```
rgildein, line 237: I think that this is not needed, since all apps will be in upgraded order, where nova-compute is first. If not I think we should rather make sure that the order of apps is correct even for pre- and post-upgrade steps.

cou/steps/hypervisor.py
```diff
@@ -261,7 +265,12 @@ def _generate_post_upgrade_steps(
 
             units = group.app_units[app.name]
             logger.info("generating post-upgrade steps for %s units of %s app", app.name, units)
-            steps.extend(app.post_upgrade_steps(target, units=units))
+            if app.charm == "nova-compute":
+                # NOTE(gabrielcocenza) it's important to enable the scheduler after upgrading
+                # all applications.
+                steps.extend(app.post_upgrade_steps(target, units=units))
+            else:
+                steps = app.post_upgrade_steps(target, units) + steps
```
rgildein, line 273: Same here.

rgildein (changes requested):

cou/steps/hypervisor.py
gabrielcocenza: Nova-compute might not be the first when colocated with another OpenStack service. E.g: cinder is up into the upgrade order and will be the first application to enter the data-plane group

cou/steps/hypervisor.py
gabrielcocenza: As said above, the app sequence might start with OpenStack apps if colocatd. This probably it's not necessary, but it makes it explicit.

cou/steps/hypervisor.py
rgildein: Ok, in tht case could we re-order all apps and then pass them to the HypervosorUpgradePlaner?

It will be much cleaner then these if statments. Would that be ok?

Now I realized, if cinder should be upgraded first, it's ok that we change that here for nova-compute to be first? Maybe we can simply move disable to pre-upgrade and enable to post-upgrade (we already did that).

cou/steps/hypervisor.py
rgildein: Since post-upgrade are only there to chceck and not upgrade, is the order when scheduler is enabled important?

Issue #299, opened 2024-03-12
Hypervisors should disable scheduler before starting upgrading


gabrielcocenza:
Right now COU are not necessarily disabling the nova-compute scheduler before upgrading colocated apps like cinder. That means that COU can start upgrading a hypervisor machine starting with cinder without disabling it. 
The sequence should be:

1. Application level steps for nova-compute and colocated apps
2. disable nova-compute scheduler
3. unit upgrade steps for nova-compute and colocated apps
4. enable nova-compute scheduler
5. post-upgrade steps for the apps

PR #337, opened 2024-03-27
Fix enable and disable scheduler order on upgrade

gabrielcocenza:
- disable the nova-compute scheduler should happen before upgrading any colocated application. E.g: It's not good practice to start upgrading cinder, colocated with nova-compute,  without disabling the scheduler before
- enable the nova-compute scheduler should happen after all apps of the hypervisor are done. E.g: It's not good practice enable the nova-compute scheduler if cinder, colocated with nova-compute, didn't upgrade yet.

Fix: #299


cou/apps/core.py
```diff
@@ -87,8 +117,7 @@ def _get_unit_upgrade_steps(self, unit: Unit, force: bool) -> UnitUpgradeStep:
         :return: Unit upgrade step
         :rtype: UnitUpgradeStep
         """
-        unit_plan = UnitUpgradeStep(f"Upgrade plan for unit '{unit.name}'")
-        unit_plan.add_step(self._get_disable_scheduler_step(unit))
+        unit_plan = UnitUpgradeStep(description=f"Upgrade plan for unit '{unit.name}'")
```
agileshaw, line 120: non-blocking: I think it would be better to add a new class `UnitUpgradePlan` to be used here.

cou/apps/core.py
```diff
@@ -116,32 +144,42 @@ def _get_empty_hypervisor_step(self, unit: Unit) -> UnitUpgradeStep:
...
+                description=f"Enable nova-compute scheduler from unit: '{unit.name}'",
+                coro=self.model.run_action(
+                    unit_name=unit.name, action_name="enable", raise_on_failure=True
+                ),
+            )
+            for unit in units
+        ]
+
+    def _get_disable_scheduler_step(self, units: Optional[list[Unit]]) -> list[PreUpgradeStep]:
+        """Get the step to disable the scheduler,  so the unit cannot create new VMs.
```
agileshaw, line 168:
```suggestion
        """Get the step to disable the scheduler, so the unit cannot create new VMs.
```

agileshaw (commented):
The changes LGTM. But I think we should wait until #339 is merged to `main` (or rebase on it) so that we can update the documentation also in this PR. Otherwise it's easy to forget.

cou/apps/core.py
rgildein: I agree that that would be nice, it's not related with this PR, so I will suggest to no to it here. We can report an issue about this.

not-blocking: Why you are adding `description=` back? I do not see any benefit of it and in a lot of places I tried to remove it, so make lines shorter.

cou/apps/core.py
```diff
@@ -116,32 +144,42 @@ def _get_empty_hypervisor_step(self, unit: Unit) -> UnitUpgradeStep:
...
-
-    def _get_disable_scheduler_step(self, unit: Unit) -> UnitUpgradeStep:
-        """Get the step to disable the scheduler, so the unit cannot create new VMs.
-
-        :param unit: Unit to be disabled.
-        :type unit: Unit
-        :return: Step to disable the scheduler.
-        :rtype: UnitUpgradeStep
+        if not units:
+            units = list(self.units.values())
```
rgildein, line 156: not-blocking: 
```suggestion
            units = self.units.values()
```

cou/apps/core.py
```diff
@@ -116,32 +144,42 @@ def _get_empty_hypervisor_step(self, unit: Unit) -> UnitUpgradeStep:
...
-        )
-
-    def _get_disable_scheduler_step(self, unit: Unit) -> UnitUpgradeStep:
-        """Get the step to disable the scheduler, so the unit cannot create new VMs.
-
-        :param unit: Unit to be disabled.
-        :type unit: Unit
-        :return: Step to disable the scheduler.
-        :rtype: UnitUpgradeStep
+        if not units:
```
rgildein, line 155: We are using same if condition in multiple places, but now I realized that it maybe an issue if I want to generate upgrade plan for app but not touching any units, so I passed `units == []`. Should we rather use `if units is None`? What do you think?

cou/steps/hypervisor.py
```diff
@@ -236,9 +237,7 @@ def _generate_post_upgrade_steps(
         :rtype: list[PostUpgradeStep]
         """
         steps = []
-        # NOTE(rgildein): Using the reverse order of post-upgrade steps, so these steps starts from
-        #                 subordinate or colocated apps and not nova-compute.
-        for app in self.apps[::-1]:
+        for app in self.apps:
```
rgildein, line 240: I still think that running in reverse order make more sense, but if you think that it can be harmful, I'm ok to keep this change.

tests/unit/apps/test_core.py
```diff
@@ -769,28 +793,73 @@ def test_nova_compute_get_empty_hypervisor_step(model):
...
 
 
-def test_nova_compute_get_enable_scheduler_step(model):
+@pytest.mark.parametrize("units_slice", [1, 2, 3])
+def test_nova_compute_get_enable_scheduler_step(model, units_slice):
+    """Enable the scheduler on selected units."""
     app = _generate_nova_compute_app(model)
-    units = list(app.units.values())
-    unit = units[0]
+    units_selected = list(app.units.values())[0:units_slice]
```
rgildein, line 800:
```suggestion
    units_selected = [app.units[unit] for unit in units]  # units instead of units_slice, where units is list of names
```

cou/apps/core.py
```diff
@@ -116,32 +144,42 @@ def _get_empty_hypervisor_step(self, unit: Unit) -> UnitUpgradeStep:
...
-                unit_name=unit.name, action_name="disable", raise_on_failure=True
-            ),
-        )
+        if not units:
+            units = list(self.units.values())
+        return [
+            PreUpgradeStep(
+                description=f"Disable nova-compute scheduler from unit: '{unit.name}'",
+                coro=self.model.run_action(
+                    unit_name=unit.name, action_name="disable", raise_on_failure=True
```
rgildein, line 181: not-blocking: What if we use one functions with `enable=True` for enabling (run enable action) and `enable=False` for disabling (run disable action).

rgildein (changes requested):

cou/apps/core.py
rgildein: Good point, resolving. :+1:

cou/apps/core.py
gabrielcocenza: Interesting...
I think passing `units = []` will only happen on all units of the hypervisor is upgraded and then `_check_application_target` will halt the upgrade plan generation. Howevwer, we can have a single place to check if units is None and not checking every time.

I will add this is this patch. It's a little bit out of the scope of this PR, but I don't think it's a big deal

rgildein: What do you mean by single place to check it?

gabrielcocenza: Digging a little bit it looks like that the changes required are more complex than I expected so I'll make it simple in this PR as you suggested. I've opened a issue #344 so we can better think about this

cou/steps/hypervisor.py
gabrielcocenza: To say the truth I don't understand why we would change the order.
1. We are not using the hypervisor to upgrade subordinates, so the description is outdated
2. Because of the upgrade order, control-plane colocated apps starts to upgrade first and then nova-compute

I think we should keep the way it is and I don't see harm

cou/apps/core.py
agileshaw: Issue for adding `UnitUpgradePlan` created: https://github.com/canonical/charmed-openstack-upgrader/issues/346

cou/apps/core.py
gabrielcocenza: > What do you mean by single place to check it?

As you said: `We are using same if condition in multiple places` That is what I meant by single place to check that condition.
This condition is happening because we need to check in multiple functions which upgrade strategy is used. We should think a better and more elegant solution for this later.

tests/unit/apps/test_core.py
```diff
@@ -825,14 +831,21 @@ def test_nova_compute_get_enable_scheduler_step_no_units(model, units_selected):
...
 
 
-@pytest.mark.parametrize("units_slice", [1, 2, 3])
-def test_nova_compute_get_disable_scheduler_step(model, units_slice):
+@pytest.mark.parametrize(
+    "units",
+    [
+        [f"nova-compute/{unit}" for unit in range(1)],
+        [f"nova-compute/{unit}" for unit in range(2)],
+        [f"nova-compute/{unit}" for unit in range(3)],
```
rgildein, line 842: not-blocking: I think it would be ok to define without range, but definitely no blocker.

rgildein (approved):

rgildein (approved):
lgtm

agileshaw (approved):

PR #338, opened 2024-03-27
Sanity check for ovn charms to check enable-version-pinning

gabrielcocenza:
- enable-version-pinning should be set to False before upgrading ovn charms. More details can be found in the [docs](https://docs.openstack.org/charm-guide/latest/project/procedures/ovn-upgrade-2203.html#disable-version-pinning).


cou/apps/auxiliary.py
```diff
@@ -204,6 +204,23 @@ def _get_change_require_osd_release_step(self) -> PreUpgradeStep:
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[Unit]], optional
+        :raises ApplicationError: When enable-version-pinning is True
+        """
+        super().upgrade_plan_sanity_checks(target, units)
+        if self.config["enable-version-pinning"].get("value"):
+            raise ApplicationError(
+                f"'{self.name}' should set 'enable-version-pinning' before upgrading"
```
agileshaw, line 221: Didn't mention what value to be set for `enable-version-pinning`.
```suggestion
                f"Cannot upgrade '{self.name}'. 'enable-version-pinning' must be set to `false`."
```

agileshaw (changes requested):

agileshaw (dismissed):
LGTM, thanks

cou/apps/auxiliary.py
```diff
@@ -204,6 +204,23 @@ def _get_change_require_osd_release_step(self) -> PreUpgradeStep:
...
+    def upgrade_plan_sanity_checks(
+        self, target: OpenStackRelease, units: Optional[list[Unit]]
+    ) -> None:
+        """Run sanity checks before generating upgrade plan.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[Unit]], optional
+        :raises ApplicationError: When enable-version-pinning is True
```
rgildein, line 216: not-blocking: This function can raise more than ApplicationError, I know it's coming from inherited class, but still I think it should be mention here. I'm doing the same thing [here](https://github.com/canonical/charmed-openstack-upgrader/pull/331/files#diff-a29275aa9b0cacf5c69098e1afcce54ff8c5c0c6cd0480d12a5b483939047b34R217).

rgildein (approved):
Small not-blocking comment.

cou/apps/auxiliary.py
gabrielcocenza: I don't see much benefit on doing this because of the maintainability. If we had a tool that help us track this kind of thing I would be favorable of doing it, but right now we make a change it's very easy to have docstrings not updated.

cou/apps/auxiliary.py
```diff
@@ -223,13 +251,27 @@ def upgrade_plan_sanity_checks(
         :type target: OpenStackRelease
         :param units: Units to generate upgrade plan, defaults to None
         :type units: Optional[list[Unit]], optional
-        :raises ApplicationError: When enable-auto-restarts is not enabled.
+        :raises ApplicationError: When enable-auto-restarts is not enabled or .
```
rgildein, line 254: or ???

rgildein (changes requested):

cou/apps/auxiliary.py
```diff
@@ -202,17 +205,42 @@ def _get_change_require_osd_release_step(self) -> PreUpgradeStep:
         )
 
 
-@AppFactory.register_application(["ovn-central", "ovn-dedicated-chassis"])
-class OvnPrincipal(AuxiliaryApplication):
-    """Ovn principal application class."""
+class Ovn(AuxiliaryApplication):
```
agileshaw, line 208: non-blocking: I personally don't like having a separate class here and do cross inheritance when it is not absolutely necessary. `app_utils` already holds helper functions for different apps, so adding another `check_ovn_version_pinning` wouldn't be so bad. And if we really want to separate the helper functions for OVN, maybe we can do what we did for `nova compute`' and create a separate utils file?

rgildein: Gabriel was doing that, but was looking strange. You can see it [here](https://github.com/canonical/charmed-openstack-upgrader/pull/332#discussion_r1538893162).

agileshaw: I saw that and TBH I didn't find it weird. But if you think this is too OVN-specific, we could have it put in a separate `ovn_utils.py`.

But then again, this is a non-blocking suggestion.

gabrielcocenza: I though it was a little bit weird too because to pass the argument to the function it's verbose like `self.config["enable-version-pinning"].get("value")`. It feels more natural to me handle in the class itself, but no strong opinion.

I'm also not a big fan of `cross inheritance`, but `ovn` apps has a lot in common and the code started duplicating on two modules. Actually lint started complaining about that.

I think it's good to have one class that have share the common things of ovn. In the future if more business logic is necessary, like new sanity check or extra upgrade step changing in one place for both classes seems nice to me

agileshaw (approved):

rgildein (approved):

Issue #336, opened 2024-03-27
Use `warning` log level as the default verbosity for STDOUT


agileshaw:
COU is currently only printing `error` level logging messages to stdout with the default verbosity. As warning messages are also providing important notes and information to the user, we should also include them in stdout without user specifically increasing verbosity levels.

The expected behavior:
- Default: both errors and warnings will be logged.
- `-v`: Errors, warnings, and general information will be logged.
- `-vv`: Detailed debugging information will be logged alongside above-mentioned messages.
- `-vvv` and up: Maximum verbosity where everything will be logged.

We would also need to change the related documentation

PR #339, opened 2024-03-27
Add data plane documentations

agileshaw:
Updated the documentation to include instructions and references for the data-plane upgrade, as well as explanations for different upgrade strategies

cherry-pick: #317 #322 #323

PR #340, opened 2024-03-27
Include WARNING level logs to STDOUT in default verbosity

agileshaw:
Printing only ERROR messages in default verbosity is not enough because WARNING messages, like some application fails to pass their sanity checks, are also important to show the user in STDOUT, even if they didn't specially turn on verbose mode.

The change is based on #339 , so please only review the last commit. And this PR should not be merged before #339 is.

fixes: #336


gabrielcocenza (approved):

rgildein (approved):

docs/reference/commands.rst
```diff
@@ -73,7 +73,11 @@ upgrade phase. Refer to the output below for a description of all available opti
                             Default to enabling database backup.
       --force               Force the plan/upgrade of non-empty hypervisors.
       --verbose, -v         Increase logging verbosity in STDOUT.
-                            Multiple 'v's yield progressively more detail (up to 3).
+                            Multiple 'v's yield progressively more detail (
+                            
+                            
+                            
+                            3).
```
rgildein, line 80: Why this in on 5 lines?

agileshaw: Seems to be a merging hiccup. Fixed, thanks

rgildein (approved):

gabrielcocenza (approved):

PR #341, opened 2024-03-27
Fix apps split between control-plane and data-plane

gabrielcocenza:
- subordinates charms don't have units, so accessing machines from units is not possible and this makes all subordinate charms being considered as control-plane
- acessing machines directly from apps solves the issue.

### DIFF
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -84,7 +84,7 @@ class Analysis:
         for app in apps:
             if is_data_plane(app):
                 data_plane.append(app)
-            elif any(unit.machine in data_plane_machines for unit in app.units.values()):
+            elif any(machine in data_plane_machines for machine in app.machines.values()):
                 data_plane.append(app)
             else:
                 control_plane.append(app)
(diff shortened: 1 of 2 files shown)


agileshaw (approved):
LGTM. Nice catch :+1:

cou/steps/analyze.py
```diff
@@ -84,7 +84,7 @@ def is_data_plane(app: OpenStackApplication) -> bool:
         for app in apps:
             if is_data_plane(app):
                 data_plane.append(app)
-            elif any(unit.machine in data_plane_machines for unit in app.units.values()):
+            elif any(machine in data_plane_machines for machine in app.machines.values()):
```
rgildein, line 87: What is the difference here?

rgildein: Och, I see, subordinate does not have units. Nice catch.

rgildein (approved):
Thanks :+1:

cou/steps/analyze.py
rgildein: I realized that this is not a proper fix, since we are getting machines from `status.units` [here](https://github.com/canonical/charmed-openstack-upgrader/blob/e00c5824bbe8f8e3ed7e263a16b42855fedced67/cou/utils/juju_utils.py#L358) and that will be empty for even subordinate.
So we need to change getting those machines for app or do something else.

PR #342, opened 2024-03-27
Verify Hypervisors CLI input just for hypervisors machines

gabrielcocenza:
- the validation of hypervisors machines could fail in a non-hyperconverged environment because the user could pass a machine id that had a data-plane charm. e.g: ceph-osd and not nova-compute


rgildein (approved):
Thanks

agileshaw (approved):

PR #345, opened 2024-03-28
Add Swift application class

agileshaw:
Add a new Swift application class for swift-proxy and swift-storage. This class will also raise ApplicationNotSupported exception when generating a plan.

Also make sure both swift-proxy and swift-storage are considered as data-plane applications.

fixes: #310

### DIFF
--- a/tests/unit/apps/test_core.py
+++ b/tests/unit/apps/test_core.py
@@ -17,8 +17,12 @@ import pytest
 from juju.client._definitions import ApplicationStatus, UnitStatus
 
 from cou.apps.base import OpenStackApplication
-from cou.apps.core import Keystone, NovaCompute
-from cou.exceptions import ApplicationError, HaltUpgradePlanGeneration
+from cou.apps.core import Keystone, NovaCompute, Swift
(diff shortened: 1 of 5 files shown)


rgildein (approved):
LGTM

rgildein (approved):

tests/unit/apps/test_core.py
```diff
@@ -1032,3 +1036,37 @@ def test_cinder_upgrade_plan_single_unit(model):
     plan = cinder.generate_upgrade_plan(target, False, [units["cinder/0"]])
 
     assert str(plan) == exp_plan
+
+
```
gabrielcocenza, line 1040: non-blocker: Maybe adding a pytest parametrize to also use `swift-proxy`

gabrielcocenza (approved):

tests/unit/apps/test_core.py
agileshaw: I thought about that but I don't think there's much added value because only names of the charm/unit is changing, and no additional functionality would be tested.

PR #347, opened 2024-04-01
Add support for latest/stable

gabrielcocenza:
- currently if the applications uses `latest/stable` channel it's not considered as a valid channel to upgrade.
- the approach is to move to an OpenStack release channel when this happens, based on the current workload version

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -54,30 +54,21 @@ class AuxiliaryApplication(OpenStackApplication):
             return True
 
         track = self._get_track_from_channel(charm_channel)
-        return (self.charm, self.series, track) in TRACK_TO_OPENSTACK_MAPPING
-
-    @property
-    def possible_current_channels(self) -> list[str]:
-        """Return the possible current channels based on the series and current OpenStack release.
-
-        :return: The possible current channels for the application.
-        :rtype: list[str]
-        :raises ApplicationError: When cannot find tracks.
-        """
         tracks = OPENSTACK_TO_TRACK_MAPPING.get(
             (self.charm, self.series, self.current_os_release.codename)
         )
-        if tracks:
-            return [f"{track}/stable" for track in tracks]
+        return (
+            self.charm,
+            self.series,
+            track,
+        ) in TRACK_TO_OPENSTACK_MAPPING and bool(tracks)
 
-        raise ApplicationError(
-            (
-                f"Cannot find a suitable '{self.charm}' charm channel for "
-                f"{self.current_os_release.codename} on series '{self.series}'. "
-                "Please take a look at the documentation: "
-                "https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html"
-            )
+    @property
+    def possible_current_channel(self) -> str:
+        *_, track = OPENSTACK_TO_TRACK_MAPPING.get(
+            (self.charm, self.series, self.current_os_release.codename), []
         )
+        return f"{track}/stable"
 
     def target_channel(self, target: OpenStackRelease) -> str:
         """Return the appropriate channel for the passed OpenStack target.
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -64,17 +64,19 @@ class AuxiliaryApplication(OpenStackApplication):
         ) in TRACK_TO_OPENSTACK_MAPPING and len(possible_tracks) > 0
 
     @property
-    def possible_current_channel(self) -> str:
-        """Return the possible current channel.
+    def expected_current_channel(self) -> str:
+        """Return the expected current channel.
 
-        Possible current channel is based in the series and in the current OpenStack release of
-        the application.
-        :return: The possible current channel of the application. E.g: "3.9/stable"
+        Expected current channel is the channel that the application is suppose to be using based
+        on the current series, workload version and, by consequence, the OpenStack release
+        identified.
+        :return: The expected current channel of the application. E.g: "3.9/stable"
         :rtype: str
         """
-        *_, track = OPENSTACK_TO_TRACK_MAPPING.get(
-            (self.charm, self.series, self.current_os_release.codename), []
-        )
+        *_, track = OPENSTACK_TO_TRACK_MAPPING[
+            (self.charm, self.series, self.current_os_release.codename)
+        ]
+
         return f"{track}/stable"
 
     def target_channel(self, target: OpenStackRelease) -> str:
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -53,18 +53,25 @@ class AuxiliaryApplication(OpenStackApplication):
             logger.debug("'%s' has been installed from the charm store", self.name)
             return True
 
-        track = self._get_track_from_channel(charm_channel)
-        tracks = OPENSTACK_TO_TRACK_MAPPING.get(
-            (self.charm, self.series, self.current_os_release.codename)
+        current_track = self._get_track_from_channel(charm_channel)
+        possible_tracks = OPENSTACK_TO_TRACK_MAPPING.get(
+            (self.charm, self.series, self.current_os_release.codename), []
         )
         return (
             self.charm,
             self.series,
-            track,
-        ) in TRACK_TO_OPENSTACK_MAPPING and bool(tracks)
+            current_track,
+        ) in TRACK_TO_OPENSTACK_MAPPING and len(possible_tracks) > 0
 
     @property
     def possible_current_channel(self) -> str:
+        """Return the possible current channel.
+
+        Possible current channel is based in the series and in the current OpenStack release of
+        the application.
+        :return: The possible current channel of the application. E.g: "3.9/stable"
+        :rtype: str
+        """
         *_, track = OPENSTACK_TO_TRACK_MAPPING.get(
             (self.charm, self.series, self.current_os_release.codename), []
         )
@@ -115,7 +122,6 @@ class AuxiliaryApplication(OpenStackApplication):
 
         track: str = self._get_track_from_channel(self.channel)
         compatible_os_releases = TRACK_TO_OPENSTACK_MAPPING[(self.charm, self.series, track)]
-        # channel setter already validate if it is a valid channel.
         return max(compatible_os_releases)
 
     def generate_upgrade_plan(
@@ -144,6 +150,20 @@ class AuxiliaryApplication(OpenStackApplication):
                 self.name,
             )
         return super().generate_upgrade_plan(target, force, None)
+
+    def _need_to_refresh(self, target: OpenStackRelease) -> bool:
+        """Check if the application needs to refresh the current channel.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: True if needs to refresh, False otherwise
+        :rtype: bool
+        """
+        track: str = self._get_track_from_channel(self.channel)
+        compatible_os_releases = TRACK_TO_OPENSTACK_MAPPING[(self.charm, self.series, track)]
+        return bool(self.can_upgrade_to) and any(
+            os_release <= target for os_release in compatible_os_releases
+        )
 
 
 @AppFactory.register_application(["rabbitmq-server"])
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -153,7 +153,7 @@ class AuxiliaryApplication(OpenStackApplication):
             )
         return super().generate_upgrade_plan(target, force, None)
 
-    def _need_to_refresh(self, target: OpenStackRelease) -> bool:
+    def _need_current_channel_refresh(self, target: OpenStackRelease) -> bool:
         """Check if the application needs to refresh the current channel.
 
         :param target: OpenStack release as target to upgrade.
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -241,12 +241,14 @@ class OpenStackApplication(Application):
         return None
 
     @property
-    def possible_current_channel(self) -> str:
-        """Return the possible current channel.
+    def expected_current_channel(self) -> str:
+        """Return the expected current channel.
 
-        Possible current channel is based in the current OpenStack release of the application.
+        Expected current channel is the channel that the application is suppose to be using based
+        on the current series, workload version and, by consequence, the OpenStack release
+        identified.
 
-        :return: The possible current channel of the application. E.g: "ussuri/stable"
+        :return: The expected current channel of the application. E.g: "ussuri/stable"
         :rtype: str
         """
         return f"{self.current_os_release.codename}/stable"
@@ -265,23 +267,12 @@ class OpenStackApplication(Application):
 
         return dict(os_versions)
 
-    @property
-    def can_refresh(self) -> bool:
-        """Check if the application can refresh.
-
-        Application can refresh if are not from charm store and it's not using
-        latest/stable channel.
-        :return: True if can refresh, false otherwise
-        :rtype: bool
-        """
-        return not self.is_from_charm_store and self.channel != LATEST_STABLE
-
     def is_valid_track(self, charm_channel: str) -> bool:
         """Check if the channel track is valid.
 
         :param charm_channel: Charm channel. E.g: ussuri/stable
         :type charm_channel: str
-        :return: True if valid, False otherwise.
+        :return: True if valid, False otherwise
         :rtype: bool
         """
         try:
@@ -401,8 +392,6 @@ class OpenStackApplication(Application):
         """
         return [
             self._get_upgrade_current_release_packages_step(units),
-            self._get_charm_hub_migration(),
-            self._get_change_to_openstack_channels(),
             self._get_refresh_charm_step(target),
         ]
 
@@ -536,7 +525,22 @@ class OpenStackApplication(Application):
 
         return step
 
-    def _get_charm_hub_migration(self) -> PreUpgradeStep:
+    def _get_refresh_charm_step(self, target: OpenStackRelease) -> PreUpgradeStep:
+        """Get step for refreshing the charm.
+
+        :param target: OpenStack release as target to upgrade
+        :type target: OpenStackRelease
+        :raises ApplicationError: When application has unexpected channel.
+        :return: Step for refreshing the charm
+        :rtype: PreUpgradeStep
+        """
+        if refresh_step := self._get_charmhub_migration_step():
+            return refresh_step
+        if refresh_step := self._get_change_to_openstack_channels_step():
+            return refresh_step
+        return self._get_refresh_current_channel_step(target)
+
+    def _get_charmhub_migration_step(self) -> PreUpgradeStep:
         """Get the step for charm hub migration from charm store if necessary.
 
         :return: Step for charmhub migration
@@ -548,11 +552,11 @@ class OpenStackApplication(Application):
         return PreUpgradeStep(
             f"Migrate '{self.name}' from charmstore to charmhub",
             coro=self.model.upgrade_charm(
-                self.name, self.possible_current_channel, switch=f"ch:{self.charm}"
+                self.name, self.expected_current_channel, switch=f"ch:{self.charm}"
             ),
         )
 
-    def _get_change_to_openstack_channels(self) -> PreUpgradeStep:
+    def _get_change_to_openstack_channels_step(self) -> PreUpgradeStep:
         """Get the step for changing to OpenStack channels if necessary.
 
         :return: Step for changing to OpenStack channels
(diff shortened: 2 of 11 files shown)


cou/apps/auxiliary.py
```diff
@@ -54,30 +54,21 @@ def is_valid_track(self, charm_channel: str) -> bool:
...
         tracks = OPENSTACK_TO_TRACK_MAPPING.get(
             (self.charm, self.series, self.current_os_release.codename)
         )
-        if tracks:
-            return [f"{track}/stable" for track in tracks]
+        return (
+            self.charm,
+            self.series,
+            track,
+        ) in TRACK_TO_OPENSTACK_MAPPING and bool(tracks)
```
rgildein, line 64: what is `bool(tacks)` for?

cou/apps/base.py
```diff
@@ -516,6 +523,33 @@ def _get_upgrade_current_release_packages_step(
...
+            f"Change '{self.name}' channel from {self.channel} to {self.possible_current_channel}. "
+            "COU cannot guarantee that this is not a downgrade",
+            coro=self.model.upgrade_charm(self.name, self.possible_current_channel),
+        )
+
+    def _need_to_refresh(self, target) -> bool:
+        return self.can_upgrade_to or self.channel_codename < target
+
+    @property
+    def _can_refresh(self) -> bool:
```
rgildein, line 550: Why doing "private" property here? And please try to keep all properties in one place.

cou/apps/base.py
```diff
@@ -139,7 +140,11 @@ def _verify_channel(self) -> None:
 
         :raises ApplicationError: Exception raised when channel is not a valid OpenStack channel.
         """
-        if self.is_from_charm_store or self.is_valid_track(self.channel):
+        if (
+            self.is_from_charm_store
+            or self.channel == LATEST_STABLE
+            or self.is_valid_track(self.channel)
+        ):
```
rgildein, line 147: It's not responsibility of these PR, but we should move this `_verify_channel` to sanity checks.

rgildein (commented):
I like this, because it makes the code more readable. Some small comments.

cou/apps/auxiliary.py
gabrielcocenza: It's just to be sure that for a certain charm in a ubuntu series and in a OpenStack release there is at least one possible track. I've changed a little bit and now is checking `len(possible_tracks) > 0
`

cou/apps/base.py
gabrielcocenza: Yeah, I think it doesn't make sense to have private property. I've changes and put it next to the other properties

cou/apps/base.py
```diff
@@ -383,6 +401,8 @@ def pre_upgrade_steps(
         """
         return [
             self._get_upgrade_current_release_packages_step(units),
+            self._get_charm_hub_migration(),
+            self._get_change_to_openstack_channels(),
```
aieri, line 405: I'm not a fan of this style: conceptually as an operator I want to prepare for an upgrade by refreshing to the latest available charm for the current OpenStack version. This can mean three things, depending on the situation (but they all fall into the same "refresh the current version" category from my point of view):
* in-channel refresh
* switch to a different channel
  * switch from charmstore to charmhub
  * switch from ch:latest/stable to a release-specific channel

The way you wrote it, you're delegating to the step-generating functions to do things if needed, which obscures the fact that the pre-upgrade steps will actually contain 2 steps, not 4.

I think it would be clearer to extend the current `_get_refresh_charm_step` function to allow switching from `latest/stable` (possibly by splitting into different nested sub-functions)

tests/unit/apps/test_base.py
```diff
@@ -485,3 +485,127 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+    assert app._get_charm_hub_migration() == PreUpgradeStep(
+        f"Migrate '{app.name}' from charmstore to charmhub",
+        coro=model.upgrade_charm(app.name, app.possible_current_channel, switch=f"ch:{app.charm}"),
+    )
+
+
+@patch("cou.apps.base.OpenStackApplication._verify_channel", return_value=None)
+@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
+def test_get_change_to_openstack_channels_latest_stable(current_os_release, _, model):
+    """Application using latest/stable channel."""
```
aieri, line 518: For clarity, you could explain what's the expectation is, e.g. "Applications using latest/stable should be switched to a release-specific channel"

tests/unit/apps/test_base.py
```diff
@@ -485,3 +485,127 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+        f"WARNING: Changing '{app.name}' channel from {app.channel} to "
+        f"{app.possible_current_channel}. COU cannot guarantee that this is not a downgrade",
+        coro=model.upgrade_charm(app.name, app.possible_current_channel),
+    )
+
+
+@patch("cou.apps.base.OpenStackApplication._verify_channel", return_value=None)
+@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
+def test_get_change_to_openstack_channels_openstack_channel(current_os_release, _, model):
+    """Application already on OpenStack channel."""
```
aieri, line 534: as above

tests/unit/apps/test_base.py
```diff
@@ -485,3 +485,127 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+        app_name, "", app_name, "ussuri/stable", {}, {}, model, "ch", "focal", [], {}, "1"
+    )
+    assert app._get_refresh_charm_step(target) == PreUpgradeStep()
+
+
+@patch("cou.apps.base.OpenStackApplication._verify_channel", return_value=None)
+@patch("cou.apps.base.OpenStackApplication.can_refresh", new_callable=PropertyMock)
+@patch("cou.apps.base.OpenStackApplication._need_to_refresh")
+def test_get_refresh_charm_step(mock_need_to_refresh, mock_can_refresh, _, model):
+    """Application does not need to refresh."""
```
aieri, line 567: it looks like it does :)

tests/unit/apps/test_base.py
```diff
@@ -485,3 +485,127 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+    assert app.can_refresh is exp_result
+
+
+@pytest.mark.parametrize(
+    "can_upgrade_to, channel, exp_result",
+    [
+        ("", "ussuri/stable", False),
+        ("ch:amd64/focal/my-app-723", "ussuri/stable", True),
+        ("ch:amd64/focal/my-app-723", "victoria/stable", True),
+        # when channel_codename is bigger than target it's not necessary to refresh
```
aieri, line 599: thank you! this comment is definitely useful

tests/unit/apps/test_auxiliary.py
```diff
@@ -1171,12 +1188,19 @@ async def test_ceph_osd_verify_nova_compute_fail(mock_lookup, model):
...
         subordinate_to=[],
-        units={},
+        units={
+            f"ceph-osd/{i}": Unit(
+                name=f"ceph-osd/{i}",
+                workload_version="17.0.1",
+                machine=machines[f"{i}"],
+            )
+            for i in range(3)
+        },
```
aieri, line 1203: [non-blocking] I don't understand how these parts relate to this PR

cou/apps/base.py
```diff
@@ -516,58 +536,74 @@ def _get_upgrade_current_release_packages_step(
...
         :return: Step for refreshing the charm.
         :rtype: PreUpgradeStep
         """
-        if not self.can_upgrade_to:
-            return PreUpgradeStep()
-
-        switch = None
-        *_, channel = self.possible_current_channels
-
-        # corner case for rabbitmq and hacluster.
```
aieri, line 536: where did this corner case end up?

cou/apps/base.py
```diff
@@ -236,13 +241,15 @@ def origin_setting(self) -> Optional[str]:
         return None
 
     @property
-    def possible_current_channels(self) -> list[str]:
-        """Return the possible current channels based on the current OpenStack release.
+    def possible_current_channel(self) -> str:
```
aieri, line 244: :thinking: if there's only one is it still a "possible" channel? What are we trying to represent here? The channel the application should be using based on the workload version?

cou/apps/base.py
```diff
@@ -236,13 +241,15 @@ def origin_setting(self) -> Optional[str]:
...
+        """Return the possible current channel.
+
+        Possible current channel is based in the current OpenStack release of the application.
```
aieri, line 247: nit
```suggestion
        Possible current channel is based on the current OpenStack release of the application.
```

cou/apps/base.py
```diff
@@ -258,6 +265,17 @@ def os_release_units(self) -> dict[OpenStackRelease, list[str]]:
 
         return dict(os_versions)
 
+    @property
+    def can_refresh(self) -> bool:
+        """Check if the application can refresh.
+
+        Application can refresh if are not from charm store and it's not using
```
aieri, line 272: nit
```suggestion
        The application can be refreshed if it is not coming from charm store and it's not using the
```

cou/apps/base.py
```diff
@@ -516,58 +536,74 @@ def _get_upgrade_current_release_packages_step(
 
         return step
 
+    def _get_charm_hub_migration(self) -> PreUpgradeStep:
```
aieri, line 539: nit: charm_hub → charmhub

cou/apps/base.py
```diff
@@ -516,58 +536,74 @@ def _get_upgrade_current_release_packages_step(
...
+        logger.warning(
+            "COU cannot guarantee that it's not a downgrade changing channel of '%s' "
+            "from %s to %s.",
+            self.name,
+            self.channel,
+            self.possible_current_channel,
+        )
+        return PreUpgradeStep(
+            f"WARNING: Changing '{self.name}' channel from {self.channel} to "
+            f"{self.possible_current_channel}. COU cannot guarantee that this is not a downgrade",
```
aieri, line 572: Just a suggestion to avoid the double negation
```suggestion
            f"{self.possible_current_channel}. This may be a charm downgrade, which is generally not supported.",
```

cou/apps/base.py
```diff
@@ -516,58 +536,74 @@ def _get_upgrade_current_release_packages_step(
...
+        """Get the step for changing to OpenStack channels if necessary.
+
+        :return: Step for changing to OpenStack channels
+        :rtype: PreUpgradeStep
+        """
+        if self.channel != LATEST_STABLE:
+            return PreUpgradeStep()
+        logger.warning(
+            "COU cannot guarantee that it's not a downgrade changing channel of '%s' "
+            "from %s to %s.",
```
aieri, line 565: nit: why not use the same message as in line 571?

aieri (changes requested):

rgildein (dismissed):
LGTM

tests/unit/apps/test_auxiliary.py
gabrielcocenza: Long story short: the changes does not direct affect this, but I was playing around and identified this issue

Long story:
I was making some changes on `is_valid_track` for auxiliary charms and in some point I started checking if the current track was inside the possible_tracks ``and some unit tests started failing because on theory you can't have ceph-osd with workload version on `pacific` (N)  and the channel on `octopus` (N-1), so I've changed the workload_version to be properly on octopus.

After some though I realize that we can't be strict like this because on a transition (E.g: octopus to pacific) you can have a channel that does not match with the current workload_version. It would make sense to check that the channel is on N or N+1 in relation to the workload but never N-1 like it was here. This check is not easy to implement right now,

cou/apps/base.py
gabrielcocenza: Now it's on `possible_current_channel` on the auxiliary.

cou/apps/auxiliary.py
```diff
@@ -53,31 +53,29 @@ def is_valid_track(self, charm_channel: str) -> bool:
...
-
-        raise ApplicationError(
-            (
-                f"Cannot find a suitable '{self.charm}' charm channel for "
-                f"{self.current_os_release.codename} on series '{self.series}'. "
-                "Please take a look at the documentation: "
-                "https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html"
-            )
+        *_, track = OPENSTACK_TO_TRACK_MAPPING.get(
+            (self.charm, self.series, self.current_os_release.codename), []
```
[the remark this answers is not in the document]
gabrielcocenza: Thanks for pointing this. Because `is_valid_track`already checks that exists some track, we can access here directly without the need of using `.get` or worry with keyerror

cou/apps/base.py
gabrielcocenza: > The channel the application should be using based on the workload version?

Yes. I think you are right on questioning this. Maybe changing it to `expected_current_channel` would be more appropriate

cou/apps/auxiliary.py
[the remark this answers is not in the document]
agileshaw: In that case let's remove the fallback or even `.get` to avoid confusion.

cou/apps/base.py
aieri: In the end it was just a info-level logline so not a big deal, but I still don't see where that (`len(self.possible_current_channels) > 1`) condition has been relocated

cou/apps/base.py
aieri: this is better now, although I would personally go a step further and move those if conditions on the calling side. Something like
```
def _get_refresh_charm_step(self, target: OpenStackRelease) -> PreUpgradeStep:
    if self.channel == LATEST_STABLE:
        step = self._get_change_to_openstack_channels_step()
    elif not self.is_from_charm_store:
        step = self._get_charmhub_migration_step()
    # etc
    return step
```
Anyway, this is now non-blocking from my point of view

aieri (dismissed):

tests/unit/apps/test_base.py
```diff
@@ -487,107 +487,234 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
 
 
-@patch("cou.apps.base.OpenStackApplication._verify_channel", return_value=None)
-@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
-def test_get_change_to_openstack_channels_openstack_channel(current_os_release, _, model):
-    """Application already on OpenStack channel."""
-    current_os_release.return_value = OpenStackRelease("ussuri")
-    app_name = "app"
+def test_get_refresh_current_channel_step(model):
+    """Expect an pre-upgrade for application that does need to refresh the current channel."""
```
agileshaw, line 543:
```suggestion
    """Expect a pre-upgrade step for application that needs to refresh the current channel."""
```

cou/apps/base.py
```diff
@@ -538,59 +527,62 @@ def _get_upgrade_current_release_packages_step(
...
+        :return: Step for refreshing the charm
+        :rtype: PreUpgradeStep
+        """
+        if self.is_from_charm_store:
+            return self._get_charmhub_migration_step()
+        if self.channel == LATEST_STABLE:
+            return self._get_change_to_openstack_channels_step()
+        if self._need_to_refresh(target):
+            return self._get_refresh_current_channel_step()
+        return PreUpgradeStep()
```
agileshaw, line 545: nit: I think it would be useful to have an info-level log message informing the user that the charm does't need to be refreshed.

tests/unit/apps/test_base.py
```diff
@@ -487,107 +487,234 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+@patch("cou.apps.base.OpenStackApplication._get_change_to_openstack_channels_step")
+@patch("cou.apps.base.OpenStackApplication._get_charmhub_migration_step")
 def test_get_refresh_charm_step_skip(
-    mock_need_to_refresh, mock_can_refresh, model, can_refresh, need_to_refresh
+    mock_ch_migration, mock_change_os_channels, mock_refresh_current_channel, model
 ):
-    """Application does not need to refresh."""
-    mock_need_to_refresh.return_value = need_to_refresh
-    mock_can_refresh.return_value = can_refresh
+    """Expect an empty pre-upgrade for application that does not need to refresh."""
```
agileshaw, line 572:
```suggestion
    """Expect an empty pre-upgrade step for application that does not need to refresh."""
```

tests/unit/apps/test_base.py
```diff
@@ -487,107 +487,234 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
-    mock_can_refresh.return_value = True
+@patch("cou.apps.base.OpenStackApplication._get_refresh_current_channel_step")
+@patch("cou.apps.base.OpenStackApplication._get_change_to_openstack_channels_step")
+@patch(
+    "cou.apps.base.OpenStackApplication._get_charmhub_migration_step",
+)
+def test_get_refresh_charm_step_refresh_current_channel(
+    mock_ch_migration, mock_change_os_channels, mock_refresh_current_channel, model
+):
+    """Expect an pre-upgrade for application that does need to refresh current channel."""
```
agileshaw, line 603:
```suggestion
    """Expect a pre-upgrade step for application that needs to refresh current channel."""
```

tests/unit/apps/test_base.py
```diff
@@ -487,107 +487,234 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+@patch("cou.apps.base.OpenStackApplication._get_charmhub_migration_step")
+@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
+def test_get_refresh_charm_step_change_to_openstack_channels(
+    current_os_release,
+    mock_ch_migration,
+    mock_change_os_channels,
+    mock_refresh_current_channel,
+    model,
+):
+    """Expect an pre-upgrade for application that does need to change to OpenStack channel."""
```
agileshaw, line 644:
```suggestion
    """Expect a pre-upgrade step for application that needs to change to OpenStack channel."""
```

tests/unit/apps/test_base.py
```diff
@@ -487,107 +487,234 @@ def test_check_mismatched_versions(mock_os_release_units, model):
...
+@patch("cou.apps.base.OpenStackApplication._get_charmhub_migration_step")
+@patch("cou.apps.base.OpenStackApplication.current_os_release", new_callable=PropertyMock)
+def test_get_refresh_charm_step_charmhub_migration(
+    current_os_release,
+    mock_ch_migration,
+    mock_change_os_channels,
+    mock_refresh_current_channel,
+    model,
+):
+    """Expect an pre-upgrade for application that does need to migrate to charmhub."""
```
agileshaw, line 689:
```suggestion
    """Expect a pre-upgrade step for application that needs to migrate to charmhub."""
```

cou/apps/base.py
```diff
@@ -519,57 +528,76 @@ def _get_upgrade_current_release_packages_step(
...
-            )
-            return PreUpgradeStep()
-        elif self.channel not in self.possible_current_channels:
-            raise ApplicationError(
-                f"'{self.name}' has unexpected channel: '{self.channel}' for the current workload "
-                f"version and OpenStack release: '{self.current_os_release.codename}'. "
-                f"Possible channels are: {','.join(self.possible_current_channels)}"
-            )
+    def _get_change_to_openstack_channels_step(self) -> PreUpgradeStep:
+        """Get the step for changing to OpenStack channels if necessary.
```
agileshaw, line 561: Now that we moved the conditional logic outside the function, we should update the docstring: 
```suggestion
        """Get the step for changing to OpenStack channels.
```

cou/apps/base.py
```diff
@@ -519,57 +528,76 @@ def _get_upgrade_current_release_packages_step(
...
+        :return: Step for refreshing the charm
         :rtype: PreUpgradeStep
         """
-        if not self.can_upgrade_to:
-            return PreUpgradeStep()
+        if self.is_from_charm_store:
+            return self._get_charmhub_migration_step()
+        if self.channel == LATEST_STABLE:
+            return self._get_change_to_openstack_channels_step()
+        if self._need_to_refresh(target):
```
agileshaw, line 543: nit: we are using the word `refresh` as a broader meaning in this function, so it's unclear what `_need_to_refresh` function would do from the name. Could you rename it to something like `_need_in_channel_refresh`?

cou/apps/base.py
```diff
@@ -519,57 +528,76 @@ def _get_upgrade_current_release_packages_step(
...
+            return self._get_refresh_current_channel_step()
```
agileshaw, line 544: nit: also here, maybe better with a name like `_get_in_channel_refresh()`

agileshaw (changes requested):
functionality-wise looks good. Just some typos that need to be fixed as well as some additional nit suggestions stated inline.

cou/apps/base.py
gabrielcocenza: I've changed to `_need_current_channel_refresh`. I think that makes it clear. Thanks

agileshaw: Thanks

agileshaw (approved):
LGTM

aieri (approved):

cou/apps/auxiliary.py
```diff
@@ -53,31 +53,31 @@ def is_valid_track(self, charm_channel: str) -> bool:
...
-        :rtype: list[str]
-        :raises ApplicationError: When cannot find tracks.
+        Expected current channel is the channel that the application is suppose to be using based
+        on the current series, workload version and, by consequence, the OpenStack release
+        identified.
+        :return: The expected current channel of the application. E.g: "3.9/stable"
+        :rtype: str
         """
-        tracks = OPENSTACK_TO_TRACK_MAPPING.get(
+        *_, track = OPENSTACK_TO_TRACK_MAPPING[
```
rgildein, line 76: not-blocking: I know that we are using `is_valid_track` before these, but if I call in another place without checking if track is valid, it can raise ValueError, and I think we should mention it in docstring.

rgildein (approved):
Before you merge it, can you add description for PR, thanks.

Issue #326, opened 2024-03-25
COU is not logging to the user hypervisors that have VMs running


gabrielcocenza:
Currently, the implementation it's not logging non-empty hypervisors. I guess we should also log if the user uses the `--force` option

PR #351, opened 2024-04-02
Add log to inform skipped(non-empty) hypervisors

zxhdaze:
This PR depends on https://github.com/canonical/charmed-openstack-upgrader/pull/334. Please consider only the latest commit for review until the dependent PR is merged.

Add log to inform user about skipped hypervisors(those having VMs running) at plan stage.

fixes: https://github.com/canonical/charmed-openstack-upgrader/issues/326


tests/mocked_plans/test_base_plan.py
```diff
@@ -32,3 +33,25 @@ async def test_base_plan(_, sample_plans):
...
+
+    with caplog.at_level(logging.INFO):
+        analysis_results = await Analysis.create(model)
+        await generate_plan(analysis_results, args)
+
+    expected_log = (
+        "Skipped (non-empty) hypervisors: [\"Machine(machine_id='1', apps=('nova-compute', "
+        "'ovn-chassis'), az='az-0')\", \"Machine(machine_id='3', apps=('nova-compute', "
+        "'ovn-chassis'), az='az-0')\"]"
+    )
```
aieri, line 53: this is a little cluttered because hypervisors will by definition host a nova-compute app, and because it's not that relevant for an operator if they also host other units. I suggest tweaking the output to only report the nova-compute unit name. So something like
```
Skipped (non-empty) hypervisors: nova-compute/0, nova-compute/1
```

agileshaw (commented):
Just a reminder to link this PR to the issue it is addressing ([ref](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests))

rgildein:
@zxhdaze Can you rebase this on current main, I merged #334. Thanks

tests/mocked_plans/test_base_plan.py
```diff
@@ -32,3 +33,27 @@ async def test_base_plan(_, sample_plans):
...
+async def test_base_plan_non_empty(_, sample_plans, caplog):
+    """Testing non-empty hypervisor logs."""
+    args = CLIargs("plan", auto_approve=True)
+    model, __ = sample_plans["base.yaml"]
+
+    with caplog.at_level(logging.INFO):
+        analysis_results = await Analysis.create(model)
+        await generate_plan(analysis_results, args)
+
+    expected_skipped_log = "Skipped (non-empty) hypervisors: nova-compute/0, nova-compute/2"
```
gabrielcocenza, line 49: I think it's better to use something like:

```python
expected_skipped_log = ("'cou.steps.plan'", logging.INFO, "Skipped (non-empty) hypervisors: nova-compute/0, nova-compute/2")
```
and then:

```python
assert expected_skipped_log in caplog.record_tuples
```

This way we guarantee the place that the logger was used, the log level and the message

cou/steps/plan.py
```diff
@@ -596,7 +595,27 @@ async def _get_upgradable_hypervisors_machines(
     if cli_force:
         return nova_compute_machines
 
-    return await get_empty_hypervisors(nova_compute_units, analysis_result.model)
+    empty_hypervisors = await get_empty_hypervisors(nova_compute_units, analysis_result.model)
+
+    machine_to_unit_name = {
+        machine.machine_id: unit.name
+        for unit, machine in zip(nova_compute_units, nova_compute_machines)
+    }
```
gabrielcocenza, line 603: I think you can change  `_get_nova_compute_units_and_machines` to be a dictionary contains machines object as keys and nova-compute units as values. It would be more handy than constructing the dictionary here

cou/steps/plan.py
```diff
@@ -596,7 +595,27 @@ async def _get_upgradable_hypervisors_machines(
...
+    empty_hypervisor_ids = {machine.machine_id for machine in empty_hypervisors}
+    all_hypervisor_ids = {machine.machine_id for machine in nova_compute_machines}
+    skipped_hypervisor_ids = all_hypervisor_ids - empty_hypervisor_ids
```
gabrielcocenza, line 606: I would prefer doing this on the object level instead of just using the ids of the objects

cou/steps/plan.py
```diff
@@ -596,7 +595,27 @@ async def _get_upgradable_hypervisors_machines(
     if cli_force:
         return nova_compute_machines
```
gabrielcocenza, line 596: Just for consistency, let's also log in this occasion and let the user know that all  hypervisors were selected. Something like

```python
logger.info("Selected all hypervisors: %s", all_hypervisors)
```

cou/steps/plan.py
```diff
@@ -596,7 +595,27 @@ async def _get_upgradable_hypervisors_machines(
...
+
+    skipped_unit_names = ", ".join(
+        machine_to_unit_name[machine_id] for machine_id in sorted(skipped_hypervisor_ids)
+    )
+    logger.info("Skipped (non-empty) hypervisors: %s", skipped_unit_names)
+
+    selected_unit_names = ", ".join(
+        machine_to_unit_name[machine_id] for machine_id in sorted(empty_hypervisor_ids)
+    )
+    logger.info("Selected (empty) hypervisors: %s", selected_unit_names)
```
gabrielcocenza, line 616: see how lines 614 and 609 are doing the same thing and can be considered code of lower level than the other lines. I suggest creating a helper function and then use something like:

```python
logger.info("Selected (empty) hypervisors: %s", _stringify_units(selected_unit_names))
```

gabrielcocenza (changes requested):

tests/mocked_plans/test_base_plan.py
```diff
@@ -32,3 +33,27 @@ async def test_base_plan(_, sample_plans):
     plan = await generate_plan(analysis_results, args)
 
     assert str(plan) == exp_plan
+
+
+@pytest.mark.asyncio
+@patch("cou.utils.nova_compute.get_instance_count", side_effect=[1, 0, 1])
+async def test_base_plan_non_empty(_, sample_plans, caplog):
```
rgildein, line 40: I do not see why this test end up here? I think it should be part of unit tests and not mocked-plans, since we are not testing plan here.

The test itself could be here if we drop part for first and last hypervisors, e.g something like  `exp_plan.replace("...", "")`.

rgildein (changes requested):

tests/mocked_plans/test_base_plan.py
zxhdaze: Should I create a separate config for this test, or is it okay to just keep sharing 'base.yaml' with mocked plans?

cou/steps/plan.py
zxhdaze: Should I put the helper function under utils/nova_compute.py ?

cou/steps/plan.py
aieri: I am confused... `_get_nova_compute_units_and_machines` returns
```
return nova_compute_units, [unit.machine for unit in nova_compute_units]
```
which seems kinda useless...
And then we zip them back together here (?)
...only to reconstruct the unit name from the machine id.

Isn't it overly convoluted?

`get_empty_hypervisors` is used only once (aside from here) so it's easy to refactor. If we change it to return the set of empty hypervisor _units_, all of this should simplify, I think, to

```
hypervisor_units, _ = _get_nova_compute_units_and_machines(
    analysis_result.apps_data_plane
) # or hypervisor_units = _get_nova_compute_units(analysis_result.apps_data_plane)
empty_hypervisor_units = await get_empty_hypervisors(hypervisor_units, analysis_result.model)
non_empty_hypervisor_units = hypervisor_units - empty_hypervisor_units # make them sets at some point

stringify_units = lambda x: ", ".join([u.name for u in x])

skipped_unit_names = stringify_units(non_empty_hypervisor_units)

# etc
```

cou/steps/plan.py
aieri: you could use a lambda for such a small helper

aieri (changes requested):

cou/steps/plan.py
rgildein, line 616: Why this is not part of `get_empty_hypervisors` and like that, we would not need to have `machine_to_unit_name` or any issue what `get_empty_hypervisors` should return?

If we move it there, we should not use words like `Skipped` and `Selected`, but we can use `Found not-empty ...`. Also, the message with empty hypervisors should not be warning level?

cou/steps/plan.py
rgildein: Should logos listed below be part of `get_empty_hypervisors` and like that, we would not need to have `machine_to_unit_name` or any issue what `get_empty_hypervisors` should return?

If we move it there, we should not use words like `Skipped` and `Selected`, but we can use `Found not-empty ...`. Also, the message with empty hypervisors should not be warning level?

aieri: yes, logging within `get_empty_hypervisors` would be easier, but not fully correct from my point of view. The point here is informing the user that some hypervisors that they expected to upgrade will not actually be upgraded, therefore the logging should occur within the plan stage, not within a utils function.

Why do you think the log level should be 'warning'?

tests/unit/steps/test_plan.py
```diff
@@ -1575,3 +1576,66 @@ def test_create_upgrade_plan_failed():
...
+    analysis_result = MagicMock(spec_set=Analysis)()
+
+    mock_all_hypervisors.return_value = all_units, all_machines
+    mock_empty_hypervisors.return_value = empty_units, empty_machines
+
+    with caplog.at_level(logging.INFO):
+        await cou_plan._get_upgradable_hypervisors_machines(cli_force, analysis_result)
+
+    for expected_log in expected_logs:
+        assert expected_log in caplog.record_tuples
```
zxhdaze, line 1641: I ended up rewriting the test and creating a separate config for it, for better flexibility. I think the test should focus on testing the logging logic, instead of the plan. Using a customized config can make it easier to control arguments such as cli_force.

cou/utils/nova_compute.py
```diff
@@ -85,3 +94,15 @@ async def verify_empty_hypervisor(unit: Unit, model: Model) -> None:
...
+def stringify_units(units: Iterable[Unit]) -> str:
+    """Convert Units into a comma-separatedstring of unit names, sorted alphabetically.
+
+    :param units: A iterable of Unit objects to be converted.
+    :type units: Iterable[Unit]
+    :return: A comma-separated string of sorted unit names.
+    :rtype: str
+    """
+    sorted_unit_names = sorted([unit.name for unit in units], key=lambda name: name)
+    return ", ".join(sorted_unit_names)
```
zxhdaze, line 108: Due to restrictions from the lint checker on using a lambda, I ended up adding the helper function here.

tests/unit/steps/test_plan.py
```diff
@@ -1575,3 +1576,66 @@ def test_create_upgrade_plan_failed():
...
+                ),
+                ("cou.steps.plan", logging.INFO, "Selected hypervisors: nova-compute/1"),
+            ],
+        ),
+    ],
+)
+@pytest.mark.asyncio
+@patch("cou.steps.plan._get_nova_compute_units_and_machines")
+@patch("cou.steps.plan.get_empty_hypervisors")
+async def test_log_upgradable_hypervisors(
```
zxhdaze, line 1612: I ended up rewriting the test and creating a separate config for it, for better flexibility. I think the test should focus on testing the logging logic, instead of the plan. Using a customized config can make it easier to control arguments such as cli_force.

cou/steps/plan.py
```diff
@@ -594,9 +594,18 @@ async def _get_upgradable_hypervisors_machines(
...
         return nova_compute_machines
 
-    return await get_empty_hypervisors(nova_compute_units, analysis_result.model)
+    empty_units, empty_machines = await get_empty_hypervisors(
+        nova_compute_units, analysis_result.model
+    )
+    skipped_units = set(nova_compute_units) - set(empty_units)
+
+    logger.info("Found non-empty hypervisors: %s", stringify_units(skipped_units))
+    logger.info("Selected hypervisors: %s", stringify_units(empty_units))
```
rgildein, line 606: not-blocking: I really do not see much benefit not doing these logs directly in `get_empty_hypervisors`.

cou/utils/nova_compute.py
```diff
@@ -85,3 +94,15 @@ async def verify_empty_hypervisor(unit: Unit, model: Model) -> None:
...
+
+def stringify_units(units: Iterable[Unit]) -> str:
+    """Convert Units into a comma-separatedstring of unit names, sorted alphabetically.
+
+    :param units: A iterable of Unit objects to be converted.
+    :type units: Iterable[Unit]
+    :return: A comma-separated string of sorted unit names.
+    :rtype: str
+    """
+    sorted_unit_names = sorted([unit.name for unit in units], key=lambda name: name)
```
rgildein, line 107: The key is unnecessary here, e.g. `sorted([unit.name for unit in units], key=lambda name: name)` is the same as `sorted([unit.name for unit in units])`

cou/utils/nova_compute.py
```diff
@@ -85,3 +99,15 @@ async def verify_empty_hypervisor(unit: Unit, model: Model) -> None:
             unit_instance_count,
         )
         raise HaltUpgradeExecution(f"Unit: {unit.name} has {unit_instance_count} VMs running")
+
+
+def stringify_units(units: Iterable[Unit]) -> str:
+    """Convert Units into a comma-separatedstring of unit names, sorted alphabetically.
```
agileshaw, line 105:
```suggestion
    """Convert Units into a comma-separated string of unit names, sorted alphabetically.
```

cou/utils/nova_compute.py
```diff
@@ -85,3 +99,15 @@ async def verify_empty_hypervisor(unit: Unit, model: Model) -> None:
...
+
+    :param units: A iterable of Unit objects to be converted.
```
agileshaw, line 107:
```suggestion
    :param units: An iterable of Unit objects to be converted.
```

cou/utils/nova_compute.py
```diff
@@ -85,3 +99,15 @@ async def verify_empty_hypervisor(unit: Unit, model: Model) -> None:
             unit_instance_count,
         )
         raise HaltUpgradeExecution(f"Unit: {unit.name} has {unit_instance_count} VMs running")
+
+
+def stringify_units(units: Iterable[Unit]) -> str:
```
agileshaw, line 104: non-blocking: If we are only going to use this function here, let's mark it as a private function
```suggestion
def _get_stringify_units(units: Iterable[Unit]) -> str:
```

Also do we really need the argument type to be `Iterable` instead of `set`?

tests/unit/utils/test_nova_compute.py
```diff
@@ -66,13 +66,19 @@ async def test_get_instance_count_invalid_result(model, result_key, value):
...
+    non_empty_log, selected_log = construct_expected_logs(hypervisors_count)
+
+    with caplog.at_level(logging.INFO):
+        result = await nova_compute.get_empty_hypervisors(
+            [_mock_nova_unit(nova_unit) for nova_unit, _ in hypervisors_count], model
+        )
+
     assert {machine.machine_id for machine in result} == expected_result
+    assert non_empty_log in caplog.record_tuples
+    assert selected_log in caplog.record_tuples
```
agileshaw, line 81: IMO this is overly complex. You could simply mock the logger and check if it has been called with the expected messages. See this test as an example: https://github.com/canonical/charmed-openstack-upgrader/blob/main/tests/unit/test_cli.py#L130-L138

agileshaw (changes requested):

rgildein (approved):
LGTM, thanks

cou/utils/nova_compute.py
```diff
@@ -36,7 +37,20 @@ async def get_empty_hypervisors(units: list[Unit], model: Model) -> list[Machine
...
+    empty_machines = []
+
+    for unit, instance_count in units_instances:
+        if instance_count == 0:
+            empty_units.add(unit)
+            empty_machines.append(unit.machine)
+
+    skipped_units = set(units) - empty_units
+
+    logger.info("Found non-empty hypervisors: %s", stringify_units(skipped_units))
```
gabrielcocenza, line 50: I think this might be a warning log. I think it's good to show it by default to the users

gabrielcocenza: There's also the issue that this is going to happen even that there are no empty_units. I'm not a huge fan of this, but we might need to add a if condition to log it:

```python
if skipped_units:
    logger.warning(....)
```

gabrielcocenza (approved):
LGTM. Thanks

tests/unit/utils/test_nova_compute.py
```diff
@@ -64,16 +63,31 @@ async def test_get_instance_count_invalid_result(model, result_key, value):
...
+    )
+    non_empty_hypervisors = ", ".join(
+        [f"nova-compute/{i}" for i, count in hypervisors_count if count != 0]
+    )
+
+    mock_logger.info.assert_has_calls([call("Selected hypervisors: %s", selected_hypervisors)])
+    if non_empty_hypervisors:
+        mock_logger.warning.assert_has_calls(
+            [call("Found non-empty hypervisors: %s", non_empty_hypervisors)]
+        )
```
agileshaw, line 89: non-blocking: since they only have one call each, we could simply use `assert_called_once_with`

agileshaw (approved):
LGTM, with one non-blocking suggestion

agileshaw (approved):
+1 from me. Thanks :)

gabrielcocenza (approved):

rgildein (approved):

aieri (approved):

Issue #218, opened 2024-01-16
docs: project name does not fit the top left banner element


aieri:
As shown in the screenshot below, the project name in the top left is too long and is partially covered by the element right to it.

![image](https://github.com/canonical/charmed-openstack-upgrader/assets/4036574/2a9afac7-0999-4e7e-abc4-4a8e23971609)

---
*Reported from: https://canonical-charmed-openstack-upgrader.readthedocs-hosted.com/en/latest/*

PR #354, opened 2024-04-03
Adjust RTD site header

agileshaw:
Move the explicitly displayed snapcraft link from header to the "More Resources" dropdown list.

Also remove "Forum" from the dropdown list because COU doesn't have a discourse page.

fixes: #218


aieri (approved):
thanks!

rgildein (approved):
Easy and nice fix, thanks!

Issue #348, opened 2024-04-02
Not all OVN applications has enable-versiong-pinning


gabrielcocenza:
OVN central does not have this config and COU plan generation will fail

PR #359, opened 2024-04-04
Support OVN apps that don't have "enable-versiong-pinning" config

gabrielcocenza:
- Not all ovn apps has this configuration and the current implementation it's not permesive on this.

Closes: #348

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -237,6 +237,12 @@ class Ovn(AuxiliaryApplication):
 
         :raises ApplicationError: When version pinning is True
         """
+        if "enable-version-pinning" not in self.config:
+            logger.debug(
+                "Ovn application: '%s' does not have the 'enable-version-pinning' configuration.",
+                self.name,
+            )
+            return
         if self.config["enable-version-pinning"].get("value"):
             raise ApplicationError(
                 f"Cannot upgrade '{self.name}'. 'enable-version-pinning' must be set to 'false'."
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -218,8 +218,8 @@ class CephMon(AuxiliaryApplication):
         )
 
 
-class Ovn(AuxiliaryApplication):
-    """Ovn generic application class."""
+class OVN(AuxiliaryApplication):
+    """OVN generic application class."""
 
     @abc.abstractmethod
     def _check_ovn_support(self) -> None:
@@ -252,7 +252,7 @@ class Ovn(AuxiliaryApplication):
         """
         if "enable-version-pinning" not in self.config:
             logger.debug(
-                "Ovn application: '%s' does not have the 'enable-version-pinning' configuration.",
+                "OVN application: '%s' does not offer the 'enable-version-pinning' configuration.",
                 self.name,
             )
             return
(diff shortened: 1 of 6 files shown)


tests/unit/apps/test_auxiliary.py
```diff
@@ -870,6 +870,87 @@ def test_ovn_no_compatible_os_release(channel, model):
...
+
+
+def test_ovn_check_version_pinning_version_pinning_config_True(model):
+    machines = {"0": MagicMock(spec_set=Machine)}
+    app = OvnPrincipal(
+        name="ovn-central",
+        can_upgrade_to="",
+        charm="ovn-central",
+        channel="22.03/stable",
+        config={"source": {"value": "distro"}, "enable-version-pinning": {"value": True}},
```
agileshaw, line 934: non-blocking: I understand that you want to keep everything else, including the name of the charm, the same to test the `enable-version-pinning` configuration, but I find it misleading to test "ovn-central" with `enable-version-pinning` config when it is not actually available. But this is by no mean a blocker, just want to point it out in case anyone has a better idea.

agileshaw (approved):

tests/unit/apps/test_auxiliary.py
gabrielcocenza: You are right. It's misleading and I'll change to ovn-dedicated-chassis when it have the config. Thanks

agileshaw: Looks good, thanks!

agileshaw (dismissed):
LGTM

cou/apps/auxiliary.py
```diff
@@ -237,6 +237,12 @@ def _check_version_pinning(self) -> None:
 
         :raises ApplicationError: When version pinning is True
         """
+        if "enable-version-pinning" not in self.config:
+            logger.debug(
+                "Ovn application: '%s' does not have the 'enable-version-pinning' configuration.",
```
aieri, line 242:
```suggestion
                "OVN application '%s' does not offer the 'enable-version-pinning' configuration option.",
```
I also don't quite get the quotes around the application name but I think we do that in all log messages

aieri (changes requested):
This is fine, you just need to fix the capitalization of "OVN" (and I also added a tiny wording suggestion).

cou/apps/auxiliary.py
rgildein: In every log and step description.

tests/unit/apps/test_auxiliary.py
```diff
@@ -870,10 +870,92 @@ def test_ovn_no_compatible_os_release(channel, model):
         )
 
 
+@patch("cou.apps.auxiliary.logger")
+def test_ovn_check_version_pinning_no_version_pinning_config(mock_logger, model):
```
rgildein, line 874: Can you join this and next steps with parametrize and `config` as argument, and I'll suggest removing mocking for logger. Logging is only informative, and I do not see reason to check if it was called.

rgildein (changes requested):

rgildein (approved):
LGTM, thanks

agileshaw (dismissed):

aieri (dismissed):

agileshaw (approved):

aieri (approved):

Issue #343, opened 2024-03-28
HypervisorUpgradePlanner needs to run sanity check for each application


rgildein:
Right now, the `HypervisorUpgradePlanner` is not calling sanity check for applications, and we need to do so. 

Also, we need to provide a way (wrapper) to catch an exception from that, something similar to what `_create_upgrade_group` will do after  #333 is merged.

PR #360, opened 2024-04-05
Run sanity checks in HypervisorUpgradePlanner

rgildein:
Run sanity checks for each in HypervisorUpgradePlanner. Improve unit tests for whole HypervisorUpgradePlanner.

fixes: #343


cou/steps/hypervisor.py
```diff
@@ -160,6 +160,31 @@ def get_azs(self, target: OpenStackRelease) -> AZs:
...
+    ) -> None:
+        """Run sanity checks before generating upgrade plan for hypervisor group.
+
+        :param target: OpenStack codename to upgrade.
+        :type target: OpenStackRelease
+        :param group: HypervisorGroup object
+        :type group: HypervisorGroup
+        """
+        for app in self.apps:
+            if app.name not in group.app_units:
```
agileshaw, line 174: It would be simpler iterating on the `group.app_units` directly so we don't need the if condition:
```python
for app in goup.app_units.keys():
```

cou/steps/hypervisor.py
```diff
@@ -269,6 +293,9 @@ def generate_upgrade_plan(self, target: OpenStackRelease, force: bool) -> Upgrad
             hypervisor_plan = HypervisorUpgradePlan(
                 f"Upgrade plan for '{group.name}' to '{target}'"
             )
+            # snity checks
```
agileshaw, line 296:
```suggestion
            # sanity checks
```

cou/steps/hypervisor.py
```diff
@@ -160,6 +160,31 @@ def get_azs(self, target: OpenStackRelease) -> AZs:
...
+                    app.name,
+                    group.name,
+                )
+                continue
+
+            units = group.app_units[app.name]
+            logger.info("running sanoty checks for %s units of %s app", app.name, units)
+            # Note(rgildein): We don't catch the error here because we shouldn't generate any
+            #                 update plan if sanity checks for any application in the group fails.
+            app.upgrade_plan_sanity_checks(target, units)
```
agileshaw, line 186: With the current implementation, hypervisor planning will stop on the first application which fails its sanity check. While this is ok, we miss the opportunity to provide information to user if multiple applications are having issues (e.g. both `nova-compute` and `cinder` have problems but the user is only made aware of one). I think it would be helpful to finish planning for all applications, at least within the same AZ, so the operator can be better informed.

agileshaw (changes requested):

cou/steps/hypervisor.py
rgildein: Yes, but there could be situation that one group do not contain all apps in hypervisor.

cou/steps/hypervisor.py
rgildein: Maybe, but as we agreed it's not important now, and it's safer to simply quit. Because if Cinder have issue, you should not update nova-compute and same for az, if one az failed you should not update next one.

cou/steps/hypervisor.py
```diff
@@ -160,6 +160,31 @@ def get_azs(self, target: OpenStackRelease) -> AZs:
...
+            if app.name not in group.app_units:
+                logger.debug(
+                    "skipping application %s because it is not part of group %s",
+                    app.name,
+                    group.name,
+                )
+                continue
+
+            units = group.app_units[app.name]
+            logger.info("running sanoty checks for %s units of %s app", app.name, units)
```
aieri, line 183:
```suggestion
            logger.info("running sanity checks for %s units of %s app", app.name, units)
```

cou/steps/hypervisor.py
agileshaw: As discussed during the sync, the current implementation makes sense if we have unit-level sanity checks and we want to continue generating plan for other AZs even if one fails. Since the conclusion is that we will not do either of thses, I think it would be better do sanity checks at once at the higher level for each application with their available-to-upgrade units (across all AZs) considered by HypervisorPlanner. I see three benefits of doing so:
1. visually clear that we are doing sanity checks on application level
2. avoid calling the same sanity checks multiple times (once per AZ in the current implementation)
3. with one less for-loop (right now we have two, one for AZ and one for applications), it would be easier to catch exceptions for multiple applications and print them all (though we agreed that this is not a requirement).

This being stated, I wouldn't block this PR just because of this if you feel strongly about keeping the current approach.

cou/steps/hypervisor.py
agileshaw: I didn't realize that we need the Application class instead of just the name, thanks @rgildein for pointing it out in the sync. But I think if you accept my other comment about directly iterating on Applications for all AZs, then we will not have this problem.

Resolving this thread.

cou/steps/hypervisor.py
```diff
@@ -264,6 +277,10 @@ def generate_upgrade_plan(self, target: OpenStackRelease, force: bool) -> Upgrad
         :return: Full upgrade plan
         :rtype: UpgradePlan
         """
+        # snity checks
```
gabrielcocenza, line 280: There is a typo here and I don't think it's necessary. We can remove it

tests/unit/steps/test_hypervisor.py
```diff
@@ -14,56 +14,156 @@
...
+
+    assert steps == exp_steps
+
+
+@patch("cou.steps.hypervisor.HypervisorUpgradePlanner.get_azs")
+@patch("cou.steps.hypervisor.HypervisorUpgradePlanner._upgrade_plan_sanity_checks")
+@patch("cou.steps.hypervisor.HypervisorUpgradePlanner._generate_pre_upgrade_steps")
+@patch("cou.steps.hypervisor.HypervisorUpgradePlanner._generate_upgrade_steps")
+@patch("cou.steps.hypervisor.HypervisorUpgradePlanner._generate_post_upgrade_steps")
+def test_generate_upgrade_plan(
```
gabrielcocenza, line 135: I'm missing an unit tests that shows the expect behavior when an `_upgrade_plan_sanity_checks` fails

gabrielcocenza (changes requested):
This comment is not a blocker, but rather a personal opinion:

- I'm not enjoying the tweaks that we are making because of `_check_mismatched_versions`.
Maybe doing the sanity check by AZ and pass units that are in a certain group from a app could makes things clear.

If we decide to continue like this I won't block but it feels to me that it's becoming more complex to understand the code.

cou/steps/hypervisor.py
```diff
@@ -160,6 +160,20 @@ def get_azs(self, target: OpenStackRelease) -> AZs:
...
+        :param target: OpenStack codename to upgrade.
+        :type target: OpenStackRelease
+        """
+        for app in self.apps:
+            logger.info("running sanity checks for '%s'", app.name)
+            # Note(rgildein): We don't catch the error here because we shouldn't generate any
+            #                 update plan if sanity checks for any application fails.
+            # Note(rgildein): Running sanity checks for all units of applications. We cannot pass
+            #                 None here, since _check_mismatched_versions could be raised.
+            app.upgrade_plan_sanity_checks(target, list(app.units.values()))
```
agileshaw, line 175: I think we should pass in only in-scope units which are from all AZs groups. Also, maybe we should change the implementation of `_check_mismatched_versions` to also check mismatches when units are passed in since even in hypervisors we should have already filtered out the units that are not in the current release.

agileshaw (changes requested):

cou/steps/hypervisor.py
gabrielcocenza: I agree...

rgildein: So I need to iterate over azs second time to get all apps and all units from each group. I really do not see any benefit of this approach.

I opened the issues #365 for second part of your comment.

tests/unit/steps/test_hypervisor.py
rgildein: What is that? If `_upgrade_plan_sanity_checks` raise an exception, this function will fail an exception continues until we catch it. What do you want to test here?

rgildein:
> This comment is not a blocker, but rather a personal opinion:
> 
>     * I'm not enjoying the tweaks that we are making because of `_check_mismatched_versions`.
>       Maybe doing the sanity check by AZ and pass units that are in a certain group from a app could makes things clear.
> 
> 
> If we decide to continue like this I won't block but it feels to me that it's becoming more complex to understand the code.

Me neither.

cou/steps/hypervisor.py
rgildein: > At the time the structure was different and was always checking from all units witch could easily raise mismatched versions.

I was asking during time, when we were introducing units to that check.

tests/unit/steps/test_hypervisor.py
rgildein: I do not see the point of that, since we are not catching any errors here. If any steps, e.g. _generate_pre_upgrade_steps failed, the whole function will fail.

cou/steps/hypervisor.py
agileshaw: Posting a note after our re-discussion.

> sanity checks are designed to be run by unit level

Not really with what we have right now. We are accepting units as arguments in sanity checks but we don't have any checks yet for units themselves. But things will change after #365 being addressed. So I'm +1 on reverting back to doing sanity checks per AZ. I appreciate all these discussions :slightly_smiling_face:

gabrielcocenza (approved):

rgildein:
> LGTM. Sorry for the back-and-forth, but I think it was a valuable discussion.

No problems, I agree.

Issue #361, opened 2024-04-05
Applications with empty origin ("openstack-origin" or "source") breaks COU


gabrielcocenza:
I just deployed a stsstack cloud and seems that vault comes without a origin configured. COU breaks with the following logs:

```
2024-04-05 13:11:22 [cou.apps.base] [DEBUG] vault application current os_release is yoga and apt source is None
2024-04-05 13:11:22 [cou.steps.plan] [ERROR] Cannot generate upgrade plan for 'vault': '>=' not supported between instances of 'NoneType' and 'OpenStackRelease'
2024-04-05 13:11:22 [cou.cli] [ERROR] Unexpected error occurred.
2024-04-05 13:11:22 [cou.cli] [ERROR] '>=' not supported between instances of 'NoneType' and 'OpenStackRelease'
Traceback (most recent call last):
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/cli.py", line 217, in entrypoint
    loop.run_until_complete(_run_command(args))
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/cli.py", line 201, in _run_command
    await get_upgrade_plan(args)
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/cli.py", line 143, in get_upgrade_plan
    upgrade_plan = await analyze_and_plan(args)
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/cli.py", line 131, in analyze_and_plan
    upgrade_plan = await generate_plan(analysis_result, args)
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/steps/plan.py", line 101, in generate_plan
    _generate_control_plane_plan(target, analysis_result.apps_control_plane, args.force)
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/steps/plan.py", line 446, in _generate_control_plane_plan
    principal_upgrade_plan = _create_upgrade_group(
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/steps/plan.py", line 645, in _create_upgrade_group
    if app_upgrade_plan := _generate_instance_plan(app, target, force):
  File "/home/ubuntu/charmed-openstack-upgrader-1/cou/steps/plan.py", line 679, in _generate_instance_plan
    instance_upgrade_plan = instance.generate_upgrade_plan(target, force)
[...]

PR #364, opened 2024-04-08
Consider applications without origin setting to sanity checks

gabrielcocenza:
- Apps without origin setting were breaking COU execution. E.g: vault
- function _need_origin_setting_update check if the application needs to update  origin setting considering the corner cases like apps without origin settings or with empty value.

Closes: #361


cou/apps/base.py
```diff
@@ -794,13 +794,28 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
         if (
             self.current_os_release >= target
             and not self.can_upgrade_to
-            and self.apt_source_codename >= target
+            and not self._need_origin_setting_update(target)
```
aieri, line 797: while this should work, I think a better solution would be altering how `apt_source_codename` behaves. That property is effectively answering the question: "if I were to run `apt upgrade` on this unit, what OpenStack release would my (potentially updated) software be compatible with?"; for units with an origin config it's either parsing the cloudarchive string or resolving series → OpenStack release because the cloudarchive string is converted into a new source entry (and "distro" is not), but why not seeing units without such setting in a similar way? If there's no origin config it simply means that it's not possible to override the unit sources (i.e. the "distro" condition), so `apt_source_codename` for vault or units with an empty origin could be whatever OpenStack release the unit series is compatible with.

[sidenote: if we were to go ahead with this PR as is, line 791 should be updated as well (since `apt_source_codename` can be `None`)]

cou/apps/base.py
```diff
@@ -794,13 +794,28 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
...
+        does not need to change origin setting.
+        Applications with empty origin setting need to update the configuration.
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: True if needs to change origin setting, False otherwise.
+        :rtype: bool
+        """
+        if self.origin_setting and self.os_origin == "":
+            return True
+        return self.apt_source_codename < target if self.apt_source_codename else False
```
rgildein, line 817:
```suggestion
        return self.apt_source_codename is not None and self.apt_source_codename < target
```

cou/apps/base.py
```diff
@@ -794,13 +794,28 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
...
             raise HaltUpgradePlanGeneration(
                 f"Application '{self.name}' already configured for release equal to or greater "
                 f"than {target}. Ignoring."
             )
 
+    def _need_origin_setting_update(self, target: OpenStackRelease) -> bool:
+        """Check if application needs to update origin setting for a certain target.
+
+        Applications with configuration equal or bigger than the target or without origin setting,
+        does not need to change origin setting.
```
rgildein, line 808: If this return False, we raised `HaltUpgradePlanGeneration`. My question is how operator will know that it's because of this check? e.g. self.apt_source_codename is None ?

cou/apps/base.py
gabrielcocenza: I think I see your point and I like the idea. However this is a configuration on the application level and not into the unit. I think a safe path here would be considering the min version on the units as the `apt_source_codename`

cou/apps/base.py
gabrielcocenza: It's not because of this check, but because of the context. If the application is on an OpenStack release equal or bigger than the target, the channel doesn't have an available update and the origin setting is already as expected, there is nothing to do.

In the case of apps without origin settings we are always returning False, meaning that there is nothing to do with this configuration.

rgildein: But if these return `False` the `HaltUpgradePlanGeneration` will be raised with message saying that application is already configured release equal or greater than target. But if application simply do not have that configuration option we could provide another message or maybe even do another steps for upgrading it?

rgildein (dismissed):

cou/apps/base.py
aieri: ftr I was using "unit" as a substitute for "machine, VM, or LXD", meaning simply "an Ubuntu installation"

cou/apps/base.py
```diff
@@ -132,42 +132,44 @@ def __str__(self) -> str:
...
+        """Extract the OpenStack release from Ubuntu Cloud Archive (UCA) sources.
+
+        :raises ApplicationError: When origin setting is not valid.
+        :return: OpenStackRelease object
+        :rtype: OpenStackRelease
+        """
+        # Ex: "cloud:focal-ussuri" will result in "ussuri"
+        *_, os_origin_parsed = self.os_origin.rsplit("-", maxsplit=1)
+        try:
+            return OpenStackRelease(os_origin_parsed)
```
agileshaw, line 168: If `openstack-origin` value should always follow the same format as `"cloud:focal-ussuri"`, we can move the parsing into the try block, plus removing `*`, to catch value error early (`os_origin` doesn't contain `-`)
```suggestion
        try:
            # Ex: "cloud:focal-ussuri" will result in "ussuri"
            _, os_origin_parsed = self.os_origin.rsplit("-", maxsplit=1)
            return OpenStackRelease(os_origin_parsed)
```

cou/apps/base.py
```diff
@@ -132,42 +132,44 @@ def __str__(self) -> str:
...
+
+        if self.os_origin == "distro":
             # find the OpenStack release based on ubuntu series
-            os_origin_parsed = DISTRO_TO_OPENSTACK_MAPPING[self.series]
-            return OpenStackRelease(os_origin_parsed)
+            return OpenStackRelease(DISTRO_TO_OPENSTACK_MAPPING[self.series])
 
-        elif self.os_origin == "":
-            return None
+        # probably because user set a ppa or an url
```
agileshaw, line 153:
```suggestion
        # probably because user set a ppa or a url
```

agileshaw (changes requested):

aieri (commented):
LGTM, I'll just wait for the fixes Tianqi requested before approving

agileshaw (approved):
LGTM, thanks

aieri (approved):

Issue #362, opened 2024-04-05
ovn-chassis is not executing sanity_checks


gabrielcocenza:
For some reason this executes the `enable-version-pinning` sanity check:

```python
def upgrade_plan_sanity_checks(
        self, target: OpenStackRelease, units: Optional[list[Unit]]
    ) -> None:
        ....
        self._check_version_pinning()
        self._check_ovn_support()
        super().upgrade_plan_sanity_checks(target, units)
```

But this does not execute:
```python
def upgrade_plan_sanity_checks(
        self, target: OpenStackRelease, units: Optional[list[Unit]]
    ) -> None:
        ....
        super().upgrade_plan_sanity_checks(target, units)
        self._check_version_pinning()
        self._check_ovn_support()
```

PR #367, opened 2024-04-09
Ensure check version pinning execution for OVN charms

gabrielcocenza:
- if sanity check called from parent class is called first, it might raise HaltUpgradePlanGeneration and not run the version pinning check
- remove origin setting from auxiliary subordinates unit tests

Closes: #362


agileshaw (approved):
LGTM

aieri (approved):

rgildein (approved):

PR #368, opened 2024-04-10
Add log with OS app string representation to Analysis._populate

rgildein:
With this, we can collect info about application during Analysis._populate, like this the log will contain the app string representation before planning.

log message example:
```bash
found OpenStack application:
app-name:
  model_name: <model-name>
  ...
```

PR #371, opened 2024-04-10
Add log with OS app string representation to Analysis._populate

gabrielcocenza:
With this, we can collect info about application during Analysis._populate, like this the log will contain the app string representation before planning.

log message example:
```
found OpenStack application:
app-name:
  model_name: <model-name>
  ...
```

This PR supersedes #368


aieri (approved):

agileshaw (approved):

rgildein (approved):
Thanks

PR #372, opened 2024-04-10
Fix support for latest/stable for channel based apps

gabrielcocenza:
- removed channel based and subordinates from calculating OpenStack release when not using releases channel


tests/unit/steps/test_analyze.py
```diff
@@ -536,3 +537,221 @@ def test_split_apps(exp_control_plane, exp_data_plane):
...
+    assert rmq_latest_stable.channel_codename == "ussuri"
+
+    assert cs_subordinate.current_os_release == "ussuri"
+    assert cs_subordinate.channel_codename == "ussuri"
+
+    assert channel_app_latest_stable.current_os_release == "ussuri"
+    assert channel_app_latest_stable.channel_codename == "ussuri"
+
+    # channel_app_latest_stable is skipped because is using latest/stable
+    # cs_subordinate is disconsidered because is from charmstore
```
agileshaw, line 744:
```suggestion
    # cs_subordinate is not considered because it's from charmstore
```

tests/unit/steps/test_analyze.py
```diff
@@ -536,3 +537,221 @@ def test_split_apps(exp_control_plane, exp_data_plane):
...
+    assert rmq_latest_stable.current_os_release == "yoga"
+    assert rmq_latest_stable.channel_codename == "ussuri"
+
+    assert cs_subordinate.current_os_release == "ussuri"
+    assert cs_subordinate.channel_codename == "ussuri"
+
+    assert channel_app_latest_stable.current_os_release == "ussuri"
+    assert channel_app_latest_stable.channel_codename == "ussuri"
+
+    # channel_app_latest_stable is skipped because is using latest/stable
```
agileshaw, line 743:
```suggestion
    # channel_app_latest_stable is skipped because it's using latest/stable
```

tests/unit/steps/test_analyze.py
```diff
@@ -536,3 +537,221 @@ def test_split_apps(exp_control_plane, exp_data_plane):
...
+    assert rmq.channel_codename == "yoga"
+
+    assert ch_subordinate_using_release.current_os_release == "victoria"
+    assert ch_subordinate_using_release.channel_codename == "victoria"
+
+    assert channel_app_using_release.current_os_release == "ussuri"
+    assert channel_app_using_release.channel_codename == "ussuri"
+
+    # channel_app_using_release and ch_subordinate_using_release are considered because are
+    # using release channel
```
agileshaw, line 637:
```suggestion
    # channel_app_using_release and ch_subordinate_using_release are considered because they are
    # using release channel
```

agileshaw (changes requested):
Changes look overall. Just some rewording + a non-blocking suggestions

cou/apps/base.py
```diff
@@ -244,6 +250,16 @@ def os_release_units(self) -> dict[OpenStackRelease, list[str]]:
...
 
+    @property
+    def using_release_channel(self) -> bool:
+        """Check if application is using release channel.
+
+        E.g of release channels: ussuri/stable, pacific/stable
+        :return: True if it is using, False otherwise
+        :rtype: bool
+        """
+        return not (self.is_from_charm_store or self.channel == LATEST_STABLE)
```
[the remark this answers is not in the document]
gabrielcocenza: I'm not a fan also. I was lacking of naming. Thanks for the suggestion

cou/apps/auxiliary.py
```diff
@@ -112,13 +112,9 @@ def channel_codename(self) -> OpenStackRelease:
...
         """
-        if self.is_from_charm_store:
+        if not self.using_release_channel:
             logger.debug(
-                (
-                    "'Application %s' installed from charm store; assuming Ussuri as the "
-                    "underlying version."
-                ),
-                self.name,
+                "%s cannot determine OpenStack release by channel. Assuming as Ussuri", self.name
```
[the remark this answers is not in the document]
aieri:
```suggestion
                "Cannot determine the OpenStack release of '%s' via its channel. Assuming Ussuri", self.name
```

cou/apps/base.py
```diff
@@ -538,24 +554,32 @@ def _get_charmhub_migration_step(self) -> PreUpgradeStep:
...
-    def _get_change_to_openstack_channels_step(self) -> PreUpgradeStep:
+    def _get_change_to_openstack_channels_step(self, target: OpenStackRelease) -> PreUpgradeStep:
         """Get the step for changing to OpenStack channels.
 
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
         :return: Step for changing to OpenStack channels
         :rtype: PreUpgradeStep
         """
+        expected_current_channel = (
```
rgildein, line 565: Should this if condition be part of the property? Or can we rename this variable?
Doing `expected_current_channel = (self.expected_current_channel if ... else ...`  do not look, right, what is expected channel then?

cou/utils/app_utils.py
```diff
@@ -135,13 +135,13 @@ async def _get_current_osd_release(unit: str, model: Model) -> str:
     return current_osd_release
 
 
-def stringify_units(units: Iterable[Unit]) -> str:
-    """Convert Units into a comma-separated string of unit names, sorted alphabetically.
+def stringify_classes(instances: Iterable[Any]) -> str:
```
rgildein, line 138:
```suggestion
def stringify_objects(objs: Iterable[Any]) -> str:
```

```suggestion
def stringify_instances(instances: Iterable[Any]) -> str:
```

cou/steps/analyze.py
```diff
@@ -156,12 +159,29 @@ def __str__(self) -> str:
...
         :rtype: Optional[OpenStackRelease]
         """
-        return min((app.current_os_release for app in apps), default=None)
+        # NOTE(gabrielcocenza) Apps based on channels to identify OpenStack release cannot
+        # be considered when using latest/stable or charmstore because it's not reliable and
+        # will be considered as Ussuri.
+        apps_skipped = {
+            app
+            for app in apps
+            if isinstance(app, (ChannelBasedApplication, SubordinateBase))
```
rgildein, line 176: To me this does not look right to check class here, we should instead bake this to classes instead.
Maybe as class argument?

cou/apps/auxiliary.py
```diff
@@ -63,13 +63,13 @@ def is_valid_track(self, charm_channel: str) -> bool:
             current_track,
         ) in TRACK_TO_OPENSTACK_MAPPING and len(possible_tracks) > 0
 
-    @property
-    def expected_current_channel(self) -> str:
+    def expected_current_channel(self, _target: OpenStackRelease) -> str:
```
rgildein, line 66: not-blocking: What is reason to use `_target` instead of `target`? Why it's not part of docstring?

cou/apps/channel_based.py
```diff
@@ -38,6 +38,24 @@ def get_latest_os_version(self, unit: Unit) -> OpenStackRelease:
         """
         return self.channel_codename
 
+    def expected_current_channel(self, target: OpenStackRelease) -> str:
```
rgildein, line 41: In base.py and auxiliary.py you used `_target` and here it's `target`. This needs to be unified.

cou/steps/analyze.py
rgildein: not-blocking: I think it would be better to add something like this to classes.
```python
class SubordianteBase(OpenstackApplication):

    based_on_channel = True
```

rgildein (changes requested):

cou/apps/auxiliary.py
rgildein: This actually good trick I did not know about :+1:

cou/apps/channel_based.py
rgildein: It's working, but it's strange :D. Now I understand how it works, so it's fine.

rgildein (dismissed):

tests/unit/apps/test_auxiliary_subordinate.py
aieri, line 1: I don't see a unit test for `expected_current_channel`

tests/unit/steps/test_analyze.py
```diff
@@ -536,3 +537,221 @@ def test_split_apps(exp_control_plane, exp_data_plane):
     control_plane, data_plane = Analysis._split_apps(all_apps)
     assert exp_control_plane == control_plane
     assert exp_data_plane == data_plane
+
+
+def test_min_os_release_apps_release_channel(model):
```
aieri, line 542: there are a lot of asserts for a single unit test, is it necessary to keep them all together?

tests/unit/steps/test_analyze.py
```diff
@@ -536,3 +537,221 @@ def test_split_apps(exp_control_plane, exp_data_plane):
...
+        )
+        == "ussuri"
+    )
+
+    assert (
+        Analysis.min_os_release_apps([keystone, rmq, ch_subordinate_using_release]) == "victoria"
+    )
+
+
+def test_min_os_release_apps_not_release_channel(model):
```
aieri, line 650: as above

aieri (changes requested):

cou/apps/base.py
```diff
@@ -244,6 +254,17 @@ def os_release_units(self) -> dict[OpenStackRelease, list[str]]:
...
+    @property
+    def using_non_release_channel(self) -> bool:
+        """Check if application is using non release channel.
+
+        E.g of non release channels: latest/stable
+        E.g of release channels: ussuri/stable
+        :return: True if it is using, False otherwise
+        :rtype: bool
+        """
+        return self.is_from_charm_store or self.channel == LATEST_STABLE
```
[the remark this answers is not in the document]
aieri: outcome of discussion: this function can probably me merged into `is_valid_track`

cou/apps/channel_based.py
```diff
@@ -38,6 +41,24 @@ def get_latest_os_version(self, unit: Unit) -> OpenStackRelease:
...
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :return: The expected current channel of the application. E.g: "ussuri/stable"
+        :rtype: str
+        """
+        return (
+            f"{target.previous_release}/stable"
+            if self.using_non_release_channel
+            else f"{self.current_os_release.codename}/stable"
+        )
```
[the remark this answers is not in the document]
aieri: result of conversation:
```
if workload_version exists:
    if workload_version maps to 1 o7k release:
        return the o7k release / stable
    if workload_version maps to n or 0 o7k releases:
        return target - 1
else:
    return target - 1
```

cou/utils/app_utils.py
```diff
@@ -135,13 +135,13 @@ async def _get_current_osd_release(unit: str, model: Model) -> str:
...
-    :param units: An iterable of Unit objects to be converted.
-    :type units: Iterable[Unit]
+    :param objs: An iterable of objects to be converted that has name attribute.
+    :type objs: Iterable[Any]
     :return: A comma-separated string of sorted unit names.
     :rtype: str
     """
-    sorted_unit_names = sorted([unit.name for unit in units])
-    return ", ".join(sorted_unit_names)
+    sorted_names = sorted([obj.name for obj in objs if hasattr(obj, "name")])
```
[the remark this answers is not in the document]
aieri: to make this work we'd also have to move the extended string representation of Application we currently have to `__repr__` and ensure we call `repr(Application)` when logging.

...or we drop the function altogether and print the Iterable directly

cou/apps/base.py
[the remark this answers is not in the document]
gabrielcocenza: In the end I think it make sense to have a separate property for this, but the name wasn't adequate. Changing to `need_crossgrade`  makes more sense and we wouldn't have the issue as you mentioned for auxiliary charms.

tests/unit/apps/test_channel_based.py
```diff
@@ -59,6 +59,54 @@ def test_application_versionless(model):
     assert app.get_latest_os_version(units["glance-simplestreams-sync/0"]) == app.channel_codename
 
 
+def test_channel_based_application_latest_stable(model):
```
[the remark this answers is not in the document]
gabrielcocenza: The test that calculates the overall cloud version is on the analyze test. This test is showing that the channel based applications are totally dependent on the target passed because the application might say that is ussuri, but it will crossgrade to the `N-1` release from the target and then change to the target channel

cou/utils/juju_utils.py
```diff
@@ -150,6 +150,14 @@ class Unit:
     machine: Machine
     workload_version: str
 
+    def __repr__(self) -> str:
```
aieri, line 153: why `__repr__` instead of `__str__`?

aieri (approved):
One question, but otherwise lgtm

cou/utils/juju_utils.py
gabrielcocenza: There is no specific reason. Generally `__repr__` is aimed at the programmer and `__str__` to the user. In this case I guess it's the same for both, so I think it's ok

rgildein (approved):

PR #379, opened 2024-04-18
Fix Hypervisor Upgrade

gabrielcocenza:
- HypervisorUpgradePlan was wrongly running steps on parallel which caused to change openstack-origin or source at the same time of action-managed-upgrade. This could wrongly trigger the upgrade on all units.
- added the Psi (Ψ) to user vizualize which steps are going to run on parallel
- updated docs


cou/steps/__init__.py
```diff
@@ -351,7 +359,7 @@ def __init__(self, description: str):
         :param description: Description of the step.
         :type description: str
         """
-        super().__init__(description=description, parallel=True, coro=None)
+        super().__init__(description=description, parallel=False, coro=None)
```
aieri, line 362: After saying "All sub-steps will run in parallel." seeing `parallel=False` is quite confusing. I think it's important to add a comment here explaining the behavior, as well as improving the docstring of `BaseStep`.

docs/explanation/planning.rst
```diff
@@ -69,5 +73,9 @@ successfully.
 The plan can also be obtained without the need to perform a cloud upgrade using
 the **plan** command. See :doc:`Plan an upgrade <../how-to/plan-upgrade>`.
 
+The Greek letter Psi `Ψ` means that the step will run in parallel compared with other steps in the
+same level. In the example above on the `data-plane hypervisors upgrade`, `unit 1` will run on
+parallel compared with `unit K`, but their `upgrade steps A to Z` will run sequentially.
```
aieri, line 78:
```suggestion
The Greek letter Psi `Ψ` indicates that the step and its sub-steps will run in parallel with other steps in the
same level. In the example above, within the `data-plane hypervisors upgrade`, the `unit 1` step will run in
parallel with the `unit K` step, but their `upgrade steps A to Z` sub-steps will run sequentially.
```

aieri (changes requested):
LGTM code-wise, just some documentation suggestions

cou/steps/__init__.py
```diff
@@ -131,11 +131,19 @@ def __str__(self) -> str:
...
+            step, indent, parallel = steps_to_visit.pop()
+            result += (
+                f"{tab * indent}{parallel_char if parallel else ''}{step.description}{os.linesep}"
+                if step
+                else ""
+            )
+            steps_to_visit.extend(
+                [(s, indent + 1, step.parallel) for s in reversed(step.sub_steps)]
+            )
+            # breakpoint()
```
rgildein, line 146: Can we remove this?

tests/unit/steps/test_steps.py
```diff
@@ -140,11 +140,11 @@ def test_step_str():
 def test_step_str_dependent():
     """Test BaseStep string representation."""
     expected = (
-        f"a\n\ta.a\n\t\t{DEPENDENCY_DESCRIPTION_PREFIX}a.a.a\n"
-        f"\t\t{DEPENDENCY_DESCRIPTION_PREFIX}a.a.b\n\ta.b\n"
+        f"a\n\ta.a\n\t\tΨ {DEPENDENCY_DESCRIPTION_PREFIX}a.a.a\n"
```
rgildein, line 143: The test is saying `test_step_str_dependent`, can we do this test in another test or renamed it.

rgildein (changes requested):

tests/unit/steps/test_steps.py
gabrielcocenza: We already have `test_plan_structure` doing this. I'll update the test docstring there and I think we are covered.

cou/steps/__init__.py
```diff
@@ -78,7 +78,9 @@ def __init__(
 
         :param description: Description of the step.
         :type description: str
-        :param parallel: Define if step should run on parallel or not.
+        :param parallel: Define if the sub-steps should run on parallel or not. Note that the step
+        itself will not run on parallel, just the sub-steps will. Each sub-step is also responsible
```
aieri, line 82: nit
```suggestion
        :param parallel: Define if the sub-steps should be run in parallel or not. Note that this BaseStep instance will not be affected by this parameter, only its sub-steps will. Each sub-step is also responsible
```

cou/steps/__init__.py
```diff
@@ -337,10 +346,9 @@ class ApplicationUpgradePlan(UpgradePlan):
 class HypervisorUpgradePlan(BaseStep):
     """Represents the plan for hypervisor upgrade.
 
-    This class is intended to be used as a group for unit-level upgrade steps, which are
-    grouped by a single hypervisor. It doesn't accept coroutine or parallel as inputs.
-
-    All sub-steps will run in parallel.
+    This class is intended to be used as a group by AZs. It doesn't accept coroutine or parallel
+    arguments as inputs. AZ groups will run sequentially and the units within the AZ will run on
```
aieri, line 350: nit
```suggestion
    arguments as inputs. AZ groups will run sequentially and the units within the AZ will run in
```

aieri (approved):
just non-blocking suggestions at this point

rgildein (approved):

aieri (approved):

PR #385, opened 2024-04-22
Run update-status hook before checking workload version upgrade

### LINKED ISSUE #384: Give a retry to post-upgrade to check workload_version change
The post-upgrade  to check if an application has upgraded its workload version sometimes fails and after some time the workload_version on charm update.

gabrielcocenza:
- Sometimes the _get_reached_expected_target_step post upgrade step fails because the application turn into active/idle and might take an extra time to actually show the change on the workload version. Running the update-status hook before checking can help reducing false positives.
- Changed the error message to say to the user wait some minutes if the workload version change check fails.

Closes #384

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -15,6 +15,7 @@
 """Base application class."""
 from __future__ import annotations
 
+import asyncio
 import logging
 import os
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


cou/apps/base.py
```diff
@@ -363,7 +369,8 @@ async def _verify_workload_upgrade(self, target: OpenStackRelease, units: list[U
 
         if units_not_upgraded:
             raise ApplicationError(
-                f"Cannot upgrade units '{', '.join(units_not_upgraded)}' to {target}."
+                f"Cannot upgrade unit(s) '{', '.join(units_not_upgraded)}' to {target}. "
+                "Try again in some minutes to see if the unit(s) have been upgraded."
```
aieri, line 373: nit: technically if we have reached this line the upgrade process itself did not fail, so it's not us who were unable to upgrade the units.
```suggestion
                f"Unit(s) '{', '.join(units_not_upgraded)}' did not complete the upgrade to {target}. "
                "Some local processes may still be executing; you may try re-running COU in a few minutes."
```

aieri (approved):
Looks good, I'd just tweak the wording a bit

rgildein (approved):

rgildein (approved):

aieri (approved):

PR #394, opened 2024-04-29
Add execution time of Applications and Hypervisors AZs

gabrielcocenza:
With this change it will be possible to show to the users how long COU took to execute an upgrade. This will make easier to calculate the expected time to upgrade control-plane and data-plane.

With this change the user will see the result of applications upgrades like this:

```shell
Running cloud upgrade...
Verify that all OpenStack applications are in idle state ✔
Upgrade plan for 'ceilometer' to 'victoria' executed in 169 seconds ✔
Upgrade plan for 'ceph-radosgw' to 'victoria' executed in 118 seconds ✔
Upgrade plan for 'cinder' to 'victoria' executed in 150 seconds ✔
....
Upgrade plan for [cinder-volume/0, nova-compute-kvm/0] in 'AZ3' to 'victoria' executed in 453 seconds ✔
```

### DIFF
--- a/cou/steps/execute.py
+++ b/cou/steps/execute.py
@@ -12,15 +12,21 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 """Execution logic."""
-
 import asyncio
 import logging
 import sys
+import time
 
 from cou.exceptions import HaltUpgradeExecution, RunUpgradeError
-from cou.steps import ApplicationUpgradePlan, BaseStep, UpgradeStep
+from cou.steps import (
+    ApplicationUpgradePlan,
+    BaseStep,
+    HypervisorUpgradePlan,
+    UpgradeStep,
+)
 from cou.utils import print_and_debug, progress_indicator, prompt_input
 
+GROUP_STEPS = (ApplicationUpgradePlan, HypervisorUpgradePlan)
 AVAILABLE_OPTIONS = ["y", "yes", "n", "no"]
 
 logger = logging.getLogger(__name__)
@@ -107,6 +113,9 @@ async def _run_step(step: BaseStep, prompt: bool, overwrite_progress: bool = Fal
                                in CLI output. True to overwrite and False (the default) to persist.
     :type overwrite_progress: bool
     """
+    if isinstance(step, GROUP_STEPS):
+        start_time = time.time()
+
     if isinstance(step, UpgradeStep):
         progress_indicator.start(step.description)
         await step.run()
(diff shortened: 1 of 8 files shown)


aieri (commented):
time-tracking aside, I don't understand the implication of extending these codepaths to apply to HypervisorUpgradePlan (in addition to ApplicationUpgradePlan). Should we have always done so?

rgildein (dismissed):
LGTM, nice and simple solution :+1:

rgildein:
> time-tracking aside, I don't understand the implication of extending these codepaths to apply to HypervisorUpgradePlan (in addition to ApplicationUpgradePlan). Should we have always done so?

I think we just forgot to do so, we could use only UpgradePlan, where both Application and Hypervisor are from UpgradePlan. I think current approach is ok.

gabrielcocenza:
> time-tracking aside, I don't understand the implication of extending these codepaths to apply to HypervisorUpgradePlan (in addition to ApplicationUpgradePlan). Should we have always done so?

The implications are that before this change all steps in the HypervisorUpgradePlan would not overwrite and we would have something like this:

```shell
    Upgrade plan for 'AZ3' to 'victoria' executed in 453 seconds 
	Disable nova-compute scheduler from unit: 'nova-compute-kvm-sriov/1' ✔
	Upgrade software packages of 'nova-compute-kvm-sriov' from the current APT repositories ✔
		Ψ Upgrade software packages on unit 'nova-compute-kvm-sriov/1'✔
	Upgrade 'nova-compute-kvm-sriov' from 'victoria/stable' to the new channel: 'wallaby/stable' ✔
	Change charm config of 'nova-compute-kvm-sriov' 'openstack-origin' to 'cloud:focal-wallaby' ✔
	Upgrade plan for units: nova-compute-kvm-sriov/1 ✔
		Ψ Upgrade plan for unit 'nova-compute-kvm-sriov/1' ✔
			Verify that unit 'nova-compute-kvm-sriov/1' has no VMs running ✔
			├── Pause the unit: 'nova-compute-kvm-sriov/1' ✔
			├── Upgrade the unit: 'nova-compute-kvm-sriov/1' ✔
			├── Resume the unit: 'nova-compute-kvm-sriov/1' ✔
```

Whereas now we'll see something like:

```
Upgrade plan for 'AZ3' to 'victoria' executed in 453 seconds ✔
```

IMO it's one thing that silent passed and I guess it should be done before. @rgildein right now `HypervisorUpgradePlan` is inheriting from `BaseStep` and `HypervisorGroupUpgradePlan` inherits from `UpgradePlan`, but it's not being used. I guess in another PR we can remove `HypervisorGroupUpgradePlan` and inherit `HypervisorUpgradePlan` from `UpgradePlan`. What do you say?

aieri (approved):

rgildein (approved):
LGTM and I agree with your comment @gabrielcocenza. Thanks

Issue #393, opened 2024-04-29
Error on the auxiliary function _verify_nova_compute


gabrielcocenza:
When running `_verify_nova_compute` the following error appears:

```
2024-04-29 15:00:07 [cou.steps.execute] [INFO] Running: Verify that all 'nova-compute' units has been upgraded
2024-04-29 15:00:07 [cou.steps] [DEBUG] running step: PreUpgradeStep(Verify that all 'nova-compute' units has been upgraded)
2024-04-29 15:00:15 [cou.cli] [ERROR] Unexpected error occurred.
2024-04-29 15:00:15 [cou.cli] [ERROR] 'str' object has no attribute 'charm'
Traceback (most recent call last):
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/cli.py", line 217, in entrypoint
    loop.run_until_complete(_run_command(args))
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/cli.py", line 203, in _run_command
    await run_upgrade(args)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/cli.py", line 189, in run_upgrade
    await apply_step(upgrade_plan, args.prompt)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/steps/execute.py", line 178, in apply_step
    await _run_step(step, prompt, overwrite_progress)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/steps/execute.py", line 134, in _run_step
    await _run_sub_steps_sequentially(step, prompt, overwrite_substeps_progress)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/steps/execute.py", line 99, in _run_sub_steps_sequentially
    await apply_step(sub_step, prompt, overwrite_progress)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/steps/execute.py", line 178, in apply_step
    await _run_step(step, prompt, overwrite_progress)
  File "/home/ubuntu/cocenza/charmed-openstack-upgrader-1/cou/steps/execute.py", line 134, in _run_step
    await _run_sub_steps_sequentially(step, prompt, overwrite_substeps_progress)
[...]

PR #397, opened 2024-05-01
Fix _verify_nova_compute

gabrielcocenza:
- logic was getting the app as a string instead of the object
- fix unit tests

Closes: #393

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -351,7 +351,7 @@ class CephOsd(AuxiliaryApplication):
         units_not_upgraded = []
         apps = await self.model.get_applications()
 
-        for app in apps:
+        for app in apps.values():
             if app.charm != "nova-compute":
                 logger.debug("skipping application %s", app.name)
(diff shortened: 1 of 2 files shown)


aieri (approved):

cou/apps/auxiliary.py
```diff
@@ -351,7 +351,7 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
         units_not_upgraded = []
         apps = await self.model.get_applications()
 
-        for app in apps:
+        for app in apps.values():
```
rgildein, line 354: Why we are not checking only the app itself? Why all nova-compute's in model.

rgildein (commented):
LGTM, only one comment.

cou/apps/auxiliary.py
rgildein: Ach yes, this is in caph-osd. I thought it is in nova-compute app. Sorry and thanks for replay.

Issue #395, opened 2024-05-01
Ceph-osd gives mismatched version when colocated with nova-compute


gabrielcocenza:
In a scenario where some ceph-osd are colocated with nova-compute units, when ceph version is upgraded in an OpenStack transition like in Victoria -> Wallaby, upgrading a nova-compute unit also upgrades the ceph-osd unit. When running `cou plan` this error show up:

```
2024-05-01 14:13:56 [WARNING] Cannot generate plan for 'ceph-osd-hdd'
	Units of application ceph-osd-hdd are running mismatched OpenStack versions: 'victoria': ['ceph-osd-hdd/0', 'ceph-osd-hdd/2'], 'xena': ['ceph-osd-hdd/1']. This is not currently handled.
2024-05-01 14:13:56 [WARNING] Running upgrades will not be possible until problems indicated in the warnings are resolved.
```

This does not happen if using `cou plan hypervisors` because ceph-osd is disconsidered

PR #248, opened 2024-02-16
Add tuple of apps to COUMachine

rgildein:
Like this, we can identify all apps deployed on the machine. This can be used in #247 to reduce the amount of for loops.

PR #398, opened 2024-05-01
Don't check for mismatched versions if colocated with nova-compute

gabrielcocenza:
- nova-compute is expected to upgrade unit by unit. So it's expected to have different versions among nova-compute units or apps colocated with it.
- _check_mismatched_versions now check if a machine of an app is colocated with nova-compute.

Closes: #395

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -857,7 +857,10 @@ class OpenStackApplication(Application):
         :raises MismatchedOpenStackVersions: When the units of the app are running
                                              different OpenStack versions.
         """
-        if units:
+        # NOTE (gabrielcocenza) nova-compute is upgraded using paused-single-unit,
+        # so it's possible to have mismatched version in applications units that are
+        # nova-compute or colocated with it.
+        if any("nova-compute" in machine.apps for machine in self.machines.values()):
             return
 
         os_versions = self.os_release_units
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -125,7 +125,7 @@ class OpenStackApplication(Application):
                 "machines": {
                     machine.machine_id: {
                         "id": machine.machine_id,
-                        "apps": machine.apps,
+                        "apps_charms": machine.apps_charms,
                         "az": machine.az,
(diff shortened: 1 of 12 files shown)


cou/apps/base.py
```diff
@@ -857,7 +857,10 @@ def _check_mismatched_versions(self, units: Optional[list[Unit]]) -> None:
         :raises MismatchedOpenStackVersions: When the units of the app are running
                                              different OpenStack versions.
         """
-        if units:
+        # NOTE (gabrielcocenza) nova-compute is upgraded using paused-single-unit,
+        # so it's possible to have mismatched version in applications units that are
+        # nova-compute or colocated with it.
+        if any("nova-compute" in machine.apps for machine in self.machines.values()):
```
aieri, line 863: is `machine.apps` a list of charm names (instead of a list of application classes or a list of application names)?

gabrielcocenza: I thought it was charm name, but It's actually a tuple of application names that are in a specific machine. See:
https://github.com/canonical/charmed-openstack-upgrader/blob/1d25afc8306595eb549438ed39ce4269df6797d4/cou/utils/juju_utils.py#L283-L296

I'll need to change the logic a little bit. Thanks

rgildein: I do not think that this is good approach, we want to avoid any libjuju object passing to our code, so Machine should not include juju.application.Application object.

rgildein: Och wait the `unit.application` in only name here not an Application object?

gabrielcocenza: It's just the name of the application and not a juju object. I guess I'll need to include the change as we almost done [before](https://github.com/canonical/charmed-openstack-upgrader/pull/248/commits/87189c184e0fa4185a89971e671d9adf263db335#r1497891237)

rgildein: You can get applications and find the charm name there.

aieri (approved):

rgildein (approved):

Issue #399, opened 2024-05-02
COU is not generating the PostUpgrade step to ensure that the require-osd-release option matches the ceph-osd version


gabrielcocenza:
COU it's not generating because it was passing data-plane apps to find ceph-mon and ceph-mon is normally a control-plane app

Issue #396, opened 2024-05-01
ceph-osd should run upgrade steps when all nova-compute units colocated are upgraded


gabrielcocenza:
In deployments that nova-compute is colocated with ceoh, upgrading all nova-compute units also upgrades the ceph packages. With that COU doesn't run the necessary steps to properly upgrade ceph, because it sees that the workload version is already upgraded.

Issue #380, opened 2024-04-19
The plan for ceph-osd disappears after upgrading hypervisors


gabrielcocenza:
Auxiliary charms, like ceph-osd, when called to give the current OpenStack release will always answer with the highest compatible one. That means if the workload_version is running `octopus`, it's compatible with `Ussuri` and `Victoria` and the application will say that is `Victoria`. If all OpenStack components are already upgraded to `Victoria`, ceph-osd will stop generating a plan to change the `source` to `cloud:focal-victoria` and will start generating the plan for `Wallaby` changing the `source` from `distro` directly to `cloud:focal-wallaby`. Although there is no change in the package version between ussuri to victoria, this is not consistent and can confuse users because they would see a ceph plan using `cou plan` and wouldn't see after upgrading all hypervisors

chanchiwai-ray:
@gabrielcocenza this was mentioned in and fixed by PR: #400, can you confirm that this issue is still valid or not

PR #400, opened 2024-05-02
current_os_release also consider OpenStack release set in configuration

gabrielcocenza:
Apps that are colocated with nova-compute will upgrade packages during nova-compute upgrade. To ensure to change the 'source' or 'openstack-origin' and run all the upgrade steps necessary, it's necessary to include the OpenStack release set in the application configuration.
    
This patch also fix the post-upgrade generation to ensure that the require-osd-release option matches the ceph-osd version by passing control-plane apps where ceph-mon is and not on data-plane apps
    
Closes: #399, #396 and #380


cou/apps/base.py
```diff
@@ -160,6 +160,8 @@ def apt_source_codename(self) -> OpenStackRelease:
 
         if self.os_origin == "distro":
             # find the OpenStack release based on ubuntu series
+            if self.series not in DISTRO_TO_OPENSTACK_MAPPING:
+                raise ApplicationError(f"Series '{self.series}' is not supported by COU.")
```
aieri, line 164: non-blocking nit: I wonder if it's good behavior having a property that could throw an exception if called. At the end of the day if the app class does not support a certain series we should probably throw an error earlier on in the constructor. And as a sidenote (out of scope here), I also wonder if it would have been a better design to push the distro and uca to release mapping within the OpenStackRelease class, to allow things like
```
OpenStackRelease('focal') # Ussuri
OpenStackRelease('cloud:focal-yoga') # Yoga
```

rgildein: I agree, it looks more like work for sanity check instead of property.

cou/apps/base.py
```diff
@@ -205,9 +207,16 @@ def channel_codename(self) -> OpenStackRelease:
     def current_os_release(self) -> OpenStackRelease:
         """Current OpenStack Release of the application.
 
+        Applications that are colocated in a same machine will upgrade all the packages in the
```
rgildein, line 210: Would be possible to do something like this in separate class, because like this we are adding more and more complexity to the base, which can end up to complex <==> easy to break.

cou/apps/base.py
```diff
@@ -205,9 +207,16 @@ def channel_codename(self) -> OpenStackRelease:
...
+        machine during an upgrade. This means that when upgrading a application X, the application
+        Y colocated with it also upgrades its packages. To ensure to change the 'source' or
+        'openstack-origin' and run all the upgrade steps necessary, it's necessary to include the
+        OpenStack release set in the application configuration.
         :return: OpenStackRelease object
         :rtype: OpenStackRelease
         """
+        if self.os_origin:
+            return min(list(self.os_release_units.keys()) + [self.apt_source_codename])
         return min(self.os_release_units.keys())
```
aieri, line 220: This seems fragile... this if seems to only exist to avoid an infinite loop with line 156. In other words the implementation of `apt_source_codename` leaks into `current_os_release` and vice versa. I suggest changing `apt_source_codename` to treat `os_origin` being empty or absent or set to `distro` in the same way, and dropping the if here. Also note https://github.com/juju/charm-helpers/blob/807f705d1d132bedb62b2eb743403e1d5867360f/charmhelpers/fetch/ubuntu.py#L654

cou/steps/plan.py
aieri, line 1: this is ok but doesn't it belong to a separate PR?

aieri (changes requested):

cou/apps/base.py
gabrielcocenza: This also can affect services like `cinder`  so I think it should be in the base

cou/steps/plan.py
gabrielcocenza: I think it does, but at the time I need one code base with both fixes for running on partner cloud. Should I create a new PR for this? It's a detail and I think there is no harm patching together, but I agree that it's not ideal.

aieri: no it's ok for now, my comment was more about keeping things separate next time

cou/apps/base.py
```diff
@@ -848,7 +845,8 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
         if (
             self.current_os_release >= target
             and not self.can_upgrade_to
-            and self.apt_source_codename >= target
+            # consider apt_source_codename just when exist or not empty
+            and (self.apt_source_codename >= target if self.os_origin else True)
```
aieri, line 849: I don't understand this bit: if `os_origin` is empty, `apt_source_codename` will follow the `distro` codepath and return some OpenStack release codename. Can you give me an example of a situation that makes this change necessary?

gabrielcocenza: Sure...
Imagine an app like Vault that does not have `source` or `openstack-origin` or a subordinate-charm. When asked the `apt_source_codename` this will return `Ussuri` that will never be bigger than the target and won't halt the upgrade plan generation.

To not happen that, I'm considering just when there is something is set to `os_origin`

aieri: @gabrielcocenza ok... but then you're leaking details pertaining to a different class into the base. In fact, the implementation of this function in `subordinate.py` is different.
...this shows the limitations of our class structure. In theory you could use the same approach and override this function in AuxiliaryApplication, except that while vault does not have a source setting, ceph-fs and ceph-radosgw do...

At this point perhaps - and not that I particularly like this - remove the override in `SubordinateApplication._check_application_target()` and handle all variations directly in base (you'll have to handle two different types of log lines though)

@samuelallan72 we have an ordered list of release codenames, the comparison is not purely alphabetical

cou/apps/base.py
rgildein: I think it would be better in some another class used as mixin or as base for application which require this. However, I'll not block this to be merged if others agree that it's not blocking.

cou/apps/base.py
rgildein: I start thinking, that it is time to start using mixin class. We are doing to many hack's in base already.

gabrielcocenza: Good catch. I've removed the function from the subordinate one because now the base can deal for all class types. Thanks

aieri (dismissed):

cou/apps/base.py
aieri: I agree it would be an important and desirable change, but not a blocking one. I would propose merging this and refactoring at a later date

rgildein (approved):

rgildein (approved):
Do not like it, but so we can do release.

aieri (approved):

aieri:
As in previous comments, we are piling more and more complexity in the base class... A refactor would be really useful here

PR #406, opened 2024-05-07
Check overall ceph versions

### LINKED ISSUE #401: ceph-mon is failing to show the right version of osd and this can be problematic
For some unknown reason, after upgrading all ceph-osds to `pacific` on partner cloud, ceph-mon still says that osd's are on `octopus`. 

E.g:
```
{
    "mon": {
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 3
    },
    "mgr": {
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 3
    },
    "osd": {
        "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)": 18
    },
    "mds": {},
    "rgw": {
        "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)": 3
    },
    "overall": {
        "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)": 21,
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 6
    }
}
```

This is problematic because it pass the `set_require_osd_release_option` once that the ` require-osd-release` is also on `octopus`and this will let the cloud continue upgrading. I guess the right thing is to halt the execution if we see that there is more than one ceph version in the cloud.

gabrielcocenza:
Check if there is more than one ceph version in the overall field

- when more than one overall field shows, this probably means that is necessary to restart ceph services or manually upgrade a ceph charm

Closes #401

### DIFF
--- a/cou/utils/app_utils.py
+++ b/cou/utils/app_utils.py
@@ -44,12 +44,21 @@ async def upgrade_packages(unit: str, model: Model, packages_to_hold: Optional[l
     await model.run_on_unit(unit_name=unit, command=command, timeout=600)
 
 
-async def set_require_osd_release_option(unit: str, model: Model) -> None:
+async def set_require_osd_release_option(
+    expected_ceph_release: str, unit: str, model: Model
+) -> None:
     """Check and set the correct value for require-osd-release on a ceph-mon unit.
 
-    This function compares the value of require-osd-release option with the current release
-    of OSDs. If they are not the same, set the OSDs release as the value for
+    This function compares the value of require-osd-release option with the expected release
+    ceph release. If they are not the same, set the OSDs release as the value for
     require-osd-release.
+    As a pre-upgrade step when upgrading ceph-mon, the expected release should match with the
+    expected current channel because ceph components are not upgraded yet, whereas as a post
+    upgrade of ceph-osd, it should match with the target because all ceph components should be
+    already upgraded.
+
+    :param expected_ceph_release: The expected ceph release of the cloud.
+    :type expected_ceph_release: str
     :param unit: The ceph-mon unit name where the check command runs on.
     :type unit: str
     :param model: Model object
@@ -58,15 +67,14 @@ async def set_require_osd_release_option(unit: str, model: Model) -> None:
     """
     # The current `require_osd_release` value set on the ceph-mon unit
     current_require_osd_release = await _get_required_osd_release(unit, model)
-    # The actual release which OSDs are on
-    current_running_osd_release = await _get_current_osd_release(unit, model)
 
-    if current_require_osd_release != current_running_osd_release:
-        set_command = f"ceph osd require-osd-release {current_running_osd_release}"
+    await _validate_ceph_component_versions(unit, model)
+
+    if current_require_osd_release != expected_ceph_release:
+        set_command = f"ceph osd require-osd-release {expected_ceph_release}"
         await model.run_on_unit(unit_name=unit, command=set_command, timeout=600)
 
 
-# Private functions
 async def _get_required_osd_release(unit: str, model: Model) -> str:
     """Get the value of require-osd-release option on a ceph-mon unit.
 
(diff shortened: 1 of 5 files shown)


gabrielcocenza:
After this patch, a ceph-mon with more than one versions on `overall` will fail.

E.g:
```shell
ubuntu@juju-204e4c-cou-0:~$ sudo ceph versions
{
    "mon": {
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 3
    },
    "mgr": {
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 3
    },
    "osd": {
        "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)": 3
    },
    "mds": {},
    "overall": {
        "ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)": 3,
        "ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)": 6
    }
}
```

Running COU will fail with the following message:
```
Would you like to start the upgrade? Continue (y/N): y
Running cloud upgrade...
Ensure that the 'require-osd-release' option in 'ceph-mon' matches the 'ceph-osd' version ✖
2024-05-07 18:32:14 [ERROR] Ceph applications are on mismatched releases:
{'ceph version 15.2.17 (8a82819d84cf884bd39c17e3236e0632ac146dc4) octopus (stable)': 3, 'ceph version 16.2.14 (238ba602515df21ea7ffc75c88db29f9e5ef12c9) pacific (stable)': 6}.Please manually upgrade or restart ceph services to have a single version.
2024-05-07 18:32:14 [ERROR] See the known issues at https://canonical-charmed-openstack-upgrader.readthedocs-hosted.com/en/stable/reference/known-issues/

```

cou/utils/app_utils.py
```diff
@@ -110,16 +110,28 @@ async def _get_current_osd_release(unit: str, model: Model) -> str:
...
+        raise RunUpgradeError("Cannot get release information on ceph applications.")
+
     # throw exception if OSDs are on mismatched releases
     if len(osd_release_output) > 1:
         raise RunUpgradeError(
             f"OSDs are on mismatched releases:\n{osd_release_output}."
             "Please manually upgrade them to be on the same release before proceeding."
         )
 
+    if len(overall_output) > 1:
```
aieri, line 129: wouldn't you normally have at least 2 versions while doing an upgrade?

aieri: although this could be ok as a pre-check before updating the min osd release

aieri (changes requested):
Overall, I think this would work, but I don't think the separation of duties is correct here.

You have a `_get_current_osd_release` function that raises error messages asking the user to upgrade components, which is a leakage of concerns from the caller to the callee.

I think a better approach would be to overhaul the `set_require_osd_release_option`, `_get_required_osd_release`, and `_get_current_osd_release` functions. At the end of the day, the idea here is to upgrade to a target and then prevent older OSDs from joining the cluster, right?

So I would very simply have a `set_require_osd_release_option(target)`, skip `_get_required_osd_release` altogether, and raise an exception if any version older than `target` is detected (because bumping the minimum version would be disruptive).

Does it make sense?

gabrielcocenza:
I didn't have time to finish this PR before my vacation. The idea of this last commit is to validate the ceph-osd configuration before upgrading ceph-mon which is the first ceph component for a target (all ceph units should have the same track as the expected for the current workload version) and after upgrading ceph-osd which is the latest ceph component (all ceph components should have the same track as the target)

Pjack:
It's on the tracking board for a while 
If we do not have time to finish this one, do you mind to close it? 
We can reopen it when we are available again.

gabrielcocenza:
Closing the PR. This should be handled later

PR #410, opened 2024-05-13
Add support for Juju 3.4

rgildein:
Fix getting results from `run_on_unit` and `run_action`. Also, series was replaced with base in juju status, so the small function to convert base to series was required.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Publish snap
 
 on:
   push:
-    branches: [ main, '*.x' ]
+    branches: [ main ]
   release:
     types: [ published ]
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Publish snap
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, '*.x' ]
   release:
     types: [ published ]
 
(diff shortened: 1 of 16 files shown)


jneo8 (approved):

chanchiwai-ray (approved):

.github/workflows/release.yaml
chanchiwai-ray, line 1: Can you remove this change from this PR since you are already doing that in https://github.com/canonical/charmed-openstack-upgrader/pull/411

chanchiwai-ray (changes requested):
Can you make this change independent of snap release

.github/workflows/release.yaml
rgildein: No, because if this is merged it will be released to the edge and when we merge #411 first there is no need to remove anything.

chanchiwai-ray: but if you remove `*.x` in the release.yaml, it won't release to edge

jneo8 (changes requested):
Need to fix lint.

rgildein:
> Need to fix lint.

It is not related with this PR.

jneo8 (approved):
LGTM.

Wait for #417 merged.

jneo8:
Hi @rgildein, can you update your PR? We just fix the linter issue.

chanchiwai-ray (approved):

PR #411, opened 2024-05-13
use separate track for 3.x

rgildein:
Change release workflow to release on push to 3.x branch and push it to the 3.x/edge instead of latest/edge.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Publish snap
 
 on:
   push:
-    branches: [ '*.x' ]
+    branches: [ '3.x' ]
   release:
     types: [ published ]
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Publish snap
 
 on:
   push:
-    branches: [ main ]
+    branches: [ '*.x' ]
   release:
     types: [ published ]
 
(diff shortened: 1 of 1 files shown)


jneo8 (changes requested):
Can you explain more about the `3.x` track?

Maybe add some description in README.

.github/workflows/release.yaml
```diff
@@ -2,7 +2,7 @@ name: Publish snap
 
 on:
   push:
-    branches: [ main ]
+    branches: [ '*.x' ]
```
chanchiwai-ray, line 5: Why use `*.x` instead of more explicit `3.x` ?

rgildein: Good point, we can use `3.x` directly.

rgildein:
> Can you explain more about the `3.x` track?
> 
> Maybe add some description in README.

What do you want to explain?

jneo8:
> What do you want to explain?

When to use 3.x and when to use master. I think we should have that on the README.md.

rgildein:
> > What do you want to explain?
> 
> When to use 3.x and when to use main. I think we should have that on the README.md. And the reason why we separate the track.

This can be done by another PR, my intention with this initiative was to help support for 3.x Juju.

rgildein:
Closing, since there should be proper task to do:
- add track to snapcraft.io
- add some note to document
- update workflow to publish to proper track/risk

PR #414, opened 2024-05-23
chores: fix tox.ini

chanchiwai-ray:
The `envdir` is not suppose to use it that way, and it's no longer supported in tox > 4.


jneo8 (approved):

rgildein (approved):
LGTM, can you backport this also to 3.x branch.

Issue #423, opened 2024-05-29
cou is frozen when juju action without `parameters` fails


valexby:
When action without any `parameters` fails, for example `pause` or `resume`, then cou fails to raise and exception with the trace below:
```
2024-05-29 13:15:44 [cou.steps.execute] [INFO] Running: Resume the unit: 'nova-compute/0'
2024-05-29 13:15:44 [cou.steps] [DEBUG] running step: UnitUpgradeStep(Resume the unit: 'nova-compute/0')
2024-05-29 13:16:13 [cou.utils.juju_utils] [INFO] function _get_waited_action_object failed [0]
Traceback (most recent call last):
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/utils/juju_utils.py", line 116, in wrapper
    return await func(*args, **kwargs)
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/utils/juju_utils.py", line 257, in _get_waited_action_object
    raise ActionFailed(action, output=action_obj.data)
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/exceptions.py", line 102, in __init__
    params[key] = getattr(action, key, "<not-set>")
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/juju/model.py", line 287, in __getattr__
    return self.safe_data[name]
KeyError: 'parameters'
````
And it just stays waiting on the action, e.g. `Resume the unit ...` for a long time, I waited for some 15 minutes and canceled it.

We should ensure:
1. If cou fails to raise and exception it should raise something
2. Cou should be able to raise exceptions for actions without parameters

I don't have an easy reproducer, I faced it during hypervisor upgrades from Ussuri to Victoria, facing the bug https://bugs.launchpad.net/charm-nova-compute/+bug/2067492

PR #424, opened 2024-05-29
Fix ActionFailed creating error message

rgildein:
The ActionFailed was failing to create error message when `parameters` was not present in action object. We used getattr, but this in action object is overwritten to do self.safe_data[key], so if key does not exists it raise KeyError.

fixes: #423


rgildein:
> LGTM (I analysed the code, but haven't set up an environment to test cou practically yet)
> 
> I'm not sure if this addresses this part of #423 though?:
> 
> > And it just stays waiting on the action, e.g. Resume the unit ... for a long time, I waited for some 15 minutes and canceled it.

The func test should cover this, since there are running `cou upgrade`.

I think that part was an issue of charm or something what we cannot identify now, I was helping during that upgrade and I saw more logs, and I believe there is nothing we can do now.

PR #425, opened 2024-05-30
Document how to install for development

samuelallan72:


CONTRIBUTING.md
```diff
@@ -65,8 +65,26 @@ organize them into different categories. For new contributors, we recommend
...
+you can install it in a virtual environment,
+for example:
+
+```
+virtualenv venv
+source venv/bin/activate
+pip install -e .
+
+# setuptools is required to avoid `ModuleNotFoundError: No module named 'pkg_resources'`
+pip install setuptools
```
rgildein, line 86: This is only true on newer Python 3.12, not 3.10. Can you make not for it or maybe we can add setuptools to requirements, it should be ok even in Python 3.10.

rgildein: not sure, maybe we can add it to requirements already... or add specific version to tox, so it will not work on any other python version. Only 3.10

rgildein: I think you can... We mention in setup.cfg that `python_requires = >=3.10` and `Programming Language :: Python :: 3.10`.
I think that we should add 3.12 and simply add setuptools as requirements.

rgildein (approved):

jneo8 (approved):

Issue #427, opened 2024-05-30
Most of time cou unable to complete nova-compute upgrade because `ceilometer-agent-compute` is down after upgrade


valexby:
Hi,

Because of the known bug [LP1947585](https://bugs.launchpad.net/charm-ceilometer-agent/+bug/1947585) `ceilometer-agent-compute` might be down in many cases after nova-compute release upgrade. Cou fails to complete upgrade in such a case, as the final juju `resume` action fails.

Given that the bugfix for nova-compute wasn't back-ported to Ussuri-Wallaby and the nova bug hasn't had any activity for a year, maybe we could make cou starting `ceilometer-agent` after the nova release upgrade if it is down. That is a natural thing to do for a human-operator upgrading a cloud.

jneo8:
Base on [LP1947585](https://bugs.launchpad.net/charm-ceilometer-agent/+bug/1947585), the workaround solution is `sudo systemctl restart ceilometer-agent` is it's not active.

Pjack:
Comment from @valexby 

> Without this one fixed on cou side or backported on nova-compute side, managed solutions will face this issue about ~500 nodes * 4 openstack releases = ~ 2000 times  during future upgrades

aieri:
it looks to me like [LP#1947585](https://bugs.launchpad.net/charm-ceilometer-agent/+bug/1947585) was backported [all the way back to ussuri](https://review.opendev.org/c/openstack/charm-nova-compute/+/890666), but somehow the fix isn't working for older releases. I think we should just add the workaround within COU.

jneo8:
[Discuss for implementation]

There is a limitation in implementation: **How can I know there is a ceilometer-agent unit relate to nova-compute as a subordinate in COU?**

The subordinate information is missing ]when we transform the origin juju status data into COU's Application class](https://github.com/canonical/charmed-openstack-upgrader/blob/b38da2799481aaf12e67e85c3bc9f17f8b85a686/cou/utils/juju_utils.py#L386). This create a awkward situation that I am not able to confirm if the ceilometer-agent unit is there in the same machine.

There will be two options:

1. We follow [LP1947585](https://bugs.launchpad.net/charm-ceilometer-agent/+bug/1947585)'s implementation. We put this restart service logic in the nova-compute application's post-upgrade step. But before that we need to put the subordinate information into COU's application class.
2. We create a new application for ceilometer-agent and include this subordinate application's upgrade step into hypervisor.

[...]

PR #435, opened 2024-06-11
feat: Restart ceilometer-agent-compute after nova-compute upgrade

jneo8:
Fix: #427

### DIFF
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -123,7 +123,7 @@ class NovaCompute(OpenStackApplication):
         :rtype: list[PostUpgradeStep]
         """
         if not units:
-            return []
+            units = list(self.units.values())
         steps = []
         for unit in units:
             for subordinate in unit.subordinates:
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -132,7 +132,8 @@ class NovaCompute(OpenStackApplication):
                     steps.append(
                         PostUpgradeStep(
                             description=(
-                                f"Restart subordinate service for unit: '{subordinate.name}'"
+                                "Restart service ceilometer-agent-compute "
+                                f"for subordinate unit: '{subordinate.name}'"
                             ),
                             coro=self.model.run_on_unit(
                                 unit_name=subordinate.name,
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -132,7 +132,7 @@ class NovaCompute(OpenStackApplication):
                     steps.append(
                         PostUpgradeStep(
                             description=(
-                                f"Restart subordinate service for unit: {subordinate.name}"
+                                f"Restart subordinate service for unit: '{subordinate.name}'"
                             ),
                             coro=self.model.run_on_unit(
                                 unit_name=subordinate.name,
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -127,7 +127,7 @@ class NovaCompute(OpenStackApplication):
         steps = []
         for unit in units:
             for subordinate in unit.subordinates:
-                if "ceilometer-agent" in subordinate.name:
+                if "ceilometer-agent" == subordinate.charm:
                     service = "ceilometer-agent-compute"
                     steps.append(
                         PostUpgradeStep(
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -106,7 +106,44 @@ class NovaCompute(OpenStackApplication):
         :return: List of post upgrade steps.
         :rtype: list[PostUpgradeStep]
         """
-        return self._get_enable_scheduler_step(units) + super().post_upgrade_steps(target, units)
+        return (
+            self._get_enable_scheduler_step(units)
+            + self._get_restart_subordinate_services_steps(units)
+            + super().post_upgrade_steps(target, units)
+        )
+
+    def _get_restart_subordinate_services_steps(
+        self, units: Optional[list[Unit]]
+    ) -> list[PostUpgradeStep]:
+        """Get step to restart all subordinate services if they aren't running.
+
+        :param units: Units to restart subordinate services.
+        :type units: Optional[list[Unit]]
+        :return: Steps to restart to subordinate services
+        :rtype: list[PostUpgradeStep]
+        """
+        if not units:
+            return []
+        steps = []
+        for unit in units:
+            for subordinate in unit.subordinates:
+                if "ceilometer-agent" in subordinate.name:
+                    service = "ceilometer-agent-compute"
+                    steps.append(
+                        PostUpgradeStep(
+                            description=(
+                                f"Restart subordinate service for unit: {subordinate.name}"
+                            ),
+                            coro=self.model.run_on_unit(
+                                unit_name=subordinate.name,
+                                command=(
+                                    f"systemctl is-active --quiet {service}"
+                                    f" || systemctl restart {service}"
+                                ),
+                            ),
+                        )
+                    )
+        return steps
 
     def _get_unit_upgrade_steps(self, unit: Unit, force: bool) -> UnitUpgradeStep:
         """Get the upgrade steps for a single unit.
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -19,9 +19,9 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Any, Callable, Optional
+from typing import Any, Callable, List, Optional
 
 from juju.action import Action
 from juju.application import Application as JujuApplication
@@ -143,12 +143,29 @@ class Machine:
 
 
 @dataclass(frozen=True)
+class SubordinateUnit:
+    """Representation of a single unit of subordinate unit."""
+
+    name: str
+    workload_version: str
+
+    def __repr__(self) -> str:
+        """App representation.
+
+        :return: Name of the application
+        :rtype: str
+        """
+        return self.name
+
+
+@dataclass(frozen=True)
 class Unit:
     """Representation of a single unit of application."""
 
     name: str
     machine: Machine
     workload_version: str
+    subordinates: List[SubordinateUnit] = field(default_factory=lambda: [], compare=False)
 
     def __repr__(self) -> str:
         """App representation.
@@ -367,6 +384,13 @@ class Model:
         full_status = await self.get_status()
         machines = await self._get_machines()
 
+        for app, status in full_status.applications.items():
+            logger.debug(app)
+            for name, unit in status.units.items():
+                logger.debug(name)
+                logger.debug(unit)
+                logger.debug(unit.subordinates)
+
         return {
             app: Application(
                 name=app,
@@ -383,7 +407,15 @@ class Model:
                 series=status.series,
                 subordinate_to=status.subordinate_to,
                 units={
-                    name: Unit(name, machines[unit.machine], unit.workload_version)
+                    name: Unit(
+                        name,
+                        machines[unit.machine],
+                        unit.workload_version,
+                        [
+                            SubordinateUnit(subordinate_name, subordinate.workload_version)
+                            for subordinate_name, subordinate in unit.subordinates.items()
+                        ],
+                    )
                     for name, unit in status.units.items()
                 },
                 workload_version=status.workload_version,
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -406,8 +406,7 @@ class Model:
                         unit.workload_version,
                         [
                             SubordinateUnit(
-                                subordinate,
-                                model.applications[subordinate.split("/")[0]]
+                                subordinate, model.applications[subordinate.split("/")[0]]
                             )
                             for subordinate, subordinate_unit in unit.subordinates.items()
                         ],
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -147,6 +147,7 @@ class SubordinateUnit:
     """Representation of a single unit of subordinate unit."""
 
     name: str
+    charm: str
     workload_version: str
 
     def __repr__(self) -> str:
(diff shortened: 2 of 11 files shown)


cou/apps/core.py
```diff
@@ -106,7 +106,44 @@ def post_upgrade_steps(
...
+        :type units: Optional[list[Unit]]
+        :return: Steps to restart to subordinate services
+        :rtype: list[PostUpgradeStep]
+        """
+        if not units:
+            return []
+        steps = []
+        for unit in units:
+            for subordinate in unit.subordinates:
+                if "ceilometer-agent" in subordinate.name:
```
gabrielcocenza, line 130: relying on application name is not robust. Unfortunately it seems that there is no link on python libjuju between the subordinate unit and the actual charm name. I tried using juju 3.1 and the charm field is empty:

```shell
>>> status.applications["ubuntu"].units["ubuntu/0"].subordinates["hardware-observer/0"].charm
''
```

cou/utils/juju_utils.py
```diff
@@ -142,13 +142,30 @@ class Machine:
     az: Optional[str] = None  # simple deployments may not have azs
 
 
+@dataclass(frozen=True)
+class SubordinateUnit:
+    """Representation of a single unit of subordinate unit."""
+
+    name: str
+    workload_version: str
```
gabrielcocenza, line 150: I'm not seeing any use for workload_version, so let's not add it if not necessary.

cou/apps/core.py
```diff
@@ -106,7 +106,44 @@ def post_upgrade_steps(
...
         :rtype: list[PostUpgradeStep]
         """
-        return self._get_enable_scheduler_step(units) + super().post_upgrade_steps(target, units)
+        return (
+            self._get_enable_scheduler_step(units)
+            + self._get_restart_subordinate_services_steps(units)
+            + super().post_upgrade_steps(target, units)
+        )
+
+    def _get_restart_subordinate_services_steps(
```
gabrielcocenza, line 115: I think renaming it to `_get_restart_ceilometer_agent_steps` it's cleaner and more accurate

gabrielcocenza (changes requested):
I'm not sure which way it's easier to implement, but we can't rely upon application name. I see two options:

1. Somehow include the charm name during the instantiation of `SubordinateUnit`
2. Don't implement the `SubordinateUnit` class and pass just the nova-compute units to a function that will access the juju status, find the ceilometer-agents and restart if necessary

I like the option 1 because it looks flexible if we need to do more things with subordinates units when upgrading a principal, but this might require more unit tests

cou/utils/juju_utils.py
```diff
@@ -367,6 +384,13 @@ async def get_applications(self) -> dict[str, Application]:
         full_status = await self.get_status()
         machines = await self._get_machines()
 
+        for app, status in full_status.applications.items():
+            logger.debug(app)
+            for name, unit in status.units.items():
+                logger.debug(name)
+                logger.debug(unit)
+                logger.debug(unit.subordinates)
+
```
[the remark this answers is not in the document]
jneo8: Deleted

cou/apps/core.py
jneo8: Changed to charm_name on newest commit.

cou/utils/juju_utils.py
jneo8: I would like to keep this one, it's like a basic information of the unit.

cou/apps/core.py
jneo8: If you check the implementation [here](https://review.opendev.org/c/openstack/charm-nova-compute/+/833886), I think `_get_restart_subordinate_services_steps` give us more flexible. I would like to keep this one.

gabrielcocenza: The implementation that you shared is `restarting all subordinate services if they aren't running` whereas this is very specific for ceilometer-agent. If we are going to restart all subordinates services like they are doing than I agree with the name.

cou/utils/juju_utils.py
gabrielcocenza: A lot of subordinate apps doesn't even have workload_version and we are determining the OpenStack release by the channel. It's not necessary and most likely will never be used

cou/utils/juju_utils.py
```diff
@@ -354,7 +381,7 @@ async def connect(self) -> None:
             retry_backoff=DEFAULT_MODEL_RETRY_BACKOFF,
         )
 
-    @retry
+    # @retry
```
gabrielcocenza, line 384: Accidentally left it commented?

cou/utils/juju_utils.py
```diff
@@ -383,7 +416,19 @@ async def get_applications(self) -> dict[str, Application]:
...
                 units={
-                    name: Unit(name, machines[unit.machine], unit.workload_version)
+                    name: Unit(
+                        name,
+                        machines[unit.machine],
+                        unit.workload_version,
+                        [
+                            SubordinateUnit(
+                                subordinate_name,
+                                unit_charm_mapping[subordinate_name],
```
gabrielcocenza, line 426: isn't possible to just use `model.applications[subordinate_name].charm` ?

Than it's not necessary to use `_get_applications` and this dict that you created

jneo8: No, the `subordinate_name` is the unit name, like `app/0`.
Also the data is already been fetch to local. Compare to handling the unit-name to application-name string transformer, there is no different.

cou/apps/core.py
jneo8: Then let's keep it.

cou/utils/juju_utils.py
jneo8: For me it's quiet weird that those two different unit class have two different attributes.

Here I will delete the workload_version since you persist in. But I will say the design of the dataclass in juju_utils have is wrong.

jneo8: Removed.

cou/utils/juju_utils.py
```diff
@@ -383,7 +415,18 @@ async def get_applications(self) -> dict[str, Application]:
...
+                            )
+                            for subordinate_name, subordinate in unit.subordinates.items()
```
gabrielcocenza, line 427: Not necessary to use `subordinate` anymore
```suggestion
                            for subordinate_unit in unit.subordinates
```

gabrielcocenza (changes requested):
I realized that for some reason the new post upgrade steps are not showing into the mocked_plans, hypervisors and core unit tests. Probably you need to add ceilometer subordinate units.

You also need to include the subordinates new field into the string representation of OpenStackApplication so in the future we can get this new information and replicate into mocked tests. I'm ok not updating on all mocked plans now, but I would like to see at least one changed and working as expected.

cou/utils/juju_utils.py
gabrielcocenza: > For me it's quiet weird that those two different unit class have two different attributes.

Juju itself treat totally different subordinate units from principal units. If you access the status of a subordinate unit, it will say that there isn't units. When you access by model.applications it will have units, but it won't have workload_version. Now you showed that by accessing this way potentially show workload_version.

If you want I'm ok to inherit from the Unit or use Unit itself and populate with all the fields, but it might be more complex.

jneo8:
> hypervisors

Can you provide more details here? I am not sure which test function need to change from this comment.

jneo8:
> I realized that for some reason the new post upgrade steps are not showing into the mocked_plans, hypervisors and core unit tests. Probably you need to add ceilometer subordinate units.
> 
> You also need to include the subordinates new field into the string representation of OpenStackApplication so in the future we can get this new information and replicate into mocked tests. I'm ok not updating on all mocked plans now, but I would like to see at least one changed and working as expected.

Can you provide more details here? I am not sure which test functions need to change here.

chanchiwai-ray (approved):

cou/apps/core.py
```diff
@@ -106,7 +106,44 @@ def post_upgrade_steps(
...
+                            description=(
+                                f"Restart subordinate service for unit: '{subordinate.name}'"
+                            ),
+                            coro=self.model.run_on_unit(
+                                unit_name=subordinate.name,
+                                command=(
+                                    f"systemctl is-active --quiet {service}"
+                                    f" || systemctl restart {service}"
+                                ),
+                            ),
```
rgildein, line 143: I think that this step could be simpler, saying: "Restart subordinate service for ceilometer-agent-compute units".
Like this we will not need to have `SubordinateUnit` and the coroutine here will simply run over all (if NovaCompute is run with units, all here will mean all subordinate to those units) subordinates units of ceilometer-agent-compute.

Like this, we lose information about specific unit in step, but for operator such information is not really relevant I think. It must properly say that it will restart only related units. The implementation will be simpler.

rgildein (changes requested):
Adding one comment, which can simplify this approach, but it needs to be discussed. Currently, this changes bring new features, which also needs to be discussed.

cou/apps/core.py
jneo8: I see your point. The complexity your proposal want to avoid is because the data transforming layer in COU, which I 100% agree with you it is a hard decision to change it.

So let me explain my:

- I would like to make each restart service as a single step, the plans is more clear to me.
- The implementation I gave doesn't require more effort to get the data from the controller.
- This SubordinatesUnit design provide more flexible possible when we need to upgrade the hypervisor.

I wouldn't say which one is better, just trade-off.

I will let other people provide their feedback.

rgildein: Your points are correct, and I'm not against them, I just try to avoid COU being more complex.

jneo8:
Hi @gabrielcocenza , can I get your review again? All the comments from you are been addressed.

cou/apps/core.py
jneo8: Hi @rgildein 

Since no other provided their feedback. I will try to pitch my idea:

There is a major differences between our proposal:

I want to show the clear plan, every service restart should have a single COU step. Without the `SubordinateUnit`, the possible way will be checking if the service exists in the unit. This require more networking effort at both planning and upgrading.

(If I misunderstanding your proposal, please correct me)

cou/utils/juju_utils.py
jneo8: Resolved because it's done.

tests/unit/apps/test_core.py
```diff
@@ -708,6 +716,32 @@ def test_nova_compute_get_enable_scheduler_step(model, units):
...
+    ],
+)
+def test_nova_compute_get_restart_subordinate_services_steps(model, units):
+    app = _generate_nova_compute_app(model)
+    units_selected = [app.units[unit] for unit in units]
+    assert app._get_restart_subordinate_services_steps(units_selected) == [
+        PostUpgradeStep(
+            description=f"Restart subordinate service for unit: '{unit.subordinates[0].name}'",
+            coro=model.run_on_unit(
+                unit_name=unit.subordinates[0].name,
```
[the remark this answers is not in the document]
jneo8: Here I would like to keep it because it's generate from the parametrize input.
The testing logic include we want to make sure all the steps will be generate for every units.

tests/unit/utils/test_juju_utils.py
```diff
@@ -600,9 +600,10 @@ async def test_get_machines(mocked_model):
     assert machines == expected_machines
 
 
-def _generate_juju_unit(app: str, machine_id: str) -> MagicMock:
+def _generate_juju_unit(app: str, id: str, machine_id: str) -> MagicMock:
```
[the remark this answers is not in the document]
chanchiwai-ray: `uuid` maybe?

tests/mocked_plans/sample_plans/018346c5-f95c-46df-a34e-9a78bdec0018.yaml
```diff
@@ -453,6 +461,10 @@ plan: |
                 Enable nova-compute scheduler from unit: 'nova-compute-kvm/6'
                 Enable nova-compute scheduler from unit: 'nova-compute-kvm/7'
                 Enable nova-compute scheduler from unit: 'nova-compute-kvm/8'
+                Restart subordinate service for unit: 'ceilometer-agent/4'
+                Restart subordinate service for unit: 'ceilometer-agent/6'
+                Restart subordinate service for unit: 'ceilometer-agent/7'
+                Restart subordinate service for unit: 'ceilometer-agent/8'
```
chanchiwai-ray, line 467: nitpick: I don't really understand the words here.
suggestion: `Restart service for subordinate unit: xxx`

cou/apps/core.py
chanchiwai-ray: non-blocking, just a comment: for me, I would simple add a post upgrade step to restart ceilometer agent-compute service in `NovaCompute` class only, since we know that service will be in `NovaCompute` unit,

tests/unit/utils/test_juju_utils.py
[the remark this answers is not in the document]
jneo8: Change to `unit_id`

chanchiwai-ray (approved):

tests/mocked_plans/sample_plans/018346c5-f95c-46df-a34e-9a78bdec0018.yaml
jneo8: Updated

cou/apps/core.py
jneo8: > since we know that service will be in NovaCompute unit

The `ceilometer-agent-compute` service exists only if `ceilometer-agent` is subordinate to nova-compute. You can not know this information without using SubordinateUnit or collecting more information from the unit.

chanchiwai-ray: check exit status of `systemctl list-unit-files ceilometer-agent-compute`

chanchiwai-ray (approved):

cou/apps/core.py
gabrielcocenza: my opinion here is that the implementation of this `SubordinateUnit`indeed brings a little bit more complexity, but it opens more doors if in the future we need to be smarter about other subordinate units. Will this happen? I don't think it will, but it's already implemented and it's more connected with what libjuju does.

gabrielcocenza (approved):
LGTM. Thanks

cou/apps/core.py
rgildein: > The ceilometer-agent-compute service exists only if ceilometer-agent is subordinate to nova-compute. You wouldn't know this information without using SubordinateUnit or collecting more information from the unit.

You can simply check if such charm is subordinate to it or not. Step will check if there is any and if yes restart it. Operator will see the description saying: "Check subordinate ceilometer-agent and restart services ".

rgildein: After discussion, we agreed that a simpler solution would be more appropriate, but since it's already done, it's good to go.

rgildein (approved):
LGTM

Issue #446, opened 2024-06-19
Custom object `subordinate.charm` cannot be dumped into yaml format


chanchiwai-ray:
Regression from #435 

Running `cou plan` or `cou upgrade` will produce the following error:

```
Traceback (most recent call last):                                                                                                                                                                                                                                                        
  File "/usr/lib/python3.10/logging/__init__.py", line 1100, in emit                                                                                                                                                                                                                      
    msg = self.format(record)                                                                                                                                                                                                                                                             
  File "/usr/lib/python3.10/logging/__init__.py", line 943, in format                                                                                                                                                                                                                     
    return fmt.format(record)                                                                                                                                                                                                                                                             
  File "/usr/lib/python3.10/logging/__init__.py", line 678, in format                                                                                                                                                                                                                     
    record.message = record.getMessage()                                                                                                                                                                                                                                                  
  File "/usr/lib/python3.10/logging/__init__.py", line 368, in getMessage                                                                                                                                                                                                                 
    msg = msg % self.args                                                                                                                                                                                                                                                                 
  File "/home/ubuntu/workspace/charmed-openstack-upgrader/cou/apps/base.py", line 147, in __str__                                                                                                                                                                                         
    return yaml.dump(summary, sort_keys=False)                                                                                                                                                                                                                                            
  File "/home/ubuntu/workspace/charmed-openstack-upgrader/.venv/lib/python3.10/site-packages/yaml/__init__.py", line 253, in dump                                                                                                                                                         
    return dump_all([data], stream, Dumper=Dumper, **kwds)                                                                                                                                                                                                                                
  File "/home/ubuntu/workspace/charmed-openstack-upgrader/.venv/lib/python3.10/site-packages/yaml/__init__.py", line 241, in dump_all                                                                                                                                                     
    dumper.represent(data)                                                                                                                                                                                                                                                                
  File "/home/ubuntu/workspace/charmed-openstack-upgrader/.venv/lib/python3.10/site-packages/yaml/representer.py", line 27, in represent                                                                                                                                                  
    node = self.represent_data(data)                                                                                                                                                                                                                                                      
  File "/home/ubuntu/workspace/charmed-openstack-upgrader/.venv/lib/python3.10/site-packages/yaml/representer.py", line 48, in represent_data                                                                                                                                             
    node = self.yaml_representers[data_types[0]](self, data)                                                                                                                                                                                                                              
[...]

PR #447, opened 2024-06-20
Fix dumping custom object into yaml format.

chanchiwai-ray:
fixes: #446


jneo8:
I kind of confuse, the charm is a string type. Where is the `name` come from?

chanchiwai-ray:
mypy did not spot this typing issue in the first place

jneo8 (approved):

gabrielcocenza (approved):
Thanks for the fix

PR #453, opened 2024-06-25
Document how to build and deploy the snap

samuelallan72:


samuelallan72:
Revisiting this - I think it would be good to get this into the contributing guide at least.  We can move it later if we want, but there is value in this being documented somewhere at least. :)

CONTRIBUTING.md
```diff
@@ -92,3 +92,16 @@ pip install -e .
 # run cou!
 cou --version
 ```
+
+## Testing the snap
+
+Installing a local snap has some manual steps because it must be installed in 'dangerous' mode:
```
chanchiwai-ray, line 98: I think you can install with devmode, or not?

samuelallan72: You can install with devmode, but it's safer to install with `--dangerous`.

- `--dangerous`: installs the snap even if signatures from the snap store are missing
- `--devmode`: disables security sandbox, implicitly adds `--dangerous` too.

chanchiwai-ray: So maybe this is explanation why the developer should test with `dangerous` mode? I feel like "must" mean they don't have other choices, but here I think we are recommending them to test with `dangerous` mode

samuelallan72: Yeah maybe the wording could be improved - what I mean there is that it cannot be installed normally - you must add the `--dangerous` flag (or of course `--devmode`).

chanchiwai-ray: Thanks

chanchiwai-ray (approved):

jneo8 (approved):

CONTRIBUTING.md
samuelallan72: Thanks for your review :)

PR #454, opened 2024-06-26
Re-add the comment about juju max frame size

samuelallan72:
This comment gives context to why a custom frame size is used. It seems to have been lost during a refactor,
so lets add it back.

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -46,7 +46,7 @@ from cou.exceptions import (
 )
 from cou.utils.openstack import is_charm_supported
 
-# Default for the Juju MAX_FRAME_SIZE to be 256MB to stop
+# Increase Juju websocket connection MAX_FRAME_SIZE to 1024MiB to stop
 # "RPC: Connection closed, reconnecting" errors and then a failure in the log.
 # See https://github.com/juju/python-libjuju/issues/458 for more details
 JUJU_MAX_FRAME_SIZE: int = 2**30


jneo8 (changes requested):
Can this be a cli argument?

rgildein:
> Can this be a cli argument?

I do not think that cli arg is good idea, but maybe we can do environment argument.

cou/utils/juju_utils.py
```diff
@@ -46,6 +46,9 @@
 )
 from cou.utils.openstack import is_charm_supported
 
+# Default for the Juju MAX_FRAME_SIZE to be 256MB to stop
+# "RPC: Connection closed, reconnecting" errors and then a failure in the log.
+# See https://github.com/juju/python-libjuju/issues/458 for more details
 JUJU_MAX_FRAME_SIZE: int = 2**30
```
chanchiwai-ray, line 52: What is the unit of `JUJU_MAX_FRAME_SIZE` and how is it related to the comment you add back?

chanchiwai-ray (commented):
Why are we discussing this as a cli argument? Isn't this about re-adding comment?

jneo8 (dismissed):

cou/utils/juju_utils.py
chanchiwai-ray: Free feel to resolve it when you have or don't have the answer 🙂 but we should keep the comment correct

chanchiwai-ray: If the source code is correct: https://github.com/juju/python-libjuju/blob/master/juju/client/connection.py#L231-L242, then the unit should be `byte`

chanchiwai-ray (approved):

gabrielcocenza (approved):

PR #469, opened 2024-06-30
Add tmate debugging option for functional tests

aieri:


samuelallan72:
@aieri what's the status of this PR? Is this something you want to pursue and merge, or was it just temporary for debugging? :)

aieri:
it's actually something I think we may want to roll out to all projects, but had forgotten about. I'll close it for now, we can revisit it at a later date

Issue #467, opened 2024-06-28
[vault] Refreshing vault from 1.7/stable to 1.8/stable cause vault to be in error state, and sealed


chanchiwai-ray:
When COU refresh vault from 1.7/stable to 1.8/stable during Jammy/Yoga -> Jammy/Zed upgrade, the vault charm goes into error state with `Connection Refused` error:

```
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm Traceback (most recent call last):                                                 
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm   File "/var/lib/juju/agents/unit-vault-0/.venv/lib/python3.10/site-packages/requests/adapters.py", line 439, in send                                                                                                           
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm     resp = conn.urlopen(                                                           
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm   File "/var/lib/juju/agents/unit-vault-0/.venv/lib/python3.10/site-packages/urllib3/connectionpool.py", line 787, in urlopen                                                                                                   
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm     retries = retries.increment(                                                   
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm   File "/var/lib/juju/agents/unit-vault-0/.venv/lib/python3.10/site-packages/urllib3/util/retry.py", line 592, in increment                                                                                                     
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm     raise MaxRetryError(_pool, url, error or ResponseError(cause))                 
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='127.0.0.1', port=8220): Max retries exceeded with url: /v1/auth/approle/login (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f9bba586bf0>: Failed to establish a new connection: [Errno 111] Connection refused'))                                                                
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm                                                                                    
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm During handling of the above exception, another exception occurred:                
unit-vault-0: 03:48:42 WARNING unit.vault/0.upgrade-charm                                                                                    
[...]

PR #485, opened 2024-07-09
fix: unseal vault

jneo8:
* Move vault to the first upgraded application
* Stop upgraded if vault is sealed
* Add unseal step to Vault upgrading process

fix: #467

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -62,6 +62,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install tox
+        # vault is required to pass the command exists check
+        snap install vault
     - name: Run mocked-plans tests
       run: tox -e mocked-plans
 
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -126,9 +126,5 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: SNAP_FILE
-      - name: Install dependencies
-        run: |
-          # vault is required to pass the command exists check
-          sudo snap install vault
       - name: Run func tests
         run: TEST_SNAP=$GITHUB_WORKSPACE/charmed-openstack-upgrader.snap tox -e func
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -126,5 +126,9 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: SNAP_FILE
+      - name: Install dependencies
+        run: |
+          # vault is required to pass the command exists check
+          sudo snap install vault
       - name: Run func tests
         run: TEST_SNAP=$GITHUB_WORKSPACE/charmed-openstack-upgrader.snap tox -e func
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -63,7 +63,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install tox
         # vault is required to pass the command exists check
-        snap install vault
+        sudo snap install vault
     - name: Run mocked-plans tests
       run: tox -e mocked-plans
 
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -14,14 +14,22 @@
 """Auxiliary application class."""
 import abc
 import logging
+import re
+import subprocess
 from typing import Optional
 
 from packaging.version import Version
 
 from cou.apps.base import LONG_IDLE_TIMEOUT, OpenStackApplication
 from cou.apps.factory import AppFactory
-from cou.exceptions import ApplicationError
+from cou.exceptions import (
+    ApplicationError,
+    SnapVaultNotInstalled,
+    VaultGetStatusFailed,
+    VaultUnsealFailed,
+)
 from cou.steps import ApplicationUpgradePlan, PostUpgradeStep, PreUpgradeStep
+from cou.utils import progress_indicator
 from cou.utils.app_utils import set_require_osd_release_option
 from cou.utils.juju_utils import Unit
 from cou.utils.openstack import (
@@ -34,7 +42,7 @@ from cou.utils.openstack import (
 logger = logging.getLogger(__name__)
 
 
-@AppFactory.register_application(["vault", "ceph-fs", "ceph-radosgw"])
+@AppFactory.register_application(["ceph-fs", "ceph-radosgw"])
 class AuxiliaryApplication(OpenStackApplication):
     """Application for charms that can have multiple OpenStack releases for a workload."""
 
@@ -457,3 +465,142 @@ class CephOsd(AuxiliaryApplication):
             raise ApplicationError(
                 f"Units '{', '.join(units_not_upgraded)}' did not reach {target}."
             )
+
+
+@AppFactory.register_application(["vault"])
+class Vault(AuxiliaryApplication):
+    """Application for vault."""
+
+    wait_timeout = LONG_IDLE_TIMEOUT
+    wait_for_model = True
+
+    def _check_snap_vault_installed(self) -> None:
+        """Make sure snap vault is installed.
+
+        :raises SnapVaultNotInstalled: When no snap valut installed.
+        """
+        result = subprocess.run(
+            "snap info vault".split(),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            check=True,
+        )
+        if "installed" not in result.stdout:
+            raise SnapVaultNotInstalled("snap vault is required to upgrade the vault")
+
+    async def _wait_for_sealed_status(self) -> None:
+        """Wait for application vault go into sealed status.
+
+        :raises ApplicationError: When application vault is not in sealed status.
+        """
+        await self.model.wait_for_idle(
+            timeout=self.wait_timeout,
+            status="blocked",
+            apps=[self.name],
+            # Vault application will go into error status first then blocked status.
+            raise_on_error=False,
+        )
+
+        app_status = await self.model.get_application_status(charm_name="vault")
+        if not app_status.status.info == "Unit is sealed":
+            raise ApplicationError("Vault not in sealed status")
+        logger.debug("Application 'vault' in sealed status")
+
+    async def _unseal_vault(self) -> None:
+        """Unseal vault on every vault unit.
+
+        :raises VaultGetStatusFailed: When failed to get sealed status from vault command
+        :raises VaultUnsealFailed: When unseal command failed
+        """
+        progress_indicator.stop()
+
+        for unit_name in self.units:
+            logger.debug("Start unseal %s", unit_name)
+            juju_unit = await self.model.get_unit(unit_name)
+            vault_address = f"http://{juju_unit.public_address}:8200"
+
+            while True:
+                status_result = subprocess.run(
+                    "/snap/bin/vault status".split(),
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    env={"VAULT_ADDR": vault_address},
+                    text=True,
+                    check=True,
+                )
+                sealed_match = re.search(
+                    r"(Sealed)\s+(?P<sealed>(true)|(false))", status_result.stdout
+                )
+                if not sealed_match:
+                    raise VaultGetStatusFailed("Failed to get vault sealed status")
+
+                if sealed_match.group("sealed").lower() == "false":
+                    logger.debug("Unseal succeeded %s", unit_name)
+                    break
+
+                unseal_result = subprocess.run(
+                    "/snap/bin/vault operator unseal".split(),
+                    env={"VAULT_ADDR": vault_address},
+                    check=False,
+                )
+                if unseal_result.returncode != 0:
+                    raise VaultUnsealFailed("Unseal vault failed")
+
+    def post_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[Unit]]
+    ) -> list[PostUpgradeStep]:
+        """Post Upgrade steps planning.
+
+        Wait until the application reaches the idle state and then check the target workload.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate post upgrade plan
+        :type units: Optional[list[Unit]]
+        :return: List of post upgrade steps.
+        :rtype: list[PostUpgradeStep]
+        """
+        steps = [
+            # Vault application should get into blocked and sealed status after upgrading.
+            PostUpgradeStep(
+                description="wait for sealed status",
+                coro=self._wait_for_sealed_status(),
+            ),
+            PostUpgradeStep(
+                description="unseal vault",
+                coro=self._unseal_vault(),
+            ),
+            PostUpgradeStep(
+                description="wait for vault active",
+                coro=self.model.wait_for_idle(
+                    timeout=self.wait_timeout,
+                    status="active",
+                    apps=[self.name],
+                    raise_on_blocked=False,
+                    raise_on_error=False,
+                ),
+            ),
+            # Some applications will get into error status because vault in sealed status.
+            # Need to resolve them.
+            PostUpgradeStep(
+                description="Resolve all",
+                coro=self.model.resolve_all(),
+            ),
+        ]
+        steps.extend(super().post_upgrade_steps(target, units))
+        return steps
+
+    def upgrade_plan_sanity_checks(
+        self, target: OpenStackRelease, units: Optional[list[Unit]]
+    ) -> None:
+        """Run sanity checks before generating upgrade plan.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan, defaults to None
+        :type units: Optional[list[Unit]], optional
+        :raises SnapVaultNotInstalled: When the snap vault is not installed.
+        """
+        self._check_snap_vault_installed()
+        super().upgrade_plan_sanity_checks(target, units)
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -24,7 +24,7 @@ from cou.apps.base import LONG_IDLE_TIMEOUT, OpenStackApplication
 from cou.apps.factory import AppFactory
 from cou.exceptions import (
     ApplicationError,
-    SnapVaultNotInstalled,
+    CommandVaultNotFound,
     VaultGetStatusFailed,
     VaultUnsealFailed,
 )
@@ -474,20 +474,20 @@ class Vault(AuxiliaryApplication):
     wait_timeout = LONG_IDLE_TIMEOUT
     wait_for_model = True
 
-    def _check_snap_vault_installed(self) -> None:
+    def _check_vault_command_exists(self) -> None:
         """Make sure snap vault is installed.
 
-        :raises SnapVaultNotInstalled: When no snap valut installed.
+        :raises CommandVaultNotFound: When no snap valut installed.
         """
         result = subprocess.run(
-            "snap info vault".split(),
+            "vault --help".split(),
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
             check=True,
         )
-        if "installed" not in result.stdout:
-            raise SnapVaultNotInstalled("snap vault is required to upgrade the vault")
+        if result.returncode != 0:
+            raise CommandVaultNotFound("vault is required to upgrade the vault")
 
     async def _wait_for_sealed_status(self) -> None:
         """Wait for application vault go into sealed status.
@@ -522,7 +522,7 @@ class Vault(AuxiliaryApplication):
 
             while True:
                 status_result = subprocess.run(
-                    "/snap/bin/vault status".split(),
+                    "vault status".split(),
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     env={"VAULT_ADDR": vault_address},
@@ -540,7 +540,7 @@ class Vault(AuxiliaryApplication):
                     break
 
                 unseal_result = subprocess.run(
-                    "/snap/bin/vault operator unseal".split(),
+                    "vault operator unseal".split(),
                     env={"VAULT_ADDR": vault_address},
                     check=False,
                 )
@@ -604,7 +604,7 @@ class Vault(AuxiliaryApplication):
         :type target: OpenStackRelease
         :param units: Units to generate upgrade plan, defaults to None
         :type units: Optional[list[Unit]], optional
-        :raises SnapVaultNotInstalled: When the snap vault is not installed.
+        :raises CommandVaultNotFound: When the snap vault is not installed.
         """
-        self._check_snap_vault_installed()
+        self._check_vault_command_exists()
         super().upgrade_plan_sanity_checks(target, units)
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -16,6 +16,7 @@ import abc
 import base64
 import getpass
 import logging
+import os
 import tempfile
 from typing import Optional
 
@@ -477,12 +478,13 @@ class Vault(AuxiliaryApplication):
         :return: Temporary file path
         :rtype: str
         """
-        cacert_b64 = self.config["ssl-ca"].get("vault")
+        cacert_b64 = self.config["ssl-ca"].get("value")
         cacert_file = None
         if cacert_b64:
             with tempfile.NamedTemporaryFile(mode="wb", delete=False) as fp:
                 fp.write(base64.b64decode(cacert_b64))
                 cacert_file = fp.name
+                logger.debug("Create tempfile: %s", cacert_file)
         return cacert_file
 
     async def _get_unit_api_url(self, unit_name: str) -> str:
@@ -493,16 +495,15 @@ class Vault(AuxiliaryApplication):
         :return: unit's vault api url
         :rtype: str
         """
-        # Use vip as address if vip is being used.
-        if self.config["vip"].get("value"):
+        if self.config["hostname"].get("value"):  # Use hostname if hostname is being used.
+            address = self.config["hostname"].get("value")
+        elif self.config["vip"].get("value"):  # Use vip as address if vip is being used.
             address = self.config["vip"].get("value")
         else:
             juju_unit = await self.model.get_unit(unit_name)
             address = juju_unit.public_address
 
-        transport = "http"
-        if self.config["ssl-cert"].get("value"):
-            transport = "https"
+        transport = "https" if self.config["ssl-cert"].get("value") else "http"
         return f"{transport}://{address}:8200"
 
     async def _wait_for_sealed_status(self) -> None:
@@ -523,40 +524,42 @@ class Vault(AuxiliaryApplication):
             raise ApplicationError("Vault not in sealed status")
         logger.debug("Application 'vault' in sealed status")
 
-    async def _get_vault_client(self, unit_name: str) -> hvac.Client:
+    async def _get_vault_client(self, unit_name: str, cacert_file: Optional[str]) -> hvac.Client:
         """Get vault client.
 
         :param unit_name: vault unit name
         :type unit_name: str
+        :param cacert_file: cacert file path
+        :type cacert_file: str
         :return: hvac vault Client object
         :rtype: hvac.Client
         """
         vault_url = await self._get_unit_api_url(unit_name)
-        cacert_file = self._get_cacert_file()
-
         client = hvac.Client(url=vault_url, verify=cacert_file)
         return client
 
     async def _unseal_vault(self) -> None:
-        """Unseal vault on every vault unit.
-
-        :raises VaultGetStatusFailed: When failed to get sealed status from vault command
-        :raises VaultUnsealFailed: When unseal command failed
-        """
+        """Unseal vault on every vault unit."""
+        # Stop progress_indicator because it will clear the unseal key input.
         progress_indicator.stop()
 
         for unit_name in self.units:
             logger.debug("Start unseal %s", unit_name)
-            client = await self._get_vault_client(unit_name)
+            cacert_file = self._get_cacert_file()
+            client = await self._get_vault_client(unit_name, cacert_file)
             while True:
-                status = client.sys.read_seal_status()  # pylint: disable=no-member
-                if status["sealed"]:
-                    unseal_key = getpass.getpass("Unseal Key (will be hidden):")
-                    if unseal_key:
-                        client.sys.submit_unseal_key(key=unseal_key)  # pylint: disable=no-member
-                    del unseal_key
-                else:
+                status = client.sys.read_seal_status()
+                if not status["sealed"]:
                     break
+                unseal_key = getpass.getpass("Unseal Key (will be hidden):")
+                if unseal_key:
+                    client.sys.submit_unseal_key(key=unseal_key)
+                del unseal_key
+
+            # Delete temporary ca file
+            if cacert_file:
+                os.remove(cacert_file)
+                logger.debug("Remove tempfile: %s", cacert_file)
 
     def post_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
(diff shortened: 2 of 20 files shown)


cou/utils/openstack.py
```diff
@@ -47,7 +47,7 @@
 DATA_PLANE_CHARMS = ["nova-compute", "ceph-osd", "swift-proxy", "swift-storage"]
 
 # https://docs.openstack.org/charm-guide/latest/admin/upgrades/openstack.html#list-the-upgrade-order
-UPGRADE_ORDER = [
+OFFICIAL_UPGRADE_ORDER = [
```
chanchiwai-ray, line 50: Nit: consider changing it to `UCA_UPGRADE_ORDER` or `OPENSTACK_SERVICES_UPGRADE_ORDER` (see https://docs.openstack.org/charm-guide/latest/admin/upgrades/openstack.html#upgradable-services)

chanchiwai-ray: on a second note, I found that the `UPGRADE_ORDER` does not follow what's in the documentation: (e.g. `rabbitmq-server` is not an openstack services) and should not be in the `UPGRADE_ORDER`

cou/exceptions.py
```diff
@@ -147,3 +147,22 @@ def __init__(self, message: str, exit_code: int) -> None:
 
 class ApplicationNotSupported(COUException):
     """COU exception when the application is known but not supported by COU."""
+
+
+class SnapVaultNotInstalled(COUException):
```
chanchiwai-ray, line 152: We should not limit the user to use snap

cou/apps/auxiliary.py
```diff
@@ -457,3 +465,142 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+
+    wait_timeout = LONG_IDLE_TIMEOUT
+    wait_for_model = True
+
+    def _check_snap_vault_installed(self) -> None:
+        """Make sure snap vault is installed.
+
+        :raises SnapVaultNotInstalled: When no snap valut installed.
+        """
+        result = subprocess.run(
```
chanchiwai-ray, line 482: we should not limit the user to use snap

cou/apps/auxiliary.py
```diff
@@ -457,3 +465,142 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+            "snap info vault".split(),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            check=True,
+        )
+        if "installed" not in result.stdout:
+            raise SnapVaultNotInstalled("snap vault is required to upgrade the vault")
+
+    async def _wait_for_sealed_status(self) -> None:
```
chanchiwai-ray, line 492: since you created `steps/vault.py` I would recommend that you move vault related function to that file

chanchiwai-ray (commented):
Left some quick comments

cou/exceptions.py
jneo8: It only fire during the vault upgrading, and we use snap vault to unseal. So it's required.

cou/apps/auxiliary.py
jneo8: What we have in `step/vault.py` is a pre-check before all the upgrading.

The required step will be in the `Vault` o7k application class.

cou/apps/auxiliary.py
samuelallan72: +1 I agree.  We should only check:

- if we can run the `vault` command
- if so, if `vault` is a compatible version (this can be out of scope here; we can follow up later with version checks)

jneo8: We force because it has better security.

samuelallan72: Sorry I don't agree; cou should not be reliant on a single distribution of vault.  Some customer clouds may have different requirements and have vault installed via a different method.

It would be a hard requirement for me that cou does not depend on the vault snap.

samuelallan72: If we want extra security, we can warn the user if vault is found but it's not installed from the snap.  We can also print the full path to the `vault` command we find.

jneo8: You want to search `vault` in user's `$PATH`? I feel it's not precise.

Pjack: +1 to not limit the vault from snap. We can have some default path when we search vault. If it doesn't fit in some case, we can let user assign the path of vault  from the command line parameter

rgildein: Why we are not running this on charm? Will cou snap even access to execute vault?
Sorry I'm maybe out of context here, but this looks as not good idea and I agree that cou should not depend on version of vault installed by user. If it failed for someone, it will be super hard to reproduce, since you have no idea what `vault` was used.
If vault is needed to be used locally, can we pack proper version with our snap?

jneo8: > Why we are not running this on charm?

Because juju log will record the unseal key.

jneo8: I change it to just `vault`, resolved.

rgildein (changes requested):
To my this is really bad idea to add vault as dependency, and we do not even mention it in any part of documentation. Also I was courious if this even will work, since we are doing snap, and apperently not. Here is example how I tested it:
```bash
x1:➜  charmed-openstack-upgrader git:(main) ✗ git diff --exit-code
diff --git a/cou/cli.py b/cou/cli.py
index 0046f58..59aae61 100644
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -16,6 +16,7 @@
 import asyncio
 import logging
 import logging.handlers
+import subprocess
 import sys
 from enum import Enum
 from signal import SIGINT, SIGTERM
@@ -206,6 +207,15 @@ async def _run_command(args: CLIargs) -> None:
 def entrypoint() -> None:
     """Execute 'charmed-openstack-upgrade' command."""
     try:
+        result = subprocess.run(
+            "vault --help".split(),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            check=True,
+        )
+        print(result)
+        return
         args = parse_args(sys.argv[1:])
 
         # disable progress indicator when in quiet mode to suppress its console output                        [0.02s]
x1:➜  charmed-openstack-upgrader git:(main) ✗ which vault
/snap/bin/vault                                                                                                [0.01s]
x1:➜  charmed-openstack-upgrader git:(main) ✗ charmed-openstack-upgrader.cou --help
Unexpected error occurred.
[Errno 2] No such file or directory: 'vault'
  File "/snap/charmed-openstack-upgrader/x1/lib/python3.10/site-packages/cou/cli.py", line 210, in entrypoint
    result = subprocess.run(
  File "/usr/lib/python3.10/subprocess.py", line 503, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1863, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'vault'                                                [1.67s]
```
We cannot access path outside of snap, this is not classic snap.

Why we are not considering adding vault binary as part of snap or even better to use [hvac](https://pypi.org/project/hvac/) package, similar to how zaza use it [here](https://github.com/openstack-charmers/zaza-openstack-tests/blob/e67dca5dc5ee1c785ed1ecc4efb7eed694b1113b/zaza/openstack/charm_tests/vault/utils.py#L40)?

jneo8:
> Why we are not considering adding vault binary as part of snap or even better to use [hvac](https://pypi.org/project/hvac/) package, similar to how zaza use it [here](https://github.com/openstack-charmers/zaza-openstack-tests/blob/e67dca5dc5ee1c785ed1ecc4efb7eed694b1113b/zaza/openstack/charm_tests/vault/utils.py#L40)?

I think include snap binary may be a better choice here, it's all about security issue.

aieri:
> Why we are not considering adding vault binary as part of snap

I don't think this is a good idea: we'd take on the burden of repackaging the right version, updating it, and so on.
Unfortunately this is a limitation of snaps: you can't call other snaps because a snap is its own universe.

hvac could be used, but **only** if it knows how to handle keys securely. This PR is already massive as it currently is; if we have to take on any more complexity I would prefer to delegate unsealing to the operator. COU is good at detecting at what stage of the upgrade a cloud is so it wouldn't seem terrible to me if we bail out and ask the operator to call us again once vault is unsealed.

jneo8:
> > Why we are not considering adding vault binary as part of snap
> 
> I don't think this is a good idea: we'd take on the burden of repackaging the right version, updating it, and so on. Unfortunately this is a limitation of snaps: you can't call other snaps because a snap is its own universe.
> 
> hvac could be used, but **only** if it knows how to handle keys securely. This PR is already massive as it currently is; if we have to take on any more complexity I would prefer to delegate unsealing to the operator. COU is good at detecting at what stage of the upgrade a cloud is so it wouldn't seem terrible to me if we bail out and ask the operator to call us again once vault is unsealed.

Using hvac seems not very hard:

```
import hvac
import getpass

client = hvac.Client(url="https://10.5.3.200:8200")

while True:
    status = client.sys.read_seal_status()
    if status["sealed"]:
        unseal_key = getpass.getpass("Unseal Key (will be hidden):")
        if unseal_key:
            client.sys.submit_unseal_key(key=unseal_key)
            del unseal_key
    else:
        break
```

samuelallan72:
+1 to hvac - the logic looks clean, no external dependencies.

I'd have a question about compatibility (will we need extra tracks of cou depending on vault version on the cloud?).  And we'd need to discuss it in the security section of cou docs.

cou/utils/juju_utils.py
```diff
@@ -710,3 +729,69 @@ async def _wait_for_active_idle() -> None:
...
+            apps=apps,
+            raise_on_blocked=raise_on_blocked,
+            raise_on_error=raise_on_error,
+        )
+
+    async def resolve_all(self) -> None:
+        """Resolve all the units in the model if they are in error status."""
+        apps = await self.get_applications()
+        for app_name, app in apps.items():
+            juju_app = await self._get_application(app_name)
```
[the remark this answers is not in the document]
rgildein: Or maybe we should use `get_status` and if app status is error, we use `juju_unit = await self._get_unit(unit_name)`.

cou/utils/juju_utils.py
```diff
@@ -710,3 +729,69 @@ async def _wait_for_active_idle() -> None:
...
+    async def get_application_status(self, charm_name: str) -> ApplicationStatus:
+        """Get ApplicationStatus by charm name.
+
+        :param charm_name: charm name of application
+        :type charm_name: str
+        :return: ApplicationStatus object
+        :rtype: ApplicationStatus
+        :raises ApplicationNotFound: When application is not found in the model.
+        """
+        status = await self.get_status()
```
rgildein, line 792: Why not using filter parameter in [`get_status`](https://github.com/juju/python-libjuju/blob/6c6d70d483937cec4404116c376b926106e05835/juju/model.py#L2460) function?

cou/utils/openstack.py
```diff
@@ -78,6 +77,8 @@
     "octavia",
 ]
 
+UPGRADE_ORDER = ["vault", "rabbitmq-server"] + UCA_UPGRADE_ORDER
```
rgildein, line 80: Why not put "vault" to the list above?

cou/steps/vault.py
```diff
@@ -0,0 +1,41 @@
...
+        app = await model.get_application_status(charm_name="vault")
+        if app.status.info == "Unit is sealed" and app.status.status == "blocked":
+            raise VaultSealed(
+                "Vault is in sealed, please follow the steps on "
+                "https://charmhub.io/vault to unseal the vault manually before upgrade"
+            )
+    except ApplicationNotFound:
+        logger.warning("Application vault not found, skip")
+    except VaultSealed as err:
+        raise err
```
rgildein, line 40: Why this is here?

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+            timeout=self.wait_timeout,
+            status="blocked",
+            apps=[self.name],
+            # Vault application will go into error status first then blocked status.
+            raise_on_error=False,
+        )
+
+        app_status = await self.model.get_application_status(charm_name="vault")
+        if not app_status.status.info == "Unit is sealed":
+            raise ApplicationError("Vault not in sealed status")
```
rgildein, line 523: Can we update the behavior of `check_vault_status`, so we can use it here? WDYT?

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+        :raises VaultGetStatusFailed: When failed to get sealed status from vault command
+        :raises VaultUnsealFailed: When unseal command failed
+        """
+        progress_indicator.stop()
+
+        for unit_name in self.units:
+            logger.debug("Start unseal %s", unit_name)
+            client = await self._get_vault_client(unit_name)
+            while True:
+                status = client.sys.read_seal_status()  # pylint: disable=no-member
```
rgildein, line 552: Why this pylint ignore is needed?

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+                if status["sealed"]:
+                    unseal_key = getpass.getpass("Unseal Key (will be hidden):")
+                    if unseal_key:
+                        client.sys.submit_unseal_key(key=unseal_key)  # pylint: disable=no-member
+                    del unseal_key
+                else:
+                    break
```
rgildein, line 559: nit:
```suggestion
                if not status["sealed"]:
                    break

                unseal_key = getpass.getpass("Unseal Key (will be hidden):")
                if unseal_key:
                    client.sys.submit_unseal_key(key=unseal_key)  # pylint: disable=no-member
                del unseal_key    
```

cou/apps/base.py
```diff
@@ -68,6 +68,10 @@ class OpenStackApplication(Application):
     packages_to_hold: Optional[list] = field(default=None, init=False)
     wait_timeout: int = field(default=STANDARD_IDLE_TIMEOUT, init=False)
     wait_for_model: bool = field(default=False, init=False)  # waiting only for application itself
+    # Raise error when any unit or app going into "blocked" status
+    raise_on_blocked: bool = field(default=False, init=False)
+    # Raise error when any unit or app going into "error" status
+    raise_on_error: bool = field(default=True, init=True)
```
rgildein, line 74: When this is used? I do not see any app, which is changing default value.

rgildein (changes requested):
Provided some comments.

I'm a little bit afraid that we are using `get_status` too many times in Vault, and that could be time-consuming in a huge model. If it's possible, can we avoid using `get_status`?

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+
+        for unit_name in self.units:
+            logger.debug("Start unseal %s", unit_name)
+            client = await self._get_vault_client(unit_name)
+            while True:
+                status = client.sys.read_seal_status()  # pylint: disable=no-member
+                if status["sealed"]:
+                    unseal_key = getpass.getpass("Unseal Key (will be hidden):")
+                    if unseal_key:
+                        client.sys.submit_unseal_key(key=unseal_key)  # pylint: disable=no-member
```
gabrielcocenza, line 556: I guess it's necessary to check the json response and see if it was 200. If user pass a wrong key, As far as I understood it won't ask again if the key passed is wrong

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+        client = hvac.Client(url=vault_url, verify=cacert_file)
+        return client
+
+    async def _unseal_vault(self) -> None:
+        """Unseal vault on every vault unit.
+
+        :raises VaultGetStatusFailed: When failed to get sealed status from vault command
+        :raises VaultUnsealFailed: When unseal command failed
+        """
+        progress_indicator.stop()
```
gabrielcocenza, line 546: Can you add a comment why this is necessary? Probably it's for the input of the seal keys?

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+        cacert_file = self._get_cacert_file()
+
+        client = hvac.Client(url=vault_url, verify=cacert_file)
+        return client
+
+    async def _unseal_vault(self) -> None:
+        """Unseal vault on every vault unit.
+
+        :raises VaultGetStatusFailed: When failed to get sealed status from vault command
+        :raises VaultUnsealFailed: When unseal command failed
```
gabrielcocenza, line 544: I quickly read the hvac code and I didn't see those exceptions. Where those exceptions are coming from? I though it was just http responses to know if it worked or not.

cou/steps/vault.py
```diff
@@ -0,0 +1,41 @@
...
+async def check_vault_status(model: Model) -> None:
+    """Make sure vault is not in sealed status.
+
+    :param model: juju model to work with
+    :type model: Model
+    :raises VaultSealed: if application in sealed status
+    """
+    try:
+        app = await model.get_application_status(charm_name="vault")
+        if app.status.info == "Unit is sealed" and app.status.status == "blocked":
```
gabrielcocenza, line 32: If during the upgrade process we try to unseal vault, why not make the same here as a pre upgrade step?

gabrielcocenza (changes requested):

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+        """
+        # Use vip as address if vip is being used.
+        if self.config["vip"].get("value"):
+            address = self.config["vip"].get("value")
+        else:
+            juju_unit = await self.model.get_unit(unit_name)
+            address = juju_unit.public_address
+
+        transport = "http"
+        if self.config["ssl-cert"].get("value"):
```
chanchiwai-ray, line 504: If it's possible to have non-TLS enabled vault, does that create a potential security issue? Should we simply let the user to unseal itself?

cou/apps/auxiliary.py
jneo8: Yes

cou/apps/auxiliary.py
```diff
@@ -457,3 +462,146 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+
+        :return: Temporary file path
+        :rtype: str
+        """
+        cacert_b64 = self.config["ssl-ca"].get("vault")
+        cacert_file = None
+        if cacert_b64:
+            with tempfile.NamedTemporaryFile(mode="wb", delete=False) as fp:
+                fp.write(base64.b64decode(cacert_b64))
+                cacert_file = fp.name
```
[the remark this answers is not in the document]
jneo8: https://snapcraft.io/docs/security-sandboxing

> Sets up a private /tmp directory using a per-snap private mount namespace and mounting a per-snap directory on /tmp.

It should have access to `/tmp`, this `/tmp` is per-snap directory. I manually test it with `snap install ... --dangerous` and it has the access.

cou/utils/juju_utils.py
jneo8: The input is charm_name, the filter doesn't support charm name.

cou/apps/auxiliary.py
jneo8: The reason I don't use `check_vault_status` here is because it belong to `steps.vault` module. I can move it to the Vault class and make it become the *classmethod*. Then I have to add error handling here. I feel it's nice to have, not required.

cou/apps/base.py
jneo8: The vault application will get into error status first then blocked. Without this the COU will raise error when vault get into error status.

cou/apps/auxiliary.py
jneo8: Not necessary, because even `submit unseal key` failed. The vault will still be in sealed status, and user need to input again.

cou/apps/auxiliary.py
jneo8: These should be deleted since the exception not exists anymore.

cou/steps/vault.py
jneo8: It's a decision. I take this as a global pre-check before upgrading. We shouldn't  help user to fix it in COU because it's a pre-check. I feel it's not necessary to unseal the vault here. We can change it's behavior later if you think it's required.

cou/apps/auxiliary.py
jneo8: What behavior you want? We stop unseal because TLS is required?

cou/apps/auxiliary.py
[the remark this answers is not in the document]
jneo8: Need to check if there is a way we can know the upgrade step exists or not.

cou/apps/base.py
rgildein: These values are not configured anywhere, the vault has only
```python
 wait_timeout = LONG_IDLE_TIMEOUT
wait_for_model = True
```
I'm not saying that you should not use them at lines 840 and 841, but no charm is using this class variable, so I suggested removing them.

```python
            coro=self.model.wait_for_active_idle(
                timeout=self.wait_timeout,
                apps=apps,
                raise_on_blocked=False,
                raise_on_error=True,
            ),
```

cou/apps/auxiliary.py
rgildein: Like this you created helper function which is used once and here you have lines which are similar to that function. Changing that function to `is_vault_sealed` and you can use it in both places.

cou/utils/juju_utils.py
rgildein: Ok maybe we can get app name from `model` and get status only for single app?

cou/apps/base.py
jneo8: It's used on function `_wait_for_sealed_status`

cou/apps/auxiliary.py
jneo8: It still a little bit different. `check_vault_status` raise if it's in sealed status, `_wait_for_sealed_status` raise if it's not sealed status. The lines we save only two or three.

rgildein: I know, that's why I'm saying if you improved it, you can use it in both places.
e.g.
```pyhon
# here
if await is_vault_sealed(app) is False:
    raise ApplicationError("Vault not in sealed status")
    
# check_vault_status
if await is_vault_sealed(app) is True:
    raise VaultSealed(
        "Vault is in sealed, please follow the steps on "
        "https://charmhub.io/vault to unseal the vault manually before upgrade"
    )
```

Two more comments:
 - Here you are raising ApplicationError, but in check_vault_status you are using VaultSealed
 - Why passing charm name to get_application_status `app_status = await self.model.get_application_status(charm_name="vault")` instead getting status for this app? There should not be two vault apps, but if they are, we should raise an exception for proper one.

cou/apps/base.py
rgildein: There is no app defined like this:

```python
@AppFactory.register_application(["vault"])
class Vault(AuxiliaryApplication):
    """Application for vault."""

    wait_timeout = LONG_IDLE_TIMEOUT
    wait_for_model = True
    raise_on_blocked = True
    raise_on_error = False
```
So those class variables are never used as class variable. That's why I think using constant would be simpler here.

jneo8: The steps are different.

The general steps are:

`upgrade` -> `wait for sealed` -> `unseal` -> `resolve all` -> `wait for model idle`.

We don't want `raise_on_error` on the last step, it only used on `wait for sealed` step.

cou/apps/auxiliary.py
jneo8: > Here you are raising ApplicationError, but in check_vault_status you are using VaultSealed

Yes, or they should both be an `ApplicationError`. `VaultSealed` is more precise and only raise on this pre-check. Or I misunderstand what you want me to change?

> Why passing charm name to get_application_status app_status = await self.model.get_application_status(charm_name="vault") instead getting status for this app? There should not be two vault apps, but if they are, we should raise an exception for proper one.

I see your point, I can change this part.

My point is a `@classmethod` belong to `Vault` o7k application class should not be used in `steps` package. I feel we miss a module that provides helper function for specific juju application, which not belong to either `apps` or `steps` package.

Sorry this make me hesitate because it break my understand about the module levels in this project. And we don't get too much benefit from this practice. I suggest we keep what we have right now.

rgildein: > Yes, or they should both be an ApplicationError. VaultSealed is more precise and only raise on this pre-check. Or I misunderstand what you want me to change?

Both should be `VaultSealed`? Here I'm asking, for reason why one is `ApplicationError` and second `VaultSealed`.

> Please check function check_vault_status, in this case you won't know the application name.

We are in the Vault application, so you can do `self.charm` to get charm name instead of using `"vault"` string, however I think it would be more propriety to get status for application only, so using `self.name`.

> My point is a @classmethod belong to Vault o7k application class should not be used in steps package. I feel we miss a module that provides helper function for specific juju application, which not belong to either apps or steps package.
Sorry this make me hesitate because it break my understand about the module levels in this project. And we don't get too much benefit from this practice. I suggest we keep what we have right now.

You can put `is_vault_sealed` in to utils and also `check_vault_status`, or `check_vault_status` could be in steps/vault.py as you did.  The module structure is not good, I know.

cou/apps/base.py
rgildein: Ok, now I'm confused. 
> We don't want raise_on_error on the last step, it only used on wait for sealed step.

Check line 841, you are using it there.

I mean the 2 class variables you defined in lines 72 and 74 are used only on lines 840 and 841, nowhere else.
So my suggestion is to remove them and use simple constants
e.g. 
```python
           coro=self.model.wait_for_active_idle(
                timeout=self.wait_timeout,
                apps=apps,
                raise_on_blocked=RAISE_ON_BLOCKED,
                raise_on_error=RAISE_ON_ERROR,
            ),
```
or even better, completely remove them, since those are default values (default behavior) for wait_for_active_idle function.
```python
           coro=self.model.wait_for_active_idle(timeout=self.wait_timeout, apps=apps),
```

cou/apps/auxiliary.py
jneo8: > Both should be VaultSealed? Here I'm asking, for reason why one is ApplicationError and second VaultSealed.

The VaultSealed raise because we don't support to upgrade cloud which vault in sealed status. Second case is that we want to reach the sealed status. It's wired we raise same error on the opposite situation.

> We are in the Vault application, so you can do self.charm to get charm name instead of using "vault" string, however I think it would be more propriety to get status for application only, so using self.name.

The `check_vault_status` in `steps.vault` module, it's not in the application.

cou/apps/base.py
jneo8: I see your point.

cou/steps/vault.py
gabrielcocenza: I'm ok with this decision :+1:

cou/apps/auxiliary.py
```diff
@@ -457,3 +463,156 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+                status = client.sys.read_seal_status()
+                if not status["sealed"]:
+                    break
+                unseal_key = getpass.getpass("Unseal Key (will be hidden):")
+                if unseal_key:
+                    client.sys.submit_unseal_key(key=unseal_key)
+                del unseal_key
+
+            # Delete temporary ca file
+            if cacert_file:
```
gabrielcocenza, line 560: nit: What is the difference of generating for every unit the `cacert_file` and delete instead of generating once and delete at the end?

cou/utils/juju_utils.py
```diff
@@ -329,6 +329,17 @@ async def _get_supported_apps(self) -> list[str]:
             name for name, app in model.applications.items() if is_charm_supported(app.charm_name)
         ]
 
+    async def get_unit(self, name: str) -> JujuUnit:
```
gabrielcocenza, line 332: Wouldn't be easier and cleaner transform `_get_unit` into `get_unit`?

cou/utils/juju_utils.py
```diff
@@ -649,20 +660,24 @@ async def upgrade_charm(
...
 
-    async def wait_for_active_idle(
+    async def wait_for_idle(
         self,
         timeout: int,
+        status: str = "active",
         idle_period: int = DEFAULT_MODEL_IDLE_PERIOD,
         apps: Optional[list[str]] = None,
         raise_on_blocked: bool = False,
+        raise_on_error: bool = True,
```
gabrielcocenza, line 670: This PR is way too long. I wish this kind of new functionalities on juju_utils were splitted into different PR

cou/utils/juju_utils.py
```diff
@@ -710,3 +729,67 @@ async def _wait_for_active_idle() -> None:
...
+        :type raise_on_error: bool
+        """
+        await self.wait_for_idle(
+            timeout=timeout,
+            status="active",
+            idle_period=idle_period,
+            apps=apps,
+            raise_on_blocked=raise_on_blocked,
+            raise_on_error=raise_on_error,
+        )
```
gabrielcocenza, line 768: I think this is too verbose to just call `wait_for_idle` with the default values. I didn't see a difference to call it directly

gabrielcocenza (changes requested):

cou/utils/juju_utils.py
jneo8: Because `wait_for_active_idle` exists already, I shouldn't delete it on a feature PR.
I try not to refactor and add feature at the same time.

cou/utils/juju_utils.py
jneo8: I shouldn't do that because it's a breaking refactor change. It shouldn't happen in this PR.

.github/workflows/check.yaml
```diff
@@ -62,6 +62,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install tox
+        # vault is required to pass the command exists check
+        sudo snap install vault
```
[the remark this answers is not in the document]
samuelallan72: Agreed, we don't need it; we're using the hvac lib instead.

cou/utils/juju_utils.py
```diff
@@ -710,3 +729,67 @@ async def _wait_for_active_idle() -> None:
...
+    async def resolve_all(self) -> None:
+        """Resolve all the units in the model if they are in error status."""
+        model = await self._get_model()
+        for _, juju_app in model.applications.items():
+            if juju_app.status != "error":
+                continue
+            for unit in juju_app.units:
+                if unit.workload_status != "error":
+                    continue
+                await unit.resolved(retry=True)
```
samuelallan72, line 779: Could this be simplified to:

```suggestion
            for unit in juju_app.units:
                if unit.workload_status == "error":
                    await unit.resolved(retry=True)
```

samuelallan72 (changes requested):
This is a large PR and hard to review... :/  I left only a couple of comments, but I agree with comments from the other reviewers.

cou/apps/auxiliary.py
rgildein: > The VaultSealed raise because we don't support to upgrade cloud which vault in sealed status. Second case is that we want to reach the sealed status. It's weird we raise same error on the opposite situation.

Ok, make sense.

> The check_vault_status in steps.vault module, it's not in the application.

I'm speaking about this line:
`app_status = await self.model.get_application_status(charm_name="vault")`

This will get any vault, what I'm suggesting is to use only current vault app status.

cou/apps/base.py
rgildein: You are adding it here, so it's related.

cou/apps/auxiliary.py
jneo8: > This will get any vault, what I'm suggesting is to use only current vault app status.

Any vault should be fine, I don't think we will have a case that will have multiple vault applications.

Issue #488, opened 2024-07-09
--purge is broken


samuelallan72:
`cou upgrade` with `--purge` option fails:

```
$ cou upgrade --no-backup --purge
...
Would you like to start the upgrade? Continue (y/N): y
Running cloud upgrade...
Verify that all OpenStack applications are in idle state ✔
Archive old database data on nova-cloud-controller ✔
Purge all data from shadow tables on nova-cloud-controller |2024-07-09 08:43:18 [ERROR] Unexpected error occurred.
2024-07-09 08:43:18 [ERROR] 'results'
```

It appears that this line still uses the juju2 api for getting the action output:

https://github.com/canonical/charmed-openstack-upgrader/blob/7b13f8c1d1ff680a9d634107b62d98c5d9b767f8/cou/steps/nova_cloud_controller.py#L89-L89

It must have been missed after https://github.com/canonical/charmed-openstack-upgrader/pull/444 was updated with main after https://github.com/canonical/charmed-openstack-upgrader/pull/451 was merged.

PR #451, opened 2024-06-25
Port to Juju 3

samuelallan72:
Upgrade python libjuju to version 3.

This drops support for working with models deployed with juju 2.x,
but adds support for models deployed with juju 3.x.

This is a rebase/rework of https://github.com/canonical/charmed-openstack-upgrader/pull/410

PR #489, opened 2024-07-09
Fix purge step

samuelallan72:
Getting the action output was still using the juju2 api method.

Updating it must have been missed in #444 (add the purge step)
after it was updated with main after #451 (port to juju3) was merged.

Fixes: #488


samuelallan72:
superceded by #490

PR #493, opened 2024-07-12
Add how-to query nova db soft-deleted instances

samuelallan72:
These are useful commands for debugging or verifying the archive and purge steps.
We may like them kept around for reference later.

### DIFF
--- a/developing/how-to-query-nova-db-soft-deleted-instances.md
+++ b/developing/how-to-query-nova-db-soft-deleted-instances.md
@@ -1,0 +1,70 @@
+# How to query nova db for soft deleted instances
+
+This is useful for verifying or debugging archive and purge steps.
+Instructions below detail how to connect to the nova database and query common information for soft deleted instances.
+
+## Credentials
+
+First you'll need the MySQL user password:
+
+```sh
+PASS="$(juju exec --unit mysql-innodb-cluster/leader -- leader-get cluster-password)"
+echo "$PASS"
+```
+
+## SQL reference
+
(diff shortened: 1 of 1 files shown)


gabrielcocenza (commented):
That is a nice piece of information. However, IMHO this does not make part of COU development. Maybe moving this to a section in the docs/ how-to archive-old-data.rst would make more sense?

Pjack (approved):

developing/how-to-query-nova-db-soft-deleted-instances.md
```diff
@@ -0,0 +1,70 @@
+# How to query nova db for soft deleted instances
+
+This is useful for verifying or debugging archive and purge steps.
```
chanchiwai-ray, line 3: (non-blocking) Following from @gabrielcocenza 's comment, it might be good to put this line to `docs/ how-to archive-old-data.rst ` and put a link to this doc `developing/how-to-query-nova-db-soft-deleted-instances.md` if the user is interested in verifying the archive and purge

chanchiwai-ray (approved):

jneo8 (approved):

developing/how-to-query-nova-db-soft-deleted-instances.md
samuelallan72: Agreed, it could be better elsewhere, but I'm not sure where is best.  We can refactor later I guess; at least we have it recorded somewhere for now. :)

PR #497, opened 2024-07-17
Add subordinate units information for Application.

chanchiwai-ray:
This augmented the `Application` class with `subordinate_units` information so that you can get subordinate units easily for an application.

For example, If the model contains the following:

```
App
nova-compute
ovn-chassis

Unit
nova-compute/0
    ovn-chassis/0
nova-compute/1
    ovn-chassis/2
```

**Before the PR:**

```python
Application.units = {"nova-compute/0": Unit(...), nova-compute/1": Unit(...)}  # for nova-compute
Application.units = {}` for `ovn-chassis  # for ovn-chassis
```

**After the PR:**

```python
Application.units = {"nova-compute/0": Unit(...), nova-compute/1": Unit(...)}  # for nova-compute
Application.subordinate_units = []  # for nova-compute
Application.units = {}` for `ovn-chassis 
Application.subordinate_units = [SubordinateUnit("ovn-chassis/0"), SubordinateUnit("ovn-chassis/1")] # for ovn-chassis
```

see `tests/unit/utils/test_juju_utils.py` for updated test case

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -175,7 +175,8 @@ class Application:
     origin: str
     series: str
     subordinate_to: list[str]
-    units: dict[str, Unit]
+    units: dict[str, Unit]  # subordainte apps does not have units, use subodinate_units instead
+    subordinate_units: list[SubordinateUnit]  # this is the units for subordinate apps
     workload_version: str
     actions: dict[str, str] = field(default_factory=lambda: {}, compare=False)
 
@@ -367,6 +368,10 @@ class Model:
         #                 information the status than from objects. e.g. workload_version for unit
         full_status = await self.get_status()
         machines = await self._get_machines()
+        subordinate_units = {
+            SubordinateUnit(name, model.applications[name.split("/")[0]].charm_name)
+            for name in model.subordinate_units
+        }
 
         return {
             app: Application(
@@ -398,6 +403,7 @@ class Model:
                     )
                     for name, unit in status.units.items()
                 },
+                subordinate_units=[su for su in subordinate_units if su.name.split("/")[0] == app],
                 workload_version=status.workload_version,
                 actions=await model.applications[app].get_actions(),
             )
--- a/tests/unit/apps/test_auxiliary_subordinate.py
+++ b/tests/unit/apps/test_auxiliary_subordinate.py
@@ -22,6 +22,7 @@ from cou.apps.auxiliary_subordinate import (
 )
 from cou.exceptions import ApplicationError, HaltUpgradePlanGeneration
 from cou.steps import ApplicationUpgradePlan, PreUpgradeStep, UpgradeStep
+from cou.utils.juju_utils import SubordinateUnit
 from cou.utils.openstack import OpenStackRelease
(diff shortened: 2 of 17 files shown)


chanchiwai-ray:
Note: only 8 lines of 300+ lines are actual code changes, others are changes to the unit tests

gabrielcocenza (commented):
If I understood correctly, this this new field `subordinate_units` is going to show only when the application itself is subordinate, right? If that is the case, please add a comment about this in the PR description and also in the code.

Would the impact be huge if we populate the field units with `SubordinateUnit` when the application is subordinate?

chanchiwai-ray:
> Would the impact be huge if we populate the field units with `SubordinateUnit` when the application is subordinate?

Yes, I think it will be. It's because in COU there are places where we iterate over the units using `self.app.units` and do something with it. If we add subordinate units itself to `units`, then the `self.app.units != []` and it will try to running the code inside the loop. Often, the loop inside `self.app.units` assumes it's `Unit` and do something like get machine id etc.

Moreover, I think it's better not to add them into `units` because that will deviate from the output from libjuju itself, making it harder to understand what's going on

tests/unit/apps/test_auxiliary_subordinate.py
```diff
@@ -101,6 +104,7 @@ def test_ovn_subordinate(model):
         origin="ch",
         series="focal",
         subordinate_to=["nova-compute"],
+        subordinate_units=[SubordinateUnit("ovn-chassisr/0", "ovn-chassis")],
```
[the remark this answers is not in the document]
chanchiwai-ray: Fixed

cou/utils/juju_utils.py
```diff
@@ -175,7 +175,8 @@ class Application:
     origin: str
     series: str
     subordinate_to: list[str]
-    units: dict[str, Unit]
+    units: dict[str, Unit]  # subordainte apps does not have units, use subodinate_units instead
+    subordinate_units: list[SubordinateUnit]  # this is the units for subordinate apps
```
jneo8, line 179: **suggestion:** 

The SubordinateUnit is already exists in Unit.subordinates.

Is that possible we create a property `subordinate_units` on `Application` class instead of adding a new field?

Also SubordinateUnit in Unit can show the subordinate relationship between the units.

jneo8 (changes requested):
Want to discuss the possible solution.

Pjack (changes requested):
LGTM

cou/utils/juju_utils.py
chanchiwai-ray: Sorry, I am not sure what do you mean. Can you explain more?

`subordinate_units` is not the same as `Unit.subordinates`

jneo8: Yes, they are different. But you can generate the `subordinate_units` by iterator over units to get them.

chanchiwai-ray: Maybe, but we need this information in the `Application` class and in `get_applications` method

Pjack (approved):

jneo8 (approved):

Issue #466, opened 2024-06-28
[cinder / cinder-ceph] upgrading `cinder` will cause its subordinate `cinder-ceph` to failed on relation changed


chanchiwai-ray:
```
Wait for up to 300s for app 'cinder' to reach the idle state ✖
2024-06-24 04:45:08 [ERROR] Timed out waiting for model:
  cinder/0 [executing] maintenance: Running openstack upgrade
2024-06-24 04:45:08 [ERROR] See the known issues at https://canonical-charmed-openstack-upgrader.readthedocs-hosted.com/en/stable/reference/known-issues/
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed Traceback (most recent call last):
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed   File "/var/lib/juju/agents/unit-cinder-ceph-0/charm/charmhelpers/core/strutils.py", line 94, in __init__
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed     self.index = self._list.index(item)
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed ValueError: tuple.index(x): x not in tuple
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed During handling of the above exception, another exception occurred:
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed Traceback (most recent call last):
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed   File "/var/lib/juju/agents/unit-cinder-ceph-0/charm/hooks/storage-backend-relation-changed", line 469, in <module>
unit-cinder-ceph-0: 05:24:20 WARNING unit.cinder-ceph/0.storage-backend-relation-changed     hooks.execute(sys.argv)
[...]

PR #498, opened 2024-07-18
fix: Change upgrade order of control plane

jneo8:
- Update subordinate applications first then principal

Fix: #466


cou/steps/plan.py
```diff
@@ -534,7 +534,9 @@ def _generate_control_plane_plan(
     )
 
     logger.debug("Generation of the control plane upgrade plan complete")
-    control_plane_upgrade_plan = [principal_upgrade_plan, subordinate_upgrade_plan]
+    # Note: relate to issue #466
+    # The subordinate application should be upgrade first before principle
```
samuelallan72, line 538: Something like this may make it clearer:

```suggestion
    # Subordinate charms should be upgraded before principal charms
    # to avoid the subordinates being on an unsupported platform after principal charms are upgraded.
    # See https://github.com/canonical/charmed-openstack-upgrader/issues/466#issuecomment-2209991680 for more information.
```

samuelallan72 (changes requested):
Just a docs request, thanks looks pretty good.  Nice that this is a simple change. :)

Pjack (approved):
LGTM

chanchiwai-ray (approved):

gabrielcocenza (approved):
Glad to see that the changes were not painful for the project.

I would just add in the description a link to the PR that is changing the `data-plane` for future travelers navigate easily into the changes

samuelallan72 (approved):

chanchiwai-ray (approved):

PR #499, opened 2024-07-18
Workaround ceilometer-agent-compute not running

samuelallan72:
This is a bug in nova-compute and/or ceilometer-agent
( https://bugs.launchpad.net/charm-ceilometer-agent/+bug/1947585 ),
where the nova-compute `resume` action can sometimes fail.
The workaround added here is to restart ceilometer-agent-compute
if it fails for this reason.

Fixes: #427

### DIFF
--- a/cou/apps/core.py
+++ b/cou/apps/core.py
@@ -18,9 +18,9 @@ from typing import Optional
 
 from cou.apps.base import LONG_IDLE_TIMEOUT, OpenStackApplication
 from cou.apps.factory import AppFactory
-from cou.exceptions import ApplicationNotSupported
+from cou.exceptions import ActionFailed, ApplicationNotSupported
 from cou.steps import PostUpgradeStep, PreUpgradeStep, UnitUpgradeStep, UpgradeStep
-from cou.utils.juju_utils import Unit
+from cou.utils.juju_utils import Model, Unit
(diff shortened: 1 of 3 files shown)


cou/apps/core.py
```diff
@@ -221,6 +221,61 @@ def _get_disable_scheduler_step(self, units: Optional[list[Unit]]) -> list[PreUp
...
+        logger.debug("Resume failed because ceilometer-agent-compute not running.")
+        logger.debug("Restarting ceilometer-agent-compute on %s", unit.name)
+        await model.run_on_unit(unit.name, "sudo systemctl restart ceilometer-agent-compute")
+
+        # Update status manually, otherwise nova-compute and ceilometer-agent
+        # will be blocked until next update-status hook.
+        await model.update_status(unit.name)
+        for subordinate in unit.subordinates:
+            if subordinate.charm == "ceilometer-agent":
+                await model.update_status(subordinate.name)
```
samuelallan72, line 273: After this, everything should be active/idle again.  I don't think we need to re-run the `resume` action?

jneo8: should be

jneo8 (approved):

chanchiwai-ray (approved):
Looks reasonable, let's see if it fixes the issue

Issue #468, opened 2024-06-28
[nova-compute / ovn-chassis] Upgrade 'ovn-chassis' from '22.03/stable' to '22.09/stable' cause `RuntimeError`


chanchiwai-ray:
When COU refresh ovn-chassis from '22.03/stable' to '22.09/stable' during Jammy/Yoga -> Jammy/Zed upgrade, the ovn-chassis charm goes into error state with `RuntimeError`:

```
unit-ovn-chassis-0: 02:41:51 ERROR unit.ovn-chassis/0.juju-log nova-compute:43: Hook error:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms/reactive/__init__.py", line 74, in main
    bus.dispatch(restricted=restricted_mode)
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms/reactive/bus.py", line 390, in dispatch
    _invoke(other_handlers)
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms/reactive/bus.py", line 359, in _invoke
    handler.invoke()
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms/reactive/bus.py", line 181, in invoke
    self._action(*args)
  File "/var/lib/juju/agents/unit-ovn-chassis-0/charm/reactive/layer_openstack.py", line 134, in default_request_certificates
    with charm.provide_charm_instance() as instance:
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms_openstack/charm/core.py", line 99, in __enter__
    return BaseOpenStackCharm.singleton
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms_openstack/charm/core.py", line 352, in singleton
    _singleton = get_charm_instance(release=release,
  File "/var/lib/juju/agents/unit-ovn-chassis-0/.venv/lib/python3.10/site-packages/charms_openstack/charm/core.py", line 165, in get_charm_instance
    return (_get_charm_instance_function or default_get_charm_instance)(
[...]

PR #501, opened 2024-07-18
Fix: Change upgrade order of data plane

jneo8:
fix: Change upgrade order of data plane
    
- Move subordinate applications before hypervisor and principle
    
fix: #468
Follow: #498


cou/steps/plan.py
```diff
@@ -596,10 +611,10 @@ async def _generate_data_plane_hypervisors_plan(
     return hypervisor_plan
 
 
-def _generate_data_plane_remaining_plan(
+def _generate_data_plane_principle_remaining_plan(
```
gabrielcocenza, line 614: typo:
```suggestion
def _generate_data_plane_principal_remaining_plan(
```

cou/steps/plan.py
```diff
@@ -129,13 +129,26 @@ async def _generate_data_plane_plan(
     hypervisor_apps, non_hypervisors_apps = _separate_hypervisors_apps(
         analysis_result.apps_data_plane
     )
+    plans = []
+    if args.upgrade_group != HYPERVISORS:
+        plans.extend(
+            _generate_data_plane_subordinate_remaining_plan(
+                target, non_hypervisors_apps, args.force
+            )
+        )
```
gabrielcocenza, line 138: I guess this is not ideal anymore. We'll always need to upgrade subordinates before the hypervisors. This can be moved to the code bellow.

cou/steps/plan.py
```diff
@@ -534,7 +547,9 @@ def _generate_control_plane_plan(
     )
 
     logger.debug("Generation of the control plane upgrade plan complete")
-    control_plane_upgrade_plan = [principal_upgrade_plan, subordinate_upgrade_plan]
+    # Note: relate to issue #466
+    # The subordinate application should be upgrade first before principle
+    control_plane_upgrade_plan = [subordinate_upgrade_plan, principal_upgrade_plan]
```
gabrielcocenza, line 552: Maybe not scope of this PR?

gabrielcocenza (changes requested):

cou/steps/plan.py
```diff
@@ -129,13 +129,26 @@ async def _generate_data_plane_plan(
...
+                target, analysis_result, args, hypervisor_apps
+            )
+        ]
+    )
 
     if args.upgrade_group != HYPERVISORS:
-        plans.extend(_generate_data_plane_remaining_plan(target, non_hypervisors_apps, args.force))
+        plans.extend(
+            _generate_data_plane_principle_remaining_plan(target, non_hypervisors_apps, args.force)
+        )
```
chanchiwai-ray, line 151: I think it should be like

```python3
plans = []
(
  non_hypervisor_subordinate, 
  non_hypervisor_principle, 
  hypervisors_subordinate, 
  hypervisors_principle
)  = _separate_hypervisors_apps(analysis_result.apps_data_plane)

if args.upgrade_group != HYPERVISORS:
    plans.extend(
        _generate_data_plane_non_hypervisor_subordinate_plan(
            target, non_hypervisor_subordinate, args.force
        )
    )
    plans.extend(
        _generate_data_plane_non_hypervisor_principle_plan(target, non_hypervisor_principle, args.force)
    )
plans.extend(
    [
        await _generate_data_plane_hypervisors_subordinate_plan(
            target, analysis_result, args, hypervisors_subordinate
        ),
        await _generate_data_plane_hypervisors_principle_plan(
            target, analysis_result, args, hypervisors_principle
        )
    ]
)
```

chanchiwai-ray (changes requested):
I recommend this **IF** we still want to group hypervisor charm into the same group AND upgrade subordinate first, then principle. Otherwise, feel free to resolve it, and request review

cou/steps/plan.py
jneo8: Thanks Ray.

We have to be very careful about this one. Maybe we should discuss every applications upgrade strategy here. 

For example, ceph-osd is a data plan principle. If you move `ceph-osd` to `non_hypervisor_principle`, it will break the upgrade order.`ceph-osd` should upgrade after `nova-compute`.

Most of the applications doesn't support single unit upgrade as I understand. Which means on hypervisor upgrading strategy, we will break the recommend upgrade order. For example you upgrade nova then ceph-osd on machine 1, but ceph-osd will be refresh at the same time on machine 2 which nova still no upgraded yet.

Maybe we should define the whole plan first? Like below is a example:

```
non_hypervisor_subordinate
non_hypervisor_principle_before_nova
hypervisor_unit_by_unit_upgrade
  unit_upgrade_supported_principle, before nova
  nova_upgrade
  unit_upgrade_supported_principle, after nova
non_hypervisor_principle_after_nova
```

cc: @gabrielcocenza

gabrielcocenza: As mentioned by James, non-hypervisors should upgrade `after nova-compute` because of ceph-osd. Note that even if ceph-osd is deployed in the same nova-compute machines we consider as a non-hypervisor because it does not have the ability to upgrade unit-by-unit.

We are already upgrading applications by sequence, meaning that if cinder is deployed with nova-compute, we will upgrade cinder first and then nova-compute. Note that you need to disable nova-compute scheduler before upgrading cinder to avoid new instances in a hypervisor while the upgrade is happening.

In summary I do believe that the current approach is solid, the only thing is that we need to upgrade hypervisors-subordinates first.

- hypervisors_subordinate
- hypervisors(including other principles E.g: cinder) respecting the current srtucture
- -non_hypervisors_subordinates
- non_hypervisors

As far as I understand is just a question of separating subordinates from hypervisors and non-hypervisors and respect the sequence. Note:
- subordinates of other principals like cinder will be considered as hypervisors_subordinate
- nova-compute subordinates are considered as hypervisors_subordinate
- ceph-osd subordinates are considered as non_hypervisors_subordinates
- if ceph-osd and nova-compute share a same subordinate we can consider as hypervisors_subordinate

jneo8:
Move implementation to #518

PR #502, opened 2024-07-18
Support enable auto restarts false ovn chassis

chanchiwai-ray:
Add support for `enable-auto-restarts=False` in ovn-chassis

Depends on https://github.com/canonical/charmed-openstack-upgrader/pull/504

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -14,7 +14,7 @@
 """Auxiliary application class."""
 import abc
 import logging
-from typing import Optional
+from typing import Optional, Sequence
 
 from packaging.version import Version
 
@@ -28,7 +28,7 @@ from cou.steps import (
     UnitUpgradeStep,
 )
 from cou.utils.app_utils import set_require_osd_release_option
-from cou.utils.juju_utils import Unit
+from cou.utils.juju_utils import SubordinateUnit, Unit
 from cou.utils.openstack import (
     OPENSTACK_TO_TRACK_MAPPING,
     TRACK_TO_OPENSTACK_MAPPING,
@@ -175,7 +175,7 @@ class AuxiliaryApplication(OpenStackApplication):
         )
 
     def get_run_deferred_hooks_and_restart_pre_upgrade_step(
-        self, units: Optional[list[Unit]]
+        self, units: Sequence[Unit | SubordinateUnit]
     ) -> list[PreUpgradeStep]:
         """Get the steps for run deferred hook and restart services for before upgrade.
 
@@ -204,7 +204,7 @@ class AuxiliaryApplication(OpenStackApplication):
                         unit.name, "run-deferred-hooks", raise_on_failure=True
                     ),
                 )
-                for unit in units or self.units.values()
+                for unit in units
             ]
         )
         wait_step = PreUpgradeStep(
@@ -221,7 +221,7 @@ class AuxiliaryApplication(OpenStackApplication):
         ]
 
     def get_run_deferred_hooks_and_restart_post_upgrade_step(
-        self, units: Optional[list[Unit]]
+        self, units: Sequence[Unit | SubordinateUnit]
     ) -> list[PostUpgradeStep]:
         """Get the step for run deferred hook and restart services for after upgrade.
 
@@ -258,7 +258,7 @@ class AuxiliaryApplication(OpenStackApplication):
                         unit.name, "run-deferred-hooks", raise_on_failure=True
                     ),
                 )
-                for unit in units or self.units.values()
+                for unit in units
             ]
         )
         return [
@@ -294,7 +294,9 @@ class RabbitMQServer(AuxiliaryApplication):
         """
         steps = super().pre_upgrade_steps(target, units)
         if self.config.get("enable-auto-restarts", {}).get("value") is False:
-            steps.extend(self.get_run_deferred_hooks_and_restart_pre_upgrade_step(units))
+            steps.extend(
+                self.get_run_deferred_hooks_and_restart_pre_upgrade_step(list(self.units.values()))
+            )
         return steps
 
     def post_upgrade_steps(
@@ -313,7 +315,11 @@ class RabbitMQServer(AuxiliaryApplication):
         """
         steps = []
         if self.config.get("enable-auto-restarts", {}).get("value") is False:
-            steps.extend(self.get_run_deferred_hooks_and_restart_post_upgrade_step(units))
+            steps.extend(
+                self.get_run_deferred_hooks_and_restart_post_upgrade_step(
+                    list(self.units.values())
+                )
+            )
         steps.extend(super().post_upgrade_steps(target, units))
         return steps
 
--- a/cou/apps/auxiliary_subordinate.py
+++ b/cou/apps/auxiliary_subordinate.py
@@ -12,9 +12,20 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 """Auxiliary subordinate application class."""
+
+# We have to disable duplicate code because some code are intended to add to
+# the child class, it cannot be added to parent class to reuse the code.
+#
+# pylint: disable=duplicate-code
+
+from typing import Optional
+
 from cou.apps.auxiliary import OVN, AuxiliaryApplication
 from cou.apps.factory import AppFactory
 from cou.apps.subordinate import SubordinateApplication
+from cou.steps import PostUpgradeStep, PreUpgradeStep
+from cou.utils.juju_utils import Unit
+from cou.utils.openstack import OpenStackRelease
 
 
 @AppFactory.register_application(["mysql-router", "ceph-dashboard"])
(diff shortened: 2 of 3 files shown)


chanchiwai-ray:
For the refactoring: although the plan looks different (different way of adding steps in COU) but the behaviors is the same.

I've tested that part with

```
juju deploy rabbitmq-server --channel 3.8/stable --base ubuntu@20.04 -n3 --config source=distro
juju config rabbitmq-server enable-auto-restarts=False
cou upgrade --no-backup --auto-approve --no-archive
```

samuelallan72:
@chanchiwai-ray 

> * Restart how to handle enable-auto-restarts=False in rabbitmq server
> * Add support for enable-auto-restarts=False in ovn-chassis

This sounds like we could do this in two PRs: first to refactor how we handle rabbitmq server in preparation for adding support for ovn-chassis, the second to add the new feature.

chanchiwai-ray:
> @chanchiwai-ray
> 
> > * Restart how to handle enable-auto-restarts=False in rabbitmq server
> > * Add support for enable-auto-restarts=False in ovn-chassis
> 
> This sounds like we could do this in two PRs: first to refactor how we handle rabbitmq server in preparation for adding support for ovn-chassis, the second to add the new feature.

I can split it into two PRs, I have the refactoring branch there. But I decided to merge them in to one in the last second, since it's kind of related

cou/apps/auxiliary.py
```diff
@@ -183,6 +183,14 @@ class RabbitMQServer(AuxiliaryApplication):
     # COU changes to 3.9 if the channel is set to 3.8
     multiple_channels = True
 
+    def _check_auto_restarts(self) -> None:
+        """No-op, skip check auto restarts option.
+
+        This method override the parent class's `_check_auto_restarts()` method
+        because the parent class's will raise an `ApplicationError` if
+        `enable-auto-restarts` is `True`.
+        """
```
samuelallan72, line 192: The pattern of inheritance we have in cou is not nice - we shouldn't have to add no-op override methods. :confused: 

Either way, could you move this back to its original location to avoid the diff?

chanchiwai-ray:
I move refactoring here: https://github.com/canonical/charmed-openstack-upgrader/pull/504

samuelallan72 (changes requested):

cou/apps/base.py
```diff
@@ -607,6 +609,82 @@ def _get_charmhub_migration_step(self, target: OpenStackRelease) -> PreUpgradeSt
             ),
         )
 
+    def get_run_deferred_hooks_and_restart_pre_upgrade_step(self) -> list[PreUpgradeStep]:
```
jneo8, line 612: **suggestion:** 

If `get_run_deferred_hooks_and_restart_pre_upgrade_step` and `get_run_deferred_hooks_and_restart_post_upgrade_step` will only be used by anxiliary applications. Why we don't put this function in `AuxiliaryApplication` class?

jneo8 (commented):
One question.

cou/apps/base.py
chanchiwai-ray: see https://github.com/canonical/charmed-openstack-upgrader/pull/504

cou/apps/auxiliary.py
chanchiwai-ray: Resolving: addressed in the previous PR

cou/apps/base.py
chanchiwai-ray: Resolving: addressed in the previous PR

cou/apps/auxiliary.py
```diff
@@ -175,7 +175,7 @@ def _need_current_channel_refresh(self, target: OpenStackRelease) -> bool:
         )
 
     def get_run_deferred_hooks_and_restart_pre_upgrade_step(
-        self, units: Optional[list[Unit]]
+        self, units: Sequence[Unit | SubordinateUnit]
     ) -> list[PreUpgradeStep]:
         """Get the steps for run deferred hook and restart services for before upgrade.
```
gabrielcocenza, line 181: nit: Update docstring type for units

cou/apps/auxiliary_subordinate.py
```diff
@@ -33,6 +44,59 @@ def _check_ovn_support(self) -> None:
         """
         OVNSubordinate._validate_ovn_support(self.workload_version)
 
+    def _check_auto_restarts(self) -> None:
```
gabrielcocenza, line 47: add a TODO to remove this once all charms supports auto restarts to False

cou/apps/auxiliary_subordinate.py
```diff
@@ -33,6 +44,59 @@ def _check_ovn_support(self) -> None:
...
+        :return: List of post upgrade steps.
+        :rtype: list[PostUpgradeStep]
+        """
+        steps = []
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
+            steps.extend(
+                self.get_run_deferred_hooks_and_restart_post_upgrade_step(self.subordinate_units)
+            )
+        steps.extend(super().post_upgrade_steps(target, units))
+        return steps
```
gabrielcocenza, line 98: I think this is not enough for subordinates.

Imagine the following scenario:
ovn-chassis is related with nova-compute and we start with the subordinate:

1- You do the upgrade steps for ovn-chasis. 
2 - You upgrade nova-compute (Here is when the packages update and ovn might request to restart)
3 - You 'll leave the cloud without restarting the ovn-chassis

This should be called as a post upgrade step of the principal charm and we need to receive specific ovn-chassis units, otherwise you can create an outage of data-plane.

cou/apps/auxiliary_subordinate.py
```diff
@@ -33,6 +44,59 @@ def _check_ovn_support(self) -> None:
...
+        :rtype: list[PreUpgradeStep]
+        """
+        # OVN subordinate charm does not have upgrade_step; the "upgrade" is
+        # done during the pre-upgrade step, so we need to put the run deferred
+        # hook step before pre-upgrade step.
+        steps = []
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
+            steps.extend(
+                self.get_run_deferred_hooks_and_restart_pre_upgrade_step(self.subordinate_units)
+            )
```
gabrielcocenza, line 74: You need to receive specific units from a higher level. Upgrading all units at once here you'll create a possible data-plane network outage if all nova-compute units restarts the ovn services at the same time

gabrielcocenza (changes requested):

cou/apps/auxiliary_subordinate.py
aieri: @chanchiwai-ray the deferred restarts mechanism was introduced to avoid restarting all units at once and allow operators to do a rolling restart. This is to avoid generating downtime when quorum in the cluster is lost.  When an operator runs manual upgrades, they will run deferred hooks / service restarts on one unit, wait until the cluster is back to full redundancy, and then move on to the next one.
Are we running deferred hooks / service restarts all at once in parallel for all services we are supporting? If so we need to revisit those codepaths.
Also note that for ovn-chassis things are a _little_ more relaxed: doing one at a time is fine, but I think it would also be fine to restart all ovn-chassis belonging to empty hypervisors (because the blast radius of those restarts is all the VMs running on that specific host). If we're dealing with ovn-chassis as a subordinate for anything other than nova-compute though, I suspect a strict one at a time strategy would be the only safe one.

aieri: ok thinking about this some more...
In order not to go crazy with overcomplicating things about rolling upgrades, we should think in terms of classes of applications:
* pure control plane apps: downtime doesn't matter → restart whatever you want without a care for quorum
* empty hypervisors or hypervisors that the user told us to go ahead and upgrade anyway: downtime doesn't matter → quorum isn't a thing here, and you can restart whatever and whenever
* rabbitmq: it's a control plane app, sort of. If it stays down for too long neutron won't be happy and data plane will be affected → handle deferred restarts carefully as to never lose quorum
* ovn-chassis as a subordinate of octavia → if restarting it only causes downtime in the control plane (i.e. pushing updates to amphorae) then there is no need for rolling restarts, if on the other hand it causes LBs to lose packets we should strive to maintain quorum

chanchiwai-ray: > Are we running deferred hooks / service restarts all at once in parallel for all services we are supporting? If so we need to revisit those codepaths.

No, the substeps are not run in parallel, it's run unit-by-unit until the action completes.

- Refresh the ovn-chassis without restarting the service
  - restart ovn-chassis unit 1 until it's done
  - restart ovn-chassis unit 2 until it's done
  - ...

chanchiwai-ray:
Closing it because the support of `enable-auto-restarts=False` for all OVN charms are not necessary, ovn charms / ovn can handle rolling upgrade themselves with `enable-auto-restarts=True` and `enable-version-pinning=False` and if ovn-central is upgraded before ovn-chassis.

https://docs.ovn.org/en/latest/intro/install/ovn-upgrades.html#rolling-upgrade

PR #504, opened 2024-07-19
Refactor how to restart service when `enable-auto-restart` is False.

chanchiwai-ray:
Restart how to handle `enable-auto-restarts=False` in rabbitmq server.

I've tested with

```
juju deploy rabbitmq-server --channel 3.8/stable --base ubuntu@20.04 -n3 --config source=distro
juju config rabbitmq-server enable-auto-restarts=False
cou upgrade --no-backup --auto-approve --no-archive
```


cou/utils/juju_utils.py
```diff
@@ -198,6 +198,15 @@ def is_from_charm_store(self) -> bool:
...
         return self.origin == "cs"
 
+    def get_units_to_run_action(self) -> list[Unit | SubordinateUnit]:
+        """Get the unit or subordinate units in the application to run action.
+
+        :return: The unit or subordinate units
+        :rtype: List of Unit or SubordinateUnit
+        """
+        units = self.subordinate_units if self.is_subordinate else self.units.values()
+        return units  # type: ignore[return-value]
```
samuelallan72, line 208: Why do we need to ignore the return type?

cou/utils/juju_utils.py
```diff
@@ -198,6 +198,15 @@ def is_from_charm_store(self) -> bool:
         """
         return self.origin == "cs"
 
+    def get_units_to_run_action(self) -> list[Unit | SubordinateUnit]:
+        """Get the unit or subordinate units in the application to run action.
+
```
samuelallan72, line 203: I'm not sure of the purpose from reading this - what does it mean to run action?

tests/unit/apps/test_auxiliary.py
```diff
@@ -395,21 +395,7 @@ def test_rabbitmq_server_upgrade_plan_ussuri_to_victoria_auto_restart_False(mode
...
-        )
-        for unit in app.units.values()
-    ]
-    upgrade_steps += [
-        PreUpgradeStep(
-            description=f"Wait for up to 2400s for app '{app.name}' to reach the idle state",
-            parallel=False,
-            coro=model.wait_for_active_idle(2400, apps=[app.name]),
-        ),
+        *tuple(app.get_run_deferred_hooks_and_restart_pre_upgrade_step()),
```
samuelallan72, line 398: Non blocker for this PR, but a general note that these tests are not really useful, because they aren't very declarative.  We're using the same method to generate the expected output as to generate the output we're testing. :confused:

cou/apps/auxiliary.py
```diff
@@ -196,35 +196,8 @@ def pre_upgrade_steps(
...
         steps = super().pre_upgrade_steps(target, units)
-        # Since auto restart is disabled, we don't know the if the service
-        # has pending events or not, so we want to run `run-deferred-hooks`
-        # action to clear the events before performing upgrade. If there
-        # are no pending events, this step should be a no-op, so it's safe
-        # to run anyway
         if self.config.get("enable-auto-restarts", {}).get("value") is False:
-            # Run any deferred events and restart the service. See
-            # https://charmhub.io/rabbitmq-server/actions#run-deferred-hooks
-            units_to_run_action = self.units.values() if units is None else units
```
samuelallan72, line 207: Old logic runs on the provided `units` arg if available.  The new logic ignores this arg.  I'm not sure if this is a problem?

cou/apps/base.py
```diff
@@ -607,6 +609,82 @@ def _get_charmhub_migration_step(self, target: OpenStackRelease) -> PreUpgradeSt
...
+        """
+        units = self.get_units_to_run_action()
+        run_hook_step = PreUpgradeStep(
+            description=(
+                "Execute run-deferred-hooks for all units to clear any leftover events:"
+                f" {', '.join([unit.name for unit in units])}"
+            ),
+            parallel=False,
+        )
+        run_hook_step.add_steps(
```
samuelallan72, line 626: I don't see the advantage of adding heirarchy here?

samuelallan72 (changes requested):

cou/utils/juju_utils.py
chanchiwai-ray: I think `get_units` is better, but it's kind of conflicting with `self.units`

cou/apps/base.py
chanchiwai-ray: it's cleaner in the plan

```
                        Execute run-deferred-hooks for all units to clear any leftover events: rabbitmq-server/6, rabbitmq-server/7, rabbitmq-server/8
                                Execute run-deferred-hooks on unit: 'rabbitmq-server/6'
                                Execute run-deferred-hooks on unit: 'rabbitmq-server/7'
                                Execute run-deferred-hooks on unit: 'rabbitmq-server/8'
```

cou/utils/juju_utils.py
samuelallan72: It's really confusing that we have get_units, units, subordinates, subordinate_units, and now get_units_to_run_action...

cou/utils/juju_utils.py
chanchiwai-ray: it kept complaining about return type should be `Collection[object]`, even though it seems correct: `list[Unit | SubordinateUnit]`

cou/apps/base.py
samuelallan72: Ah ok that's fair.  Could you remove the unit listing after `leftover events:` though?  That will make the line shorter.  The units are already listed underneath.

cou/utils/juju_utils.py
samuelallan72: Ah - that's because `units.values()` is a dict values class, not a list.  I guess `Collection` is a superset of list and dict values.

cou/apps/auxiliary.py
chanchiwai-ray: I don't think the `units` is even used. I check the root where the `units` is passed, and it kept as the default, which is `None` https://github.com/canonical/charmed-openstack-upgrader/blob/911328c1354340b8c59819c6b46519cdb14e41cf/cou/steps/plan.py#L758

If you prefer to be consistent, I can added that back

cou/apps/base.py
chanchiwai-ray: Thanks, I was thinking the same

cou/utils/juju_utils.py
chanchiwai-ray: I try doing `self.subordinate_units if self.is_subordinate else list(self.units.values())`, and same

cou/utils/juju_utils.py
chanchiwai-ray: 👌🏼

chanchiwai-ray:
I am moving `get_run_deferred_hooks_and_restart_pre_upgrade_step` and `get_run_deferred_hooks_and_restart_post_upgrade_step` to `auxiliary.py`  since it was mentioned by https://github.com/canonical/charmed-openstack-upgrader/pull/503#discussion_r1682830528

cou/utils/juju_utils.py
samuelallan72: Strange, I don't know :thinking:

cou/apps/auxiliary.py
samuelallan72: yeah that makes sense.  I don't mind; we can refactor that units arg out later. :)

chanchiwai-ray: Okay, let me add that back, and refactor the use of `units` in the future if it's necessary

cou/apps/base.py
chanchiwai-ray: Done

cou/apps/auxiliary.py
chanchiwai-ray: Done

tests/unit/apps/test_auxiliary.py
chanchiwai-ray: fixed

cou/apps/auxiliary.py
```diff
@@ -196,35 +283,8 @@ def pre_upgrade_steps(
         :rtype: list[PreUpgradeStep]
         """
         steps = super().pre_upgrade_steps(target, units)
-        # Since auto restart is disabled, we don't know the if the service
-        # has pending events or not, so we want to run `run-deferred-hooks`
-        # action to clear the events before performing upgrade. If there
-        # are no pending events, this step should be a no-op, so it's safe
-        # to run anyway
```
samuelallan72, line 203: Could you re-instate this information, perhaps in the docstring of the new `get_run_deferred_hooks_and_restart...` methods? :)

samuelallan72 (approved):
Thanks, LGTM.  Just one docs nit. :)

cou/apps/auxiliary.py
chanchiwai-ray: 🙆🏼‍♂️

samuelallan72 (approved):

jneo8 (approved):

PR #509, opened 2024-07-19
Feat/unseal vault

jneo8:
- Add unseal vault as post upgrade steps for vault upgrading
- Use hvac to unseal vault
    
fix: #467

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -13,9 +13,14 @@
 # limitations under the License.
 """Auxiliary application class."""
 import abc
+import base64
+import getpass
 import logging
+import os
+import tempfile
 from typing import Optional
 
+import hvac
 from packaging.version import Version
 
 from cou.apps.base import LONG_IDLE_TIMEOUT, OpenStackApplication
@@ -27,6 +32,7 @@ from cou.steps import (
     PreUpgradeStep,
     UnitUpgradeStep,
 )
+from cou.utils import progress_indicator
 from cou.utils.app_utils import set_require_osd_release_option
 from cou.utils.juju_utils import Unit
 from cou.utils.openstack import (
@@ -39,7 +45,7 @@ from cou.utils.openstack import (
 logger = logging.getLogger(__name__)
 
 
-@AppFactory.register_application(["vault", "ceph-fs", "ceph-radosgw"])
+@AppFactory.register_application(["ceph-fs", "ceph-radosgw"])
 class AuxiliaryApplication(OpenStackApplication):
     """Application for charms that can have multiple OpenStack releases for a workload."""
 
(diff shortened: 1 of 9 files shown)


cou/apps/auxiliary.py
```diff
@@ -502,3 +508,156 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+    async def _unseal_vault(self) -> None:
+        """Unseal vault on every vault unit."""
+        # Stop progress_indicator because it will clear the unseal key input.
+        progress_indicator.stop()
+
+        for unit_name in self.units:
+            logger.debug("Start unseal %s", unit_name)
+            cacert_file = self._get_cacert_file()
+            client = await self._get_vault_client(unit_name, cacert_file)
+            while True:
```
samuelallan72, line 595: I think the UX could be improved here - could we add some more feedback to the user in the loop?  For example:

- a message first to give instructions to the user (vault is sealed, you need to enter the unseal keys, cou will ask for unseal keys one at a time until vault is unsealed, blank lines are ignored)
- if a blank line is entered (with the current logic), a message should be printed that it was ignored (so the user doesn't worry)
- when vault is still sealed after entering an unseal key, tell the user that the unseal key was submitted successfully, but vault is still sealed (try entering more keys)
- when vault is successfully unsealed, print a success message
- also can we print the number of keys required? and the number of keys remaining to be entered?

cou/apps/auxiliary.py
```diff
@@ -502,3 +508,156 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+        :rtype: list[PostUpgradeStep]
+        """
+        upgrade_step = self._get_upgrade_charm_step(target=target)
+        steps = []
+
+        # Add unseal steps only if chaneel is changed.
+        if upgrade_step:
+            steps.extend(
+                [
+                    # Vault application should get into blocked and sealed status after upgrading.
```
samuelallan72, line 630: Are there cases where it will not become sealed?  For example, if there wasn't any available updates for vault and it doesn't upgrade and restart?

setup.cfg
```diff
@@ -33,6 +33,7 @@ install_requires =
     aioconsole
     halo
     setuptools
+    hvac
```
samuelallan72, line 36: It would possibly be safer to pin this to v2 major version, to avoid breaking changes when they release a v3.

cou/apps/auxiliary.py
```diff
@@ -502,3 +508,156 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+            timeout=self.wait_timeout,
+            status="blocked",
+            apps=[self.name],
+            # Vault application will go into error status first then blocked status.
+            raise_on_error=False,
+        )
+
+        app_status = await self.model.get_application_status(app_name=self.name)
+        if not app_status.status.info == "Unit is sealed":
+            raise ApplicationError("Vault not in sealed status")
```
samuelallan72, line 569: This might be confusing to the user - perhaps a longer explanation would be useful?  For example: Expected vault to become sealed after upgrade, but it did not.

samuelallan72 (changes requested):
Thanks, this looks pretty good.  I left some requests for improving the UX, a question about the requirement for vault to become sealed, and a couple of code thoughts.

cou/apps/auxiliary.py
jneo8: The logic inside the charm is every time the channel upgrade, it will be sealed. I don't think the situation exists.

cou/apps/auxiliary.py
jneo8: Can this one be a separate one? I would like the limit the scope here.

cou/apps/auxiliary.py
```diff
@@ -502,3 +508,156 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+            cacert_file = self._get_cacert_file()
+            client = await self._get_vault_client(unit_name, cacert_file)
+            while True:
+                status = client.sys.read_seal_status()
+                if not status["sealed"]:
+                    break
+                unseal_key = getpass.getpass("Unseal Key (will be hidden):")
+                if unseal_key:
+                    client.sys.submit_unseal_key(key=unseal_key)
+                del unseal_key
```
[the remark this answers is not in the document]
jneo8: Updated

cou/apps/base.py
```diff
@@ -68,6 +68,10 @@ class OpenStackApplication(Application):
     packages_to_hold: Optional[list] = field(default=None, init=False)
     wait_timeout: int = field(default=STANDARD_IDLE_TIMEOUT, init=False)
     wait_for_model: bool = field(default=False, init=False)  # waiting only for application itself
+    # Raise error when any unit or app going into "blocked" status
+    raise_on_blocked: bool = field(default=False, init=False)
+    # Raise error when any unit or app going into "error" status
+    raise_on_error: bool = field(default=True, init=True)
```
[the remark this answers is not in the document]
jneo8: Updated

cou/apps/auxiliary.py
jneo8: Updated

setup.cfg
jneo8: Updated

chanchiwai-ray (changes requested):
Using this PR, I cannot fix vault sealed issue because vault is in error status during the upgrade (due to connection refused). You need to resolved the vault unit to complete the upgrade.

Even after I manually `juju resolved  vault/0` ,cou still failed with the following error: https://pastebin.ubuntu.com/p/xGFhXvgnvm/. However, this appear to be a race condition: after you run `resolved` for nova-compute/0, it recovers before the timeout.

Subsequent, `cou upgrade` completes the upgrade.

In summary, I guess the plan should look like this:

```
Upgrade plan for 'vault' to 'zed'
        Upgrade software packages of 'vault' from the current APT repositories
                Ψ Upgrade software packages on unit 'vault/0'
        Upgrade 'vault' from '1.7/stable' to the new channel: '1.8/stable'
        + juju resolved vault/0
        Wait for up to 2400s for vault to reach the sealed status
        Unseal vault
        Wait for up to 2400s for vault to reach active status
        Resolve all applications in error status <--- optionally, loop this step until all resolved
        Wait for up to 2400s for model 'cou' to reach the idle state <--- this step timeout before nova-compute recovers
        Verify that the workload of 'vault' has been upgraded on units: vault/0
```

cou/apps/auxiliary.py
```diff
@@ -502,3 +508,162 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+            address = self.config["vip"].get("value")
+        else:
+            juju_unit = await self.model.get_unit(unit_name)
+            address = juju_unit.public_address
+
+        transport = "https" if self.config["ssl-cert"].get("value") else "http"
+        return f"{transport}://{address}:8200"
+
+    async def _wait_for_sealed_status(self) -> None:
+        """Wait for application vault go into sealed status.
```
chanchiwai-ray, line 555: sealed status is a bit confusing, juju does not have `sealed` status

jneo8:
> Using this PR, I cannot fix vault sealed issue because vault is in error status during the upgrade (due to connection refused). You need to resolved the vault unit to complete the upgrade.

The error status is the expected status, vault will auto-resolved by itself. You need to wait.

`Wait for up to 2400s for vault to reach the sealed status` This step will not raise exception even vault in error status. It's use `wait_for_idle` like:

```
wait_for_idle(..., raise_on_error=False)
```

I am a little bit confuse your issue here.

cou/apps/auxiliary.py
jneo8: It's blocked status which message is `Unit is sealed`. Sorry but I think sealed is more precise.

chanchiwai-ray:
>  vault will auto-resolved by itself. You need to wait.

Maybe I am not patient enough, the vault/0 is just in error state, and cannot upgrade

chanchiwai-ray:
After `Unseal vaul` step, it will be better to run update-status hook to speed up to process, but that can be in another PR

cou/apps/auxiliary.py
```diff
@@ -502,3 +508,162 @@ async def _verify_nova_compute(self, target: OpenStackRelease) -> None:
...
+
+        :raises ApplicationError: When application vault is not in sealed status.
+        """
+        await self.model.wait_for_idle(
+            timeout=self.wait_timeout,
+            status="blocked",
+            apps=[self.name],
+            # Vault application will go into error status first then blocked status.
```
samuelallan72, line 563: Could you add more information about this error status?  What error is expected?  Could you link to (or create and link to) a vault charm bug report?

cou/apps/auxiliary.py
samuelallan72: I think Ray means to avoid using the word "status" here, because it could be confused with juju status.

samuelallan72 (dismissed):
LGTM, thanks. :grin:  The remaining things I flagged are just nits.

samuelallan72 (approved):

chanchiwai-ray (approved):
I cannot verify this patch, but @jneo8 confirmed it was working for him

PR #510, opened 2024-07-22
Support `enable-auto-restarts=False` for ovn principal.

chanchiwai-ray:
Testing environment:

Using https://github.com/canonical/stsstack-bundles.git with `./generate-bundle.sh --name cou -r yoga -s jammy --ceph --run`, then 
```
juju deploy ovn-dedicated-chassis --channel 22.03/stable --base ubuntu@22.04
juju config ovn-dedicated-chassis enable-version-pinning=False
juju integrate ovn-central:ovsdb     ovn-dedicated-chassis:ovsdb
juju integrate vault:certificates    ovn-dedicated-chassis:certificates
juju config ovn-dedicated-chassis bridge-interface-mappings=$(juju config ovn-chassis bridge-interface-mappings)
juju remove-relation nova-compute ovn-chassis
juju config ovn-central  enable-auto-restarts=False
juju config ovn-dedicated-chassis enable-auto-restarts=False
```

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -431,6 +431,51 @@ class OVN(AuxiliaryApplication):
 class OVNPrincipal(OVN):
     """OVN principal application class."""
 
+    def pre_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[Unit]]
+    ) -> list[PreUpgradeStep]:
+        """Pre Upgrade steps planning.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[Unit]]
+        :return: List of pre upgrade steps.
+        :rtype: list[PreUpgradeStep]
+        """
+        steps = super().pre_upgrade_steps(target, units)
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
+            steps.extend(self.get_run_deferred_hooks_and_restart_pre_upgrade_step(units))
+        return steps
+
+    def post_upgrade_steps(
+        self, target: OpenStackRelease, units: Optional[list[Unit]]
+    ) -> list[PostUpgradeStep]:
+        """Post Upgrade steps planning.
+
+        Wait until the application reaches the idle state and then check the target workload.
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate post upgrade plan
+        :type units: Optional[list[Unit]]
+        :return: List of post upgrade steps.
+        :rtype: list[PostUpgradeStep]
+        """
+        steps = []
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
+            steps.extend(self.get_run_deferred_hooks_and_restart_post_upgrade_step(units))
+        steps.extend(super().post_upgrade_steps(target, units))
+        return steps
+
+    def _check_auto_restarts(self) -> None:
+        """No-op, skip check auto restarts option.
+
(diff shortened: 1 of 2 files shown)


samuelallan72 (approved):
Looks good to me.  Although I think we need at least manual confirmation the generated plan and a successful upgrade.

chanchiwai-ray:
> Looks good to me. Although I think we need at least manual confirmation the generated plan and a successful upgrade.

Testing right now

cou/apps/auxiliary.py
```diff
@@ -431,6 +431,51 @@ def upgrade_plan_sanity_checks(
...
+
+        :param target: OpenStack release as target to upgrade.
+        :type target: OpenStackRelease
+        :param units: Units to generate upgrade plan
+        :type units: Optional[list[Unit]]
+        :return: List of pre upgrade steps.
+        :rtype: list[PreUpgradeStep]
+        """
+        steps = super().pre_upgrade_steps(target, units)
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
```
jneo8, line 447: **nitpick (non-blocking):** 

Can this be a property?

cou/apps/auxiliary.py
```diff
@@ -431,6 +431,51 @@ def upgrade_plan_sanity_checks(
...
+        :return: List of post upgrade steps.
+        :rtype: list[PostUpgradeStep]
+        """
+        steps = []
+        if self.config.get("enable-auto-restarts", {}).get("value") is False:
+            steps.extend(self.get_run_deferred_hooks_and_restart_post_upgrade_step(units))
+        steps.extend(super().post_upgrade_steps(target, units))
+        return steps
+
+    def _check_auto_restarts(self) -> None:
```
jneo8, line 471: **nitpick:** 

Instead of override the method, can the behavior of function change base on the class attribute?

```
class OpenstackApplication:

    check_auto_restarts: bool = True
    
    def _check_auto_restarts(self) -> None:
        if self.check_auto_restarts:
            ...
            
```

jneo8 (changes requested):

cou/apps/auxiliary.py
chanchiwai-ray: I think this check will be gone when we support all charms that has `enable-auto-restarts` (ovn-dedicated-chassis, ovn-central, ovn-chassis, rabbitmq-server), so it's not important to change the behavior now

jneo8 (approved):

chanchiwai-ray:
I will need to ask ovn team for confirmation: the `run-deferred-hooks` does not behavior like the doc says https://charmhub.io/ovn-dedicated-chassis/actions#run-deferred-hooks

```
ovn-central/1                            active    idle   14       10.5.1.129      6641-6642/tcp   Services queued for restart: ovn-central.service
ovn-central/2*                           active    idle   15       10.5.3.164      6641-6642/tcp   Services queued for restart: ovn-central.service
ovn-dedicated-chassis/0*                 active    idle   19       10.5.0.207                      Services queued for stop: ovs-record-hostname.service, ovs-vswitchd.service, ovsdb-server.service. Hooks skipped due ...
```

cou/apps/auxiliary.py
chanchiwai-ray: same as below, I think in the future it won't be needed

chanchiwai-ray:
Closing it because the support of `enable-auto-restarts=False` for all OVN charms are not necessary, ovn charms / ovn can handle rolling upgrade themselves with `enable-auto-restarts=True` and `enable-version-pinning=False` and if ovn-central is upgraded before ovn-chassis.

https://docs.ovn.org/en/latest/intro/install/ovn-upgrades.html#rolling-upgrade

PR #517, opened 2024-07-29
refactor: Change the control/data plane fields to property func

jneo8:
Change Analysis class's apps_data_plane & apps_control_plane fields to property functions for more extensibility.


samuelallan72:
This is being worked on and reviewed over at #518, so I think we can close this @jneo8

jneo8:
Merging in #518

PR #518, opened 2024-07-29
Feat/ovn subordinate upgrade before control plane

jneo8:
Move ovn subordinate upgrade steps before principle

Fix: #468

### DIFF
--- a/cou/steps/analyze.py
+++ b/cou/steps/analyze.py
@@ -40,8 +40,7 @@ class Analysis:
     """
 
     model: juju_utils.Model
-    apps_control_plane: list[OpenStackApplication]
-    apps_data_plane: list[OpenStackApplication]
+    apps: list[OpenStackApplication]
     min_o7k_version_control_plane: Optional[OpenStackRelease] = None
     min_o7k_version_data_plane: Optional[OpenStackRelease] = None
 
@@ -54,6 +53,26 @@ class Analysis:
         self.min_o7k_version_data_plane = self.min_o7k_release_apps(self.apps_data_plane)
         self.current_cloud_o7k_release = self._get_minimum_cloud_o7k_release()
         self.current_cloud_series = self._get_minimum_cloud_series()
+
+    @property
+    def apps_control_plane(self) -> list[OpenStackApplication]:
+        """Return list of control plane applications.
+
+        :return: Control plane application lists.
+        :rtype: list[OpenStackApplication]
+        """
+        control_plane, _ = self._split_apps(self.apps)
+        return control_plane
+
+    @property
+    def apps_data_plane(self) -> list[OpenStackApplication]:
+        """Return list of data plane applications.
+
+        :return: data plane application lists.
+        :rtype: list[OpenStackApplication]
+        """
+        _, data_plane = self._split_apps(self.apps)
+        return data_plane
 
     @staticmethod
     def _split_apps(
@@ -81,6 +100,7 @@ class Analysis:
         data_plane_machines = {
             unit.machine for app in apps if is_data_plane(app) for unit in app.units.values()
         }
+
         for app in apps:
             if is_data_plane(app):
                 data_plane.append(app)
@@ -103,9 +123,7 @@ class Analysis:
         logger.info("Analyzing the OpenStack deployment...")
         apps = await Analysis._populate(model)
 
-        control_plane, data_plane = cls._split_apps(apps)
-
-        return Analysis(model=model, apps_data_plane=data_plane, apps_control_plane=control_plane)
+        return Analysis(model=model, apps=apps)
 
     @classmethod
     async def _populate(cls, model: juju_utils.Model) -> list[OpenStackApplication]:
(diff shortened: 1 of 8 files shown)


tests/mocked_plans/sample_plans/base.yaml
```diff
@@ -4,6 +4,9 @@ plan: |
         Verify that all OpenStack applications are in idle state
         Back up MySQL databases
         Archive old database data on nova-cloud-controller
+        OVN subordinate upgrade plan
+            Upgrade plan for 'ovn-chassis' to 'victoria'
+                Refresh 'ovn-chassis' to the latest revision of '22.03/stable'
         Control Plane subordinate(s) upgrade plan
             Upgrade plan for 'keystone-ldap' to 'victoria'
                 Refresh 'keystone-ldap' to the latest revision of 'ussuri/stable'
```
samuelallan72, line 12: Why not simply move the OVN subordinate applications into the control plane subordinates list?  This would solve the problem without any added complexity.

samuelallan72 (changes requested):

tests/mocked_plans/sample_plans/base.yaml
jneo8: Because they are not control plane subordinate as I understand, if we follow the definition of control/data plane and principle/subordinate.

samuelallan72: If that's the only reason, then I vote we put it in the control plane subordinates list with a comment about why it's in that list.

jneo8: Two more reasons:

- I am thinking the OVN(ovn-central, ovn-chassis, ...) upgrade can be a separate, independent upgrade. We may lose some flexibility if we want to do this in the future.
- I don't like to make upgrade group definition become more vague. To me it's more clear not the group them together.

samuelallan72: > I am thinking the OVN(ovn-central, ovn-chassis, ...) upgrade can be a separate, independent upgrade.

What are the use cases for this?  I don't think we have a case where you only want to upgrade ovn.

> I don't like to make upgrade group definition become more vague. 

Our upgrade groups are already arbitrary.  I think the main reason is so we can upgrade the hypervisors individually.  Outside of that, we don't have to follow any other constraint.

I don't want to add all this extra complexity just to change the upgrade order.

jneo8: > What are the use cases for this? I don't think we have a case where you only want to upgrade ovn.

If we want to support ovn 20.04 -> 22.03

> I don't want to add all this extra complexity just to change the upgrade order.

Putting ovn-chassis in `apps_control_plane` also increase the complexity in code. I don't see big issue to add one more upgrade group before control plane subordinate, to me it's equal on the upgrade order and avoid the misunderstand of the control/data plane and principle/subordinate.

cou/steps/analyze.py
```diff
@@ -55,6 +59,42 @@ def __post_init__(self) -> None:
...
+
+        :return: Control plane application lists.
+        :rtype: list[OpenStackApplication]
+        """
+        control_plane, _ = self._split_apps(self.apps)
+        return control_plane
+
+    @property
+    def apps_data_plane(self) -> list[OpenStackApplication]:
+        """Return list of data plane applications.
```
Pjack, line 74: Could we add a description to explain why OVN is not included in the expected output? Users might misuse this function if they only read the description. 

Another option is to exclude OVN by the caller. (out side of this function)

Pjack (changes requested):

cou/steps/analyze.py
jneo8: It will be nice we filter ovn-chassis here. It will be more easy to maintain in the future.

tests/mocked_plans/sample_plans/base.yaml
samuelallan72: Ok I see how the current design of COU makes it different to cleanly implement this.

samuelallan72 (approved):
I don't like this design and continuing this way will continue to snowball the complexity in cou, continuing to make cou more difficult to maintain.  However, as discussed, we're currently constrained by COU's architecture and aren't planning to restructure COU right now.

So this seems ok to go with for now.  I highly recommend spending some time to redesign how we manage upgrade groups and ordering soon though.

cou/steps/analyze.py
jneo8: Updated

cou/steps/analyze.py
```diff
@@ -55,6 +59,42 @@ def __post_init__(self) -> None:
...
+
+        :return: data plane application lists.
+        :rtype: list[OpenStackApplication]
+        """
+        _, data_plane = self._split_apps(self.apps)
+        # Remove ovn subordinates from the data plane since it have to upgrade before
```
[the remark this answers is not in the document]
jneo8: Updated

samuelallan72 (approved):

chanchiwai-ray (approved):
Although this addresses the issue in the title, I think a proper fix is in the design of COU: we should make the plan declarative.

Pjack (approved):

PR #515, opened 2024-07-25
Ignore juju docs for linkcheck

samuelallan72:
This url is valid, and the anchor exists:
https://juju.is/docs/juju/manage-controllers#heading--bootstrap-a-controller

However, often linkcheck fails on this url with this message:

```
https://juju.is/docs/juju/manage-controllers#heading--bootstrap-a-controller - Anchor 'heading--bootstrap-a-controller' not found
```

PR #519, opened 2024-07-30
Fix doc builds

chanchiwai-ray:
Fix the documentation build failures: https://readthedocs.com/projects/canonical-charmed-openstack-upgrader/builds/

### DIFF
--- a/docs/how-to/index.rst
+++ b/docs/how-to/index.rst
@@ -18,3 +18,4 @@ are possible with **COU**.
    archive-old-data
    verbosity
    purge-data-on-shadow-table
+   test-on-juju-openstack-provider
--- a/docs/reference/security.rst
+++ b/docs/reference/security.rst
@@ -14,4 +14,4 @@ prerequisite for using Charmed OpenStack Upgrader.
 Please refer to `Juju's documentation`_ for more details.
 
 .. LINKS:
-.. _Juju documentation: https://juju.is/docs/juju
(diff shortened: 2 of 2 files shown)


samuelallan72:
While we're looking at docs fixes, can we also merge https://github.com/canonical/charmed-openstack-upgrader/pull/515 that fixes this issue:

> (how-to/different-model: line   20) broken    https://juju.is/docs/juju/environment-variables#heading--jujudata - Anchor 'heading--jujudata' not found

samuelallan72 (commented):
It sounds like this was broken for a while - what do we need to do to avoid this in future?  Can we improve the workflow to check that docs build successfully in CI?

jneo8 (approved):

jneo8:
> It sounds like this was broken for a while - what do we need to do to avoid this in future? Can we improve the workflow to check that docs build successfully in CI?

Shouldn't check by https://github.com/canonical/documentation-workflows ?

chanchiwai-ray:
> It sounds like this was broken for a while - what do we need to do to avoid this in future? Can we improve the workflow to check that docs build successfully in CI?

There is a new(?) integration that does exactly this `docs/readthedocs.com:canonical-charmed-openstack-upgrader — Read the Docs build succeeded!`

samuelallan72:
Ah awesome, good to see :+1:

samuelallan72 (approved):

PR #520, opened 2024-08-01
Remove config related to Sonar

samuelallan72:
We experimented with using Sonar,
but decided not to continue due to it not working as desired with our open source workflow.


.github/workflows/check.yaml
```diff
@@ -35,13 +35,6 @@ jobs:
...
-    - name: Save PR number to file
-      run: echo ${{ github.event.number }} > PR_NUMBER.txt
-    - name: Archive PR number
-      uses: actions/upload-artifact@v4
-      with:
-        name: PR_NUMBER
-        path: PR_NUMBER.txt
     - name: Archive code coverage results
       uses: actions/upload-artifact@v4
       with:
```
chanchiwai-ray, line 40: I believe this is also for sonar

chanchiwai-ray (changes requested):

.github/workflows/check.yaml
samuelallan72: +1

samuelallan72: Updated :)

chanchiwai-ray (approved):
LGTM

rgildein (approved):

Pjack (approved):

Issue #472, opened 2024-07-01
[vault] newer version stopping cou plan/upgrade


honghan-wong:
When we are running cou on some clouds on ussuri, the vault charm is running on a later version `1.10.6`. 

The current cou is allow only vault version `1.7` on ussuri version.

Is it possible to have cou support some charms where a later version will not cause incompatible between openstack components?

Example is vault `1.10.6` doesn't introduce breaking change and should be able to support the upgrade between ussuri and yoga.
If the vault is on a later version, cou might be able to skip the upgrade for vault.

jneo8:
A very quick option is to update `cpu/utils/openstack_to_track mapping.csv` from your expectation.

A more general option is that we we provide `--skip-apps` argument for the user to skip the upgrade steps for some application.

gabrielcocenza:
One thing that it's not clear to me is why the vault is using `1.10`. By the charm [page](https://charmhub.io/vault) seems that we don't have this option on channels. @honghan-wong what is the channel  that the charm is using?

honghan-wong:
vault charm is using rev `54`, channel `latest/stable`
```
vault                     1.10.6   active      3  vault                     latest/stable   54  no       Unit is ready (active: true, mlock: enabled)
```

We have met issue before on vault charm where the snap has refreshed the vault to a later version and triggered an alert from nrpe check mentioning version mismatch.
To solve it, we have to restart the vault service and the charm will refreshed to the later version.
And I believe this is one of the case, where the charm is on an older revision but vault has upgrade to a later version.

aieri:
@honghan-wong you're also using the `latest/stable` track, which is closed. The charm should be switched to track `1.7/stable`. I don't know if you can downgrade the vault snap without breakages though.

samuelallan72:
An option is to add some kind of ignore flag, along the lines of https://github.com/canonical/charmed-openstack-upgrader/issues/478 .  Then if it's not possible on a cloud to switch the charm or workload to something recognised by COU, and the engineer upgrading the cloud determines that it's fine, they can tell cou to ignore or skip it.

jneo8:
Proposal to skip +1

We can have a CLI argument to skip vault upgrading.

PR #522, opened 2024-08-02
Add skip apps command line flag.

chanchiwai-ray:
Add `--skip-apps` flag to allow COU to skip some applications, likely some auxiliary charms in charmed openstack deployment such as vault. And this feature currently only support vault.

The motivation behind this is that we observe some applications in a charmed openstack deployment can be deviated from the recommended setup and still working fine. Vault is an example: the vault version (1.10.x) is above the recommended version for current openstack version ([reference](https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html)) but still working fine on production. The problem is worsen because users are not recommended to [downgrade vault](https://developer.hashicorp.com/vault/docs/v1.10.x/upgrading).  So we need a workaround for user to skip upgrading some applications that are known to be working and can't  afford to switch to recommend version.

Closes: #472

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -217,6 +217,9 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         nargs="+",
         help=(
             "Skip upgrading the given applications."
+            "\nNote that skip upgrading applications is dangerous, and could leave"
+            "\nthe cloud in an unstable state. You should only use this option if"
+            "\nyou know the applications will not affect the cloud during an upgrade."
             "\nCurrently, it only supports skip upgrading vault."
         ),
         required=False,
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -494,7 +494,7 @@ class CLIargs:
     availability_zones: Optional[set[str]] = None
     purge: bool = False
     purge_before: Optional[str] = None
-    skip_apps: Optional[list[str]] = field(default_factory=list)
+    skip_apps: list[str] = field(default_factory=list)
 
     @property
     def prompt(self) -> bool:
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -145,6 +145,20 @@ class PurgeBeforeArgumentAction(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
+class SkipAppsArgumentAction(argparse.Action):
+    """Custom action to make sure the arguments is unique."""
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Any,
+        option_string: Optional[str] = None,
+    ) -> None:
+        unique_values = set(values)
+        setattr(namespace, self.dest, unique_values)
+
+
 def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
     """Create a shared parser for options specific to subcommands.
 
@@ -207,6 +221,18 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
             "\nWithout --purge-before-date the purge step will delete all the data."
         ),
         type=purge_before_arg,
+        required=False,
+    )
+    subcommand_common_opts_parser.add_argument(
+        "--skip-apps",
+        dest="skip_apps",
+        choices=["vault"],
+        action=SkipAppsArgumentAction,
+        nargs="*",
+        help=(
+            "Skip upgrading the given applications."
+            "\nCurrently, it only supports skip upgrading vault."
+        ),
         required=False,
     )
     subcommand_common_opts_parser.add_argument(
@@ -479,6 +505,7 @@ class CLIargs:
     availability_zones: Optional[set[str]] = None
     purge: bool = False
     purge_before: Optional[str] = None
+    skip_apps: Optional[set[str]] = None
 
     @property
     def prompt(self) -> bool:
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -15,7 +15,7 @@
 """Command line arguments parsing for 'charmed-openstack-upgrader'."""
 import argparse
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any, Iterable, Optional
 
@@ -145,20 +145,6 @@ class PurgeBeforeArgumentAction(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
-class SkipAppsArgumentAction(argparse.Action):
-    """Custom action to make sure the arguments is unique."""
-
-    def __call__(
-        self,
-        parser: argparse.ArgumentParser,
-        namespace: argparse.Namespace,
-        values: Any,
-        option_string: Optional[str] = None,
-    ) -> None:
-        unique_values = set(values)
-        setattr(namespace, self.dest, unique_values)
-
-
 def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
     """Create a shared parser for options specific to subcommands.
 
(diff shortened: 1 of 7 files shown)


cou/commands.py
```diff
@@ -209,6 +223,18 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
         type=purge_before_arg,
         required=False,
     )
+    subcommand_common_opts_parser.add_argument(
+        "--skip-apps",
+        dest="skip_apps",
+        choices=["vault"],
+        action=SkipAppsArgumentAction,
```
samuelallan72, line 230: I think we can drop the action - it doesn't bring much benefit.  If the user does add duplicated options, it won't change the logic.  Keeping it as a list also doesn't have much performance impact because the list has such a small number of items.

cou/commands.py
```diff
@@ -209,6 +223,18 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
...
+        nargs="*",
```
samuelallan72, line 231:
```suggestion
        nargs="+",
        default=[],
```

With nargs `*`, it's possible to pass `--skip-apps` with no arguments, which doesn't make much sense.  

Also, adding a default to empty list means that we'll always have a list as the value for args.skip_apps.  This means we can also drop the `skip_apps or set()` logic.

samuelallan72 (changes requested):

cou/commands.py
chanchiwai-ray: > With nargs *, it's possible to pass --skip-apps with no arguments, which doesn't make much sense.

Yeah, I saw that too, that's why there's `skip_apps or set()`.

cou/commands.py
chanchiwai-ray: Sure, that's the alternative implementation, I don't oppose change to that

jneo8 (commented):
Do you think there is any check required because we provide an option to skip the app. This may cause a risk that user make the cloud not in a consistence application version. For example ovn-central required ovn-chassis to be upgraded first, if you skip ovn-chassis, you will have a chance to upgrade ovn-central first.

chanchiwai-ray:
> Do you think there is any check required because we provide an option to skip the app. This may cause a risk that user make the cloud not in a consistence application version. For example ovn-central required ovn-chassis to be upgraded first, if you skip ovn-chassis, you will have a chance to upgrade ovn-central first.

That is a valid concern for the future.

jneo8:
I am ok to separate the skip and checks, but can you open an issue to trace it?

jneo8 (dismissed):

cou/commands.py
gabrielcocenza: There is already the SplitArgs that transform the options passed into a set to avoid duplication that can be used. https://github.com/canonical/charmed-openstack-upgrader/blob/5f9959acdbb735ef8cddaf5bf0e9df7d8d7990c2/cou/commands.py#L80-L94

gabrielcocenza (commented):
I'm not following the full motivation behind this change. Is this being implemented because some production clouds are using a not supported version of vault and the plan fails?

I would prefer trying to not add a new command in this case. Couldn't we stick with the already implemented `--force` flag?

Moreover, I don't think we'll ever allow more apps to be skipped because it's a amazing way to let users shoot in the foot.

In my opinion the ideal scenario would let use the `--force` flag to be able to generate the plan, but not allow to upgrade.

samuelallan72:
@gabrielcocenza 

> In my opinion the ideal scenario would let use the --force flag to be able to generate the plan,

`--force` currently only forces upgrade of non-empty hypervisors.  This would be changing the semantics a lot.

 > be able to generate the plan, but not allow to upgrade.

This would not be useful.  The use case here is that some clouds may have a much newer version of vault than officially supported.  Vault cannot be downgraded, so the only option is to force COU to stop blocking on the unsupported vault version.

My preference would be to either update the version lookup.csv to allow these newer versions, or to have a generic ignore/skip so an operator can force upgrade.  I don't think this implementation is ideal, but it is a way to unblock an upcoming upgrade at short notice.

> a amazing way to let users shoot in the foot.

There is a fine line between footguns and not getting in the way of experienced users.  Clouds that will be used with COU will have oddities and COU must work with them.

cou/commands.py
```diff
@@ -479,6 +494,7 @@ class CLIargs:
     availability_zones: Optional[set[str]] = None
     purge: bool = False
     purge_before: Optional[str] = None
+    skip_apps: Optional[list[str]] = field(default_factory=list)
```
samuelallan72, line 497:
```suggestion
    skip_apps: list[str] = field(default_factory=list)
```

cou/steps/analyze.py
```diff
@@ -136,18 +136,24 @@ def is_data_plane(app: OpenStackApplication) -> bool:
...
+        :type skip_apps: set of string
         :return: Analysis object populated with the model applications.
         :rtype: Analysis
         """
         logger.info("Analyzing the OpenStack deployment...")
         apps = await Analysis._populate(model)
 
-        return Analysis(model=model, apps=apps)
+        return Analysis(
+            model=model, apps=[app for app in apps if app.name not in (skip_apps or [])]
```
samuelallan72, line 155:
```suggestion
            model=model, apps=[app for app in apps if app.name not in skip_apps]
```

cou/steps/analyze.py
```diff
@@ -136,18 +136,24 @@ def is_data_plane(app: OpenStackApplication) -> bool:
         return control_plane, data_plane
 
     @classmethod
-    async def create(cls, model: juju_utils.Model) -> Analysis:
+    async def create(
+        cls, model: juju_utils.Model, skip_apps: Optional[list[str]] = None
```
samuelallan72, line 140:
```suggestion
        cls, model: juju_utils.Model, skip_apps: list[str]
```

samuelallan72 (changes requested):
Some suggestions to fix the typing now that you've switched to `default=[]` and `nargs="+"`.

cou/steps/analyze.py
chanchiwai-ray: I am aware of this, but this is also why I don't want to make this change: it breaks compatibility of unit tests since it's not optional

samuelallan72: >  it breaks compatibility of unit tests since it's not optional

This should never be a reason to not improve application code.

samuelallan72 (approved):

cou/commands.py
chanchiwai-ray: Thanks, I don't know about this. But to me using comma separated is not as good as `--skip-apps a b c d`, it would be nice if that's for `nargs="+"` or `nargs="*"`

cou/steps/analyze.py
chanchiwai-ray: but it did not improve the code's logic: that `skip_apps` is supposed to be optional not a requirement

jneo8 (approved):

cou/steps/analyze.py
samuelallan72: It made the typing more accurate and removed redundant logic (the `or []` part). :)

cou/commands.py
chanchiwai-ray: Resolving this, thanks for the suggestion

gabrielcocenza:
@chanchiwai-ray Thanks for clarifying.

Please add some more context in the PR description saying that some clouds might have newer versions of vault and that it's not recommended to roll back because as the upstream [documentation says](https://developer.hashicorp.com/vault/docs/v1.10.x/upgrading):

> Simply replacing the newly-installed Vault binary with the previous version will not cleanly downgrade Vault, as upgrades may perform changes to the underlying data structure that make the data incompatible with a downgrade.

And tag that this PR solves #472 

 I think it's necessary to also include documentation about the `--skip-apps` on `how-to`. Do you plan to add it in a different PR?

chanchiwai-ray:
> I think it's necessary to also include documentation about the --skip-apps on how-to. Do you plan to add it in a different PR?

Let's add this to a different PR

PR #528, opened 2024-08-23
Wait for units to settle after charm refresh

samuelallan72:
Avoids hitting race conditions in charms,
such as https://bugs.launchpad.net/charm-barbican/+bug/2039604

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -31,6 +31,7 @@ from cou.steps import (
     PostUpgradeStep,
     PreUpgradeStep,
     UnitUpgradeStep,
+    UpgradeStep,
 )
 from cou.utils import progress_indicator
 from cou.utils.app_utils import set_require_osd_release_option
@@ -288,7 +289,7 @@ class RabbitMQServer(AuxiliaryApplication):
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
-    ) -> list[PreUpgradeStep]:
+    ) -> list[UpgradeStep]:
         """Pre Upgrade steps planning.
 
         :param target: OpenStack release as target to upgrade.
@@ -296,7 +297,7 @@ class RabbitMQServer(AuxiliaryApplication):
         :param units: Units to generate upgrade plan
         :type units: Optional[list[Unit]]
         :return: List of pre upgrade steps.
-        :rtype: list[PreUpgradeStep]
+        :rtype: list[UpgradeStep]
         """
         steps = super().pre_upgrade_steps(target, units)
         if self.config.get("enable-auto-restarts", {}).get("value") is False:
@@ -341,7 +342,7 @@ class CephMon(AuxiliaryApplication):
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
-    ) -> list[PreUpgradeStep]:
+    ) -> list[UpgradeStep]:
         """Pre Upgrade steps planning.
 
         :param target: OpenStack release as target to upgrade.
@@ -349,7 +350,7 @@ class CephMon(AuxiliaryApplication):
         :param units: Units to generate upgrade plan
         :type units: Optional[list[Unit]]
         :return:  List of pre upgrade steps.
-        :rtype: list[PreUpgradeStep]
+        :rtype: list[UpgradeStep]
         """
         return super().pre_upgrade_steps(target, units) + [
             self._get_change_require_osd_release_step()
@@ -462,7 +463,7 @@ class CephOsd(AuxiliaryApplication):
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
-    ) -> list[PreUpgradeStep]:
+    ) -> list[UpgradeStep]:
         """Pre Upgrade steps planning.
 
         :param target: OpenStack release as target to upgrade.
@@ -470,9 +471,9 @@ class CephOsd(AuxiliaryApplication):
         :param units: Units to generate upgrade plan
         :type units: Optional[list[Unit]]
         :return: List of pre upgrade steps.
-        :rtype: list[PreUpgradeStep]
+        :rtype: list[UpgradeStep]
         """
-        steps = [
+        steps: list[UpgradeStep] = [
             PreUpgradeStep(
                 description="Verify that all 'nova-compute' units has been upgraded",
                 coro=self._verify_nova_compute(target),
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -31,7 +31,6 @@ from cou.steps import (
     PostUpgradeStep,
     PreUpgradeStep,
     UnitUpgradeStep,
-    UpgradeStep,
 )
 from cou.utils import progress_indicator
 from cou.utils.app_utils import set_require_osd_release_option
@@ -289,7 +288,7 @@ class RabbitMQServer(AuxiliaryApplication):
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
-    ) -> list[UpgradeStep]:
+    ) -> list[PreUpgradeStep]:
         """Pre Upgrade steps planning.
 
         :param target: OpenStack release as target to upgrade.
@@ -297,7 +296,7 @@ class RabbitMQServer(AuxiliaryApplication):
         :param units: Units to generate upgrade plan
         :type units: Optional[list[Unit]]
         :return: List of pre upgrade steps.
-        :rtype: list[UpgradeStep]
+        :rtype: list[PreUpgradeStep]
         """
         steps = super().pre_upgrade_steps(target, units)
         if self.config.get("enable-auto-restarts", {}).get("value") is False:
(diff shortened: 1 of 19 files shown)


samuelallan72:
@canonical/soleng-reviewers I'd appreciate your thoughts on the implementation here.  I haven't updated all the tests yet, but please see the first commit in this PR for the actual app code changes.  I'm not sure if this is the best way to implement it, but the goal is to get COU to wait for the charm to finish any upgrade-charm hooks after running a charm refresh.

Specifically, it appears that waiting between `Upgrade <charm> from <old channel> to the new channel: <new channel>` and `Change charm config of <charm> 'openstack-origin' to 'cloud:focal-<new release>'` is a workaround to https://bugs.launchpad.net/charm-barbican/+bug/2039604

In general, it will be better for COU to wait for the charm to finish processing hooks after any changes (change config, refresh charm, etc.).

Pjack (approved):

chanchiwai-ray (commented):
the idea is safe to me since it's only adding wait steps

samuelallan72:
> However, just to leave it as note, we might need to increase global default timeout.

I don't think there is a global timeout?

chanchiwai-ray:
> I don't think there is a global timeout?

okay, maybe I was wrong. anyways, it's out of scope now

chanchiwai-ray (approved):
LGTM, but I believe there should be some doc change some where

cou/apps/auxiliary.py
```diff
@@ -288,15 +289,15 @@ class RabbitMQServer(AuxiliaryApplication):
 
     def pre_upgrade_steps(
         self, target: OpenStackRelease, units: Optional[list[Unit]]
-    ) -> list[PreUpgradeStep]:
+    ) -> list[UpgradeStep]:
```
gabrielcocenza, line 292: In the end of the day `PreUpgradeStep`, `UnitUpgradeStep`, `PostUpgradeStep` are the `UpgradeStep`, but I don't see a reason why to change the type here and in other parts of the code. Can't we keep the way that it is right now?

tests/functional/tests/smoke.py
```diff
@@ -146,6 +146,7 @@ def generate_expected_plan(self, backup: bool = True) -> str:
             "\t\t\tUpgrade 'designate-bind' from 'ussuri/stable' to the new channel: "
             "'victoria/stable'\n"
             "\t\t\tWait for up to 300s for app 'designate-bind' to reach the idle state\n"
+            "\t\t\tWait for up to 300s for app 'designate-bind' to reach the idle state\n"
```
gabrielcocenza, line 149: Why are we having a duplicate here?

tests/mocked_plans/sample_plans/018346c5-f95c-46df-a34e-9a78bdec0018.yaml
```diff
@@ -90,6 +126,7 @@ plan: |
                     Ψ Upgrade software packages on unit 'ceph-mon/1'
                     Ψ Upgrade software packages on unit 'ceph-mon/2'
                 WARNING: Changing 'ceph-mon' channel from latest/stable to octopus/stable. This may be a charm downgrade, which is generally not supported.
+                Wait for up to 2400s for model '018346c5-f95c-46df-a34e-9a78bdec0018' to reach the idle state
```
gabrielcocenza, line 129: I think that the wait time here shouldn't be the same as waiting for the application to upgrade the payload. I think it can be less like 300 seconds

cou/apps/base.py
```diff
@@ -569,28 +569,31 @@ def _get_upgrade_current_release_packages_step(
...
         :type target: OpenStackRelease
         :raises ApplicationError: When application has unexpected channel.
-        :return: Step for refreshing the charm
-        :rtype: PreUpgradeStep
+        :return: Steps for refreshing the charm
+        :rtype: list[PreUpgradeStep]
         """
         if self.is_from_charm_store:
-            return self._get_charmhub_migration_step(target)
+            return [self._get_charmhub_migration_step(target), self._get_wait_step()]
```
gabrielcocenza, line 582: I think this is a nice opportunity to create a `add_wait_step` decorator instead of changing individual functions. Something like:

```python
# add wait step in the last position of the list of steps
@add_wait_step(last=True)
def _get_refresh_charm_steps():
    pass
 
# add wait step in the first position of the list of steps
@add_wait_step(first=True)
def  post_upgrade_steps():
    pass
```

gabrielcocenza (changes requested):
@samuelallan72 so if I understood correctly if you manually upgrade the cloud the issue doesn't happen because you give more time to the application settle?

cou/apps/auxiliary.py
samuelallan72: I had to, because mypy was not happy.  Ideally we would remove  PreUpgradeStep, UnitUpgradeStep, and PostUpgradeStep and replace them all with `UpgradeStep`; the distinction is not useful.

tests/functional/tests/smoke.py
samuelallan72: Because of how everything is split into various methods - I think the post upgrade steps always puts in a wait.  From the upgrade steps I can't really communicate nicely with the post upgrade steps to deduplicate.

tests/mocked_plans/sample_plans/018346c5-f95c-46df-a34e-9a78bdec0018.yaml
samuelallan72: I don't want to make it overcomplicated; and this is just the timeout value.  It should return a lot quicker than this.

cou/apps/base.py
samuelallan72: I don't think so, because that `add_wait_step` needs context on the class (access to self.model, etc.), and it will depend on what upgrade steps are used.

I'm also very hesitant to refactor code here, especially refactoring while adding new features, because of the large effort to update the unit tests.  Already as it is, this is a 500 line change. :(

samuelallan72:
@gabrielcocenza 

> so if I understood correctly if you manually upgrade the cloud the issue doesn't happen because you give more time to the application settle?

Correct.  And waiting for the unit to settle in between making changes to it is good practice in general. :)

cou/apps/auxiliary.py
gabrielcocenza: As far as I know I agree that it's not useful. It was a bet in the past that it could possibly help us and I guess in the end it didn't help at all.

With that said, I don't feel comfortable changing the type in those functions just because it makes more convenient for this PR and let the code into an inconsistency state where some places still uses it. I know that refactoring that will increase substantially the number of lines changed and unit tests will break.

IMHO I think reusing the implementaion of `_get_wait_step` is a wrong approach because of the following reasons:

- That was designed especially for `PostUpgradeStep` and some application like ceph-osd, ocatvia waits for the whole model to be active/idle. We don't need that behavior for just refreshing the charm. Just checking the app itself would be enough in this case
- The time can be way larger than necessary (not a real issue), but it feels awkward to see that we are going to wait like 2400 seconds for a simple charm refresh.

I think we can simply manually add the wait step and keep the current types. E.g:

```python
    def _get_refresh_charm_steps(self, target: OpenStackRelease) -> list[PreUpgradeStep]:
        """Get steps for refreshing the charm.
        :param target: OpenStack release as target to upgrade
        :raises ApplicationError: When application has unexpected channel.
        :return: Steps for refreshing the charm
        :rtype: list[PreUpgradeStep]
        """
        wait_step = PreUpgradeStep(
            description="Wait up to 120s for app '{self.name}' to reach idle state",
            parallel=False,
            coro=self.model.wait_for_idle(self.wait_timeout, apps=[self.name])
         )
        if self.is_from_charm_store:
            return [self._get_charmhub_migration_step(target), wait_step]
        ....
```

tests/functional/tests/smoke.py
gabrielcocenza: I think now I see the reason. Channel based apps like `designate-bind` upgrades just by changing the channel so you'll have this duplication because it's a special case where there is not `_get_change_install_repository_step` and `_get_units_upgrade_steps`.

cou/apps/auxiliary.py
samuelallan72: > I don't feel comfortable changing the type in those functions just because it makes more convenient for this PR and let the code into an inconsistency state where some places still uses it. 

It's quite the opposite in this PR - I've changed the type of a function, and updated types across the codebase accordingly so it is consistent.

> I think reusing the implementaion of _get_wait_step is a wrong approach

I see your point here, that makes sense.  I was hoping to reduce duplication by reusing the wait logic, but if the wait logic can be simpler for these cases, I'll update it there too.

samuelallan72: Updated, thanks for your comments :)

gabrielcocenza (commented):
Thanks LGTM, but linting is failing

cou/apps/auxiliary.py
gabrielcocenza: > It's quite the opposite in this PR - I've changed the type of a function, and updated types across the codebase accordingly so it is consistent.

You would still have `PreUpgradeStep` on other parts of the project like hypervisors, plan and etc. That was the  inconsistency that I was talking about.

Closing this thread once the approach changed.

Pjack (approved):

Issue #491, opened 2024-07-10
[hacluster] request to support hacluster in cou


honghan-wong:
Hi,

The COU is reporting issue on hacluster not supported.

```
2024-07-10 06:25:45 [ERROR] Channel: stable for charm 'hacluster' on series 'focal' is not supported by COU. Please take a look at the documentation: https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html to see if you are using the right track.
```

checking the csv list, i couldn't find the hacluster entry.

the current rev running in the cloud that ran cou is
hacluster ([charmhub link](https://charmhub.io/hacluster)) `latest/stable` `83`

jneo8:
This is raise by a custom local build snap which commit hash is `d3e4718`

honghan-wong:
sorry about the title, it is misleading. I ran cou on some other environment and its working, there should be another underlying issue causing this error.

jneo8:
Hi @honghan-wong 

Is that possible you provide the cloud bundle?
So we can base on the bundle to reproduce the issue.

Pjack:
We suport hacluster based on this [csv table](https://github.com/canonical/charmed-openstack-upgrader/blob/main/cou/utils/openstack_to_track_mapping.csv)

We can enhance the message to indicate the user to change the workflow version from latest/stable to correct one. 
Furthermore, it may happen on all the other charms that we can handle them together.

chanchiwai-ray:
After investigation, I think the real issue here is primarily due `latest/stable` showed in COU via libjuju as `stable`, and hence becoming an error in COU. If the channel is showed as `latest/stable` it should pass the COU sanity check, and COU will propose to upgrade it to the corresponding channel.

PR #529, opened 2024-08-26
Make `stable` as an alias to `latest/stable` channel

chanchiwai-ray:
Sometime (old) charm in `latest/stable` channel can showed up in COU as `stable` channel, which makes COU think that it's an unsupported channel. However, this is not correct, for `latest/channel`, COU can still support proposing a possible crossgrade channel. 

Close: #491


Pjack (approved):

samuelallan72 (approved):
Nice find, looks good to me :+1:

PR #532, opened 2024-08-29
Fix --purge / --purge-before-date dependency issue

chanchiwai-ray:
Currently, the dependency check only works if `--purge` is specified first. If the `--purge` is after `--purge-before-date`, the dependency check action still report missing require args `--purge`. For example, `cou plan --purge-before-date 2000-1-1 --purge` will fail.

This PR fixed that issue, making the order of `--purge` and `--purge-before-date` unimportant

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -128,23 +128,6 @@ def purge_before_arg(value: str) -> str:
     raise argparse.ArgumentTypeError("purge before format must be YYYY-MM-DD[HH:mm][:ss]")
 
 
-class PurgeBeforeArgumentAction(argparse.Action):
-    """Custom action to make sure the arguments dependency."""
-
-    required_arg = "purge"
-
-    def __call__(
-        self,
-        parser: argparse.ArgumentParser,
-        namespace: argparse.Namespace,
-        values: Any,
-        option_string: Optional[str] = None,
-    ) -> None:
-        if getattr(namespace, self.required_arg) is False:
-            parser.error(f"--{self.dest} requires --{self.required_arg}")
-        setattr(namespace, self.dest, values)
-
-
 def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
     """Create a shared parser for options specific to subcommands.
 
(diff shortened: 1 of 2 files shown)


Pjack (dismissed):

gabrielcocenza (dismissed):

cou/commands.py
```diff
@@ -185,8 +168,11 @@ def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
     )
     subcommand_common_opts_parser.add_argument(
         "--archive-batch-size",
-        help="Batch size for nova old database data archiving.\n"
-        "Decrease the batch size if performance issues are detected.\n(default: 1000)",
+        help=(
+            "Batch size for nova old database data archiving."
+            "\nDecrease the batch size if performance issues are detected.\n(default: 1000)"
+            "\nThis option requires --archive."
```
samuelallan72, line 174: Could you also add the logic to enforce this, copying the logic for purge/purge-before-date?

samuelallan72 (changes requested):
Could you improve the commit message and PR description wording?  It really confused me, because the wording and present tense suggested that the limitation applied to these changes.  But in fact this change fixes that limitation.

cou/commands.py
samuelallan72: (Alternatively, that could be done in another PR, but perhaps the changes to --archive help message should be out of scope here then.)

chanchiwai-ray: Ah, sorry, this was accidentally added. This option does not required `--archive`

cou/commands.py
```diff
@@ -555,6 +540,13 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
                     subparsers.choices["upgrade"].print_help()
             parser.exit()
 
+        # validate arguments
+        validation_errors = []
+        if parsed_args.purge_before and not getattr(parsed_args, "purge", None):
```
samuelallan72, line 545:
```suggestion
        if parsed_args.purge_before and not parsed_args.purge:
```

cou/commands.py
```diff
@@ -555,6 +540,13 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
...
+            validation_errors.append("--purge-before-date requires --purge")
+        if len(validation_errors) > 0:
```
samuelallan72, line 547:
```suggestion
        if validation_errors:
```

samuelallan72 (changes requested):
Thanks @chanchiwai-ray , this looks pretty good.  I left some code lint comments though. :)

samuelallan72 (approved):

gabrielcocenza (approved):

cou/commands.py
chanchiwai-ray: Fixed

cou/commands.py
chanchiwai-ray: Fixed

PR #541, opened 2024-09-23
Remove 2023.02 (bobcat) from COU

chanchiwai-ray:


jneo8 (dismissed):

tests/unit/utils/test_openstack.py
```diff
@@ -289,7 +282,6 @@ def test_openstack_release_setter():
 def test_openstack_release_setter_by_date():
     openstack_release = OpenStackRelease("2023.1")
     assert openstack_release.codename == "antelope"
-    assert openstack_release.next_release == "bobcat"
```
samuelallan72, line 292: Can we assert that the next release is caracal here?

chanchiwai-ray: Thanks, it's now tested in `test_plan.py`

jneo8 (approved):

samuelallan72 (approved):
Thanks, looks good :D

gabrielcocenza:
I just had time to see this now. I'm not completely sure that this is the right approach once the community as a general thing can have a cloud running on bobcat and then COU will not support upgrading it. IMO the ideal scenario would be:

- if COU detect a cloud on Antelop, upgrade directly to Caracal
- if COU detect a cloud on Bobcat, upgrade to Caracal or in the worst scenario identify that the cloud is on Bobcat and say that is not supported and manual intervention is needed

If I understood this correctly COU will break if a Bobcat cloud is passed right?

cc: @aieri

samuelallan72:
@gabrielcocenza we've discussed this in APAC and the conclusion has been that we simply will not support a cloud on Bobcat.  Afaik this means worst case it will crash or have undefined operation, best case is identify Bobcat and stop with a message about unsupported cloud version.

chanchiwai-ray:
Thanks @samuelallan72 . 

@gabrielcocenza Yes, this is raised by me already, and we've confirmed that we (APAC) accept the risk. Please refer to the internal ticket for context.

Issue #457, opened 2024-06-27
Log path is lost in the scrollback


samuelallan72:
The path to COU's log file changes for every execution of COU.  The path is output as the first line, which is helpful to find the log file.  However, I've observed this situation:

```
$ cou upgrade
Full execution log: '/home/ubuntu/.local/share/cou/log/cou-20240627051952.log'
... 1000s of lines

Traceback / interrupt / finished execution / etc.
$ # ... uh oh, where's the log path?
```

Then it's very slow or impossible to scroll back to find the log file path.

Ideally it would be helpful if cou printed the log file path again at the end.

rgildein:
> Then it's very slow or impossible to scroll back to find the log file path.

Since you will run one COU run after another and never two at the same time, you can simply check the newest file in ` ~/.local/share/cou/log`.

PR #542, opened 2024-09-23
feat: Print log path in the end of output

jneo8:
Move log file path information to the end of cli output

Example output:

```
Full execution log: '/home/jneo8/.local/share/cou/log/cou-20240924113904.log'  
Configuring logging... ✖
...
...
...
Full execution log: '/home/jneo8/.local/share/cou/log/cou-20240924113904.log'
```

### DIFF
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -18,13 +18,15 @@ import logging
 import logging.handlers
 import sys
 from enum import Enum
+from pathlib import Path
 from signal import SIGINT, SIGTERM
+from typing import Optional
 
 from juju.errors import JujuError
 
 from cou.commands import CLIargs, parse_args
 from cou.exceptions import COUException, HighestReleaseAchieved, TimeoutException
-from cou.logging import setup_logging
+from cou.logging import get_log_file, setup_logging
 from cou.steps import UpgradePlan
 from cou.steps.analyze import Analysis
 from cou.steps.execute import apply_step
@@ -245,11 +247,13 @@ async def _run_command(args: CLIargs) -> None:
 def entrypoint() -> None:
     """Execute 'charmed-openstack-upgrade' command."""
     args = parse_args(sys.argv[1:])
+    log_file: Optional[Path] = None
     try:
         # disable progress indicator when in quiet mode to suppress its console output
         progress_indicator.enabled = not args.quiet
         log_level = get_log_level(quiet=args.quiet, verbosity=args.verbosity)
-        setup_logging(log_level)
+        log_file = get_log_file()
+        setup_logging(log_file, log_level)
 
         loop = asyncio.get_event_loop()
         loop.run_until_complete(_run_command(args))
@@ -291,4 +295,6 @@ def entrypoint() -> None:
     finally:
         if args.command == "upgrade":
             loop.run_until_complete(run_post_upgrade_sanity_check(args))
+        if log_file is not None:
+            print(f"Full execution log: '{log_file}'")
         progress_indicator.stop()
--- a/cou/cli.py
+++ b/cou/cli.py
@@ -295,6 +295,6 @@ def entrypoint() -> None:
     finally:
         if args.command == "upgrade":
             loop.run_until_complete(run_post_upgrade_sanity_check(args))
-        if log_file is not None:
+        if log_file is not None and not args.quiet:
             print(f"Full execution log: '{log_file}'")
         progress_indicator.stop()
--- a/cou/logging.py
+++ b/cou/logging.py
@@ -16,6 +16,7 @@
 import logging
 import logging.handlers
 from datetime import datetime
+from pathlib import Path
 
 from cou.utils import COU_DATA, progress_indicator
 
@@ -39,15 +40,25 @@ class TracebackInfoFilter(logging.Filter):  # pylint: disable=too-few-public-met
         return True
 
 
-def setup_logging(log_level: str = "INFO") -> None:
+def get_log_file() -> Path:
+    """Get log file path.
+
+    :return: Returns log file path
+    :rtype: Path
+    """
+    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
+    return Path(f"{COU_DIR_LOG}/cou-{time_stamp}.log")
+
+
+def setup_logging(log_file: Path, log_level: str = "INFO") -> None:
     """Do setup for logging.
 
+    :param log_file: Logging file.
+    :type log_level: Path
     :param log_level: Logging level, defaults to "INFO"
     :type log_level: str, optional
     """
     progress_indicator.start("Configuring logging...")
-    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
-    file_name = f"{COU_DIR_LOG}/cou-{time_stamp}.log"
     COU_DIR_LOG.mkdir(parents=True, exist_ok=True)
 
     log_formatter_file = logging.Formatter(
@@ -60,7 +71,7 @@ def setup_logging(log_level: str = "INFO") -> None:
     root_logger.setLevel("NOTSET")
 
     # handler for the log file. Log level is "NOTSET"
-    log_file_handler = logging.FileHandler(file_name)
+    log_file_handler = logging.FileHandler(log_file)
     log_file_handler.setFormatter(log_formatter_file)
     # suppress python libjuju and websockets debug logs
     if log_level != "NOTSET":
@@ -78,7 +89,7 @@ def setup_logging(log_level: str = "INFO") -> None:
     root_logger.addHandler(log_file_handler)
     root_logger.addHandler(console_handler)
 
-    progress_indicator.stop_and_persist(text=f"Full execution log: '{file_name}'")
+    progress_indicator.stop_and_persist(text=f"Full execution log: '{log_file}'")
 
 
 def filter_debug_logs(record: logging.LogRecord) -> bool:
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -320,9 +320,12 @@ async def test_run_command(
         mock_apply_upgrade_plan.assert_awaited_once()
 
 
+@patch("cou.cli.print")
+@patch("cou.cli.progress_indicator")
 @patch("cou.cli.run_post_upgrade_sanity_check")
 @patch("cou.cli.sys")
 @patch("cou.cli.parse_args")
+@patch("cou.cli.get_log_file")
 @patch("cou.cli.get_log_level")
 @patch("cou.cli.setup_logging")
 @patch("cou.cli._run_command")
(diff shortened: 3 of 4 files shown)


cou/logging.py
```diff
@@ -78,8 +89,6 @@ def setup_logging(log_level: str = "INFO") -> None:
     root_logger.addHandler(log_file_handler)
     root_logger.addHandler(console_handler)
 
-    progress_indicator.stop_and_persist(text=f"Full execution log: '{file_name}'")
```
samuelallan72, line 81: Can we retain printing the log file path at the beginning too?  Then the user can see the log file path to find it while cou is still running if they desire.

cou/cli.py
```diff
@@ -291,4 +292,6 @@ def entrypoint() -> None:
     finally:
         if args.command == "upgrade":
             loop.run_until_complete(run_post_upgrade_sanity_check(args))
+        if log_file:
```
samuelallan72, line 295: I think this conditional can be removed; `log_file` will always be populated.

samuelallan72 (changes requested):

cou/logging.py
jneo8: Ok, nice idea.

cou/cli.py
jneo8: No, it's defensive programming. We shouldn't trust the line before `log_file = get_log_file()`.

samuelallan72: Oh sorry my bad, I didn't realise that was in the try block.  Can we move the logging setup to before the try block?  Ideally we'd only need `loop.run_until_complete(_run_command(args))` in the try block.

But that can be out of scope for this PR if you'd prefer.

samuelallan72: Actually, if it crashes before `log_file = ...`, `log_file` won't even be declared, so the `if log_file` will crash.  We'd need to at least declare log_file before the `try`.

cou/logging.py
```diff
@@ -39,15 +40,25 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+
+def setup_logging(log_file: Path, log_level: str = "INFO") -> None:
     """Do setup for logging.
 
+    :param log_file: Logging file, defaults to "{COU_DIR_LOG}/cou-{time_stamp}.log"
+    :type log_level: Path
     :param log_level: Logging level, defaults to "INFO"
     :type log_level: str, optional
     """
     progress_indicator.start("Configuring logging...")
```
rgildein, line 61: In this line we start the indicator with message: "Configuring logging ..." and you remove part, when we stop it. Can you put it back?

cou/cli.py
rgildein: Since get_log_file return always Path and never None, I do not see reason for using `if log_file`. 
Also, I do not think that this is good approach start indicator at one function and end it here, where it can "stop" some other "start".

rgildein (commented):
Can you provide example output to description, thanks?

cou/cli.py
jneo8: > Can we move the logging setup to before the try block? Ideally we'd only need loop.run_until_complete(_run_command(args)) in the try block.

It's out of the scope.

jneo8: > Since get_log_file return always Path and never None, I do not see reason for using if log_file.

What if the line, `log_level = get_log_level(...)` failed? You can't suppose the lines before we assign the value is always no exception.

jneo8: > Actually, if it crashes before log_file = ..., log_file won't even be declared, so the if log_file will crash. We'd need to at least set log_file to something falsey before the try

Yes, this one is missing. I will update this part.

jneo8: > Also, I do not think that this is good approach start indicator at one function and end it here, where it can "stop" some other "start".

Maybe we can just use `print`

cou/cli.py
```diff
@@ -291,4 +292,6 @@ def entrypoint() -> None:
...
+            progress_indicator.stop_and_persist(text=f"Full execution log: '{log_file}'")
```
chanchiwai-ray, line 296: We should not rely on `progress_indicator` for printing text: it can be disabled with `cou -q`

chanchiwai-ray (changes requested):

cou/logging.py
jneo8: Added back, so resolved.

cou/cli.py
jneo8: Change to `print`

cou/logging.py
```diff
@@ -39,15 +40,25 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    :rtype: Path
+    """
+    time_stamp = datetime.now().strftime("%Y%m%d%H%M%S")
+    return Path(f"{COU_DIR_LOG}/cou-{time_stamp}.log")
+
+
+def setup_logging(log_file: Path, log_level: str = "INFO") -> None:
     """Do setup for logging.
 
+    :param log_file: Logging file, defaults to "{COU_DIR_LOG}/cou-{time_stamp}.log"
```
samuelallan72, line 56: I see no default here

```suggestion
    :param log_file: Logging file"
```

samuelallan72 (approved):
Thanks, looks good to me :)

I left a nit about the docs, but otherwise seems fine to me.

cou/cli.py
rgildein: > What if the line, log_level = get_log_level(...) failed? You can't suppose the lines before we assign the value is always no exception.

I see that you add `log_file: Optional[Path] = None`, so now it makes more sense.

chanchiwai-ray (commented):
Can you show some sample output

samuelallan72:
> refactor: Print log path in the end of output

small nit: this isn't a refactor, because it changes how the software works (log path is now printed at the end of the output).

jneo8:
> Can you show some sample output

Add in description

jneo8:
> > refactor: Print log path in the end of output
> 
> small nit: this isn't a refactor, because it changes how the software works (log path is now printed at the end of the output).

Squash and rebase

chanchiwai-ray (commented):
The duplicated line looks like a bug, we should fix it at some point

chanchiwai-ray (dismissed):

samuelallan72:
> The duplicated line looks like a bug, we should fix it at some point

The duplicated line (about the logfile path) is the feature we added in this PR. :)  The aim is to print it at the beginning and the end of the output, so the user can quickly find the log file.

cou/cli.py
samuelallan72, line 295: @jneo8 I remember you added `log_file = None` before the try block; what happened to that?  That will be required here.

samuelallan72 (changes requested):

cou/cli.py
jneo8: Maybe something lost when rebasing, let me change it.

chanchiwai-ray:
> The duplicated line (about the logfile path) is the feature we added in this PR. :) The aim is to print it at the beginning and the end of the output, so the user can quickly find the log file.

I know it's added by this PR. I think we should keep only one of them

cou/cli.py
jneo8: Updated, resolved

chanchiwai-ray:
> Keep both will be useful when the content output is long and take times, otherwise you won't able to know which log you are writing to when in the middle of processing.

If it's long and take times, you still can't see the log filename because it's in the `finally` block

cou/cli.py
```diff
@@ -291,4 +295,6 @@ def entrypoint() -> None:
     finally:
         if args.command == "upgrade":
             loop.run_until_complete(run_post_upgrade_sanity_check(args))
+        if log_file is not None:
+            progress_indicator.stop_and_persist(text=f"Full execution log: '{log_file}'")
```
samuelallan72, line 299: Missed this before - this should also be a plain `print`, not the progress_indicator. Maybe a few things were lost in the rebase? :sweat_smile:

samuelallan72 (changes requested):

cou/cli.py
jneo8: Updated

rgildein:
> Keep both will be useful when the content output is long and take times, otherwise you won't able to know which log you are writing to when in the middle of processing.

It's using time, so I can simply open/check latest file. I do not see any benefit of this approach.
Also, will this print path to log file even if I use `--quiet`, I think yes and it should not.

jneo8:
> > Keep both will be useful when the content output is long and take times, otherwise you won't able to know which log you are writing to when in the middle of processing.
> 
> It's using time, so I can simply open/check latest file. I do not see any benefit of this approach. Also, will this print path to log file even if I use `--quiet`, I think yes and it should not.

https://github.com/canonical/charmed-openstack-upgrader/issues/457#issuecomment-2193829076

@Pjack Do you want to block this one?

samuelallan72:
> It's using time, so I can simply open/check latest file. I do not see any benefit of this approach.

> Since you will run one COU run after another and never two at the same time, you can simply check the newest file in  ~/.local/share/cou/log.

@rgildein this is fine and easy for someone familiar with COU.  However a new user doesn't know the logs are under .local/share/cou/log, and doesn't know the convention of logs being timestamped.  Also, it's much easier to copy/paste the log file path printed from COU then to search for the newest file there. I don't see a downside of printing the log file again at the end of the output - it's just friendly UX.

If you're worried about the `--quiet` flag, we can add `if not quiet: print(log file message` at the end instead.

samuelallan72 (approved):

chanchiwai-ray (approved):
I don't want to block this small feature, but I don't think it will improve COU UI overall.

> If it's long and take times, you still can't see the log filename because it's in the finally block

samuelallan72:
> If it's long and take times, you still can't see the log filename because it's in the finally block

I don't understand.  This PR still retains printing the log filename at the beginning too.  So we have:

- printed at the beginning - the user is immediately aware of the log, can copy/paste and tail the log if they need (while COU is running)
- printed at the end - the user is reminded of the log, and can immediately copy/paste the log if they want to go back and look at the log.

samuelallan72 (approved):

chanchiwai-ray (approved):

rgildein:
> @rgildein this is fine and easy for someone familiar with COU. However a new user doesn't know the logs are under .local/share/cou/log, and doesn't know the convention of logs being timestamped. Also, it's much easier to copy/paste the log file path printed from COU then to search for the newest file there. I don't see a downside of printing the log file again at the end of the output - it's just friendly UX.
> 
> If you're worried about the `--quiet` flag, we can add `if not quiet: print(log file message` at the end instead.


@samuelallan72, If that true, we should provide better documentation. I see your points, and I was not thinking that this change should be blocked. 
IMO, this does not improve user experience, because if I want to see plan and I run `cou plan` I would expect the last thing in terminal to be the plan and not log message, but I can run with `--quite`.

PR #543, opened 2024-09-23
docs: Doc to provide a script to watch the cou logs

jneo8:


docs/how-to/watch-cou-logs.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Below is the script to watch the COU logs when doing the development.
+This script watches every log file in the directory, even newly created files.
+
+.. code:: bash
+
+    #!/bin/bash
+
+    # Change LOG_DIR to the target directory
+    DIRECTORY="$LOG_DIR"
```
samuelallan72, line 13: Isn't this a standard location?  Can we hardcode this to `~/.local/share/cou/logs/*`, so by default users who copy/paste the script don't need to make changes?

samuelallan72 (changes requested):
Thanks, looks good!  Just a comment about the var that needs changing manually.

docs/how-to/watch-cou-logs.rst
jneo8: Updated

docs/how-to/watch-cou-logs.rst
```diff
@@ -0,0 +1,48 @@
...
+
+Below is the script to watch the COU logs when doing the development.
+This script watches every log file in the directory, even newly created files.
+
+.. code:: bash
+
+    #!/bin/bash
+
+    # Change LOG_DIR to the target directory
+    DIRECTORY="~/.local/share/cou/log/"
```
samuelallan72, line 13: I tried this, and it didn't work with the tilde home shortcut unfortunately.

```suggestion
    DIRECTORY="$HOME/.local/share/cou/log/"
```

docs/how-to/watch-cou-logs.rst
samuelallan72: Ah sorry I realised the path I suggested wasn't correct, but you got it right, thanks.

docs/how-to/watch-cou-logs.rst
jneo8: Updated

samuelallan72 (approved):

chanchiwai-ray (approved):

Issue #547, opened 2024-09-26
[Bug] Should pass apps_control_plane to archive step 


jneo8:
The bug happen on below lines:

https://github.com/canonical/charmed-openstack-upgrader/blob/110e6f2f24746234754893047aa555354e42b442/cou/steps/plan.py#L579-L589

It should be:

```python
    if args.archive:
        return [
            PreUpgradeStep(
                description="Archive old database data on nova-cloud-controller",
                coro=archive(
                    analysis_result.model,
                    # Should be apps_control_plane
                    analysis_result.apps_control_plane,
                    batch_size=args.archive_batch_size,
                ),
            )
        ]
 ```

chanchiwai-ray:
Regression from https://github.com/canonical/charmed-openstack-upgrader/pull/534

PR #550, opened 2024-09-30
Fix archive step

samuelallan72:
nova-cloud-controller is in the control plane group, not data plane. The wrong group was being passed,
so archive could not found the nova-cloud-controller units.

Fixes: #547


samuelallan72:
I tested and confirm this fixed the issue for me. :)

gabrielcocenza (approved):

jneo8 (approved):

PR #551, opened 2024-09-30
Fix upgrade order for subordinate charms

samuelallan72:
All subordinate charms must be upgraded before principal charms.

This removes the specific logic for ovn-chassis (a subordinate charm),
including it in the new generic logic for subordinate charms.

Fixes: #552
Fixes: #553


samuelallan72:
I haven't updated the tests yet, but I have updated the mocked plans.  Please take an early look at the approach here and the resulting plan.  I think it helps make the upgrade order more correct (see linked isues), without changing the structure of cou too much.  Thoughts?

cou/steps/analyze.py
```diff
@@ -240,41 +196,19 @@ def _get_minimum_cloud_series(self) -> Optional[str]:
...
         }
-
-    @property
-    def machines(self) -> dict[str, juju_utils.Machine]:
-        """All OpenStack machines of the model.
-
-        :return: All OpenStack machines of the model.
-        :rtype: dict[str, Machine]
-        """
-        return {**self.data_plane_machines, **self.control_plane_machines}
```
samuelallan72, line 280: Note that there's also some other refactoring along the way - I tend to do these kind of improvements while making other changes and trying to understand the code.  I can split them out of the PR once it's ready.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
```diff
@@ -2,17 +2,13 @@ plan: |
...
-            Upgrade plan for 'ovn-chassis' to 'victoria'
-                WARNING: Changing 'ovn-chassis' channel from latest/stable to 22.03/stable. This may be a charm downgrade, which is generally not supported.
-                Wait for up to 300s for app 'ovn-chassis' to reach the idle state
         Control Plane subordinate(s) upgrade plan
             Upgrade plan for 'barbican-vault' to 'victoria'
                 Upgrade 'barbican-vault' from 'ussuri/stable' to the new channel: 'victoria/stable'
                 Wait for up to 300s for app 'barbican-vault' to reach the idle state
+            Upgrade plan for 'ceilometer-agent' to 'victoria'
+                Upgrade 'ceilometer-agent' from 'ussuri/stable' to the new channel: 'victoria/stable'
+                Wait for up to 300s for app 'ceilometer-agent' to reach the idle state
```
jneo8, line 11: **question:** 

This is a little bit weird for me, according to the definition of data plane. `ceilometer-agent` should belong to data-plane subordinate group. But it's under Control plane subordinates now.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
```diff
@@ -163,6 +159,9 @@ plan: |
             Upgrade plan for 'octavia-ovn-chassis' to 'victoria'
                 WARNING: Changing 'octavia-ovn-chassis' channel from latest/stable to 22.03/stable. This may be a charm downgrade, which is generally not supported.
                 Wait for up to 300s for app 'octavia-ovn-chassis' to reach the idle state
+            Upgrade plan for 'ovn-chassis' to 'victoria'
+                WARNING: Changing 'ovn-chassis' channel from latest/stable to 22.03/stable. This may be a charm downgrade, which is generally not supported.
+                Wait for up to 300s for app 'ovn-chassis' to reach the idle state
```
jneo8, line 164: **nitpick:** 

The same issue, the `ovn-chassis` shouldn't belong to Control plane subordinates.

jneo8 (changes requested):
I think the applications, ceilometer-agent and ovn-chassis, should not belong to Control plane subordinates. Can we create a new group for them?

samuelallan72:
> I think the applications, ceilometer-agent and ovn-chassis, should not belong to Control plane subordinates. Can we create a new group for them?

for simplicity I'd really prefer not to. Maybe we can rename "control-plane" group to something else instead. I don't want grouping to cause overly-complex code or suboptimal upgrade order.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
samuelallan72: There is no data-plane subordinate group now; only subordinate applications. Maybe we could rename the group? What do you think?

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
samuelallan72: Previously ovn-chassis was put in it's own group, which was confusing and made the code unecessarily complex.

jneo8: The reason why we need this is because ovn-chassis needs to upgrade before ovn-central.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
jneo8: I remember we discuss this before. The upgrade order should be:

`Control plane subordinate` -> `Control plane` -> `Data plane subordinate` -> `Data plane`

So we need another group.

samuelallan72: What's wrong with `subordinates` -> `control plane` -> `data plane`? That would simplify things - see internal discussion [here](https://chat.canonical.com/canonical/pl/1t7fugscdfb4pnm9rnc4ie6uzw).

I mean, we could follow your proposed upgrade order, but that makes things more complicated again.

Note that currently the upgrade order in COU is `contral plane subordinate` -> `control plane` -> `nova-compute` -> `data plane principals` -> `data plane subordinate`.

jneo8: > What's wrong with subordinates -> control plane -> data plane?

I think the point is we shouldn't touch anything on data plane before we finish the upgrade on control-plane. The change you need to do is move the whole plane to:

`OVN-chassis` -> `control plane subordinate` -> control plane -> data plane subordinate -> nova-compute -> data plane principals

cou/steps/analyze.py
```diff
@@ -240,41 +196,19 @@ def _get_minimum_cloud_series(self) -> Optional[str]:
...
-        :return: Control-plane machines of the model.
+        :return: All OpenStack machines of the model.
         :rtype: dict[str, Machine]
         """
         return {
             machine_id: app.machines[machine_id]
-            for app in self.apps_control_plane
+            for app in self.apps
             for machine_id in app.machines
         }
```
gabrielcocenza, line 214: This is basically what juju_utils already does:

https://github.com/canonical/charmed-openstack-upgrader/blob/5bfa932bb459377c7311373d8a3284442a7afc20/cou/utils/juju_utils.py#L279-L294

I prefer passing a new class attribute  `machines` to the Analysis object instead of having kind of duplicate code

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
gabrielcocenza: We are kind of upgrading data-plane subordinates first because of [ovn-chassis](https://github.com/canonical/charmed-openstack-upgrader/pull/518). I don't see a strong reason to keep this separation and increase the business logic inside COU.

I think Samuel's changes is just simplifying this to see subordinates as a group.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
gabrielcocenza: nova-compute (data-plane) and octavia (control-plane) has ovn-chassis. I'm not seeing a good reason to not have subordinates as a group once they are already mixed

cou/steps/analyze.py
samuelallan72: Ah so this was kind of a refactoring - there was already this method and this logic, along with other methods to get control plane machines and data plane machines.  They weren't really being used, so I factored them out and simplified this method.  Maybe I should revert the refactoring and do it in another PR. Thanks for flagging the duplication; even more that can be factored out. :)

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
samuelallan72: Group heading updated

cou/steps/analyze.py
samuelallan72: I've removed this refactoring

chanchiwai-ray (approved):
:+1: simplify the logic if upgrade still works

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
jneo8: Ok, let's make it easier. Thanks for explanation.

tests/mocked_plans/sample_plans/9eb9af6a-b919-4cf9-8f2f-9df16a1556be.yaml
jneo8: Thanks for explanation.

jneo8 (approved):

PR #557, opened 2024-10-01
Update mapping for ceph on caracal

samuelallan72:
According to https://ubuntu.com/ceph/docs/supported-ceph-versions , Ceph Squid release must be used with OpenStack Caracal.

WARNING: at this time I'm not sure if ceph charms will support upgrading directly from quincy to squid, so this may require more work (we do a single step antelope -> caracal openstack upgrade, but ceph charms may require two steps: quincy -> reef -> squid).


gabrielcocenza (approved):
Probably it will be necessary to upgrade from quincy -> reef -> squid as you said.

samuelallan72:
But we should still merge this, because this sets the correct release of ceph for caracal.  Separately we'll need to re-add bobcat support.

aieri (approved):

Issue #559, opened 2024-10-02
[Bug] Wait for model idle stuck forever if there are zero unit subordinate application


jneo8:
In the case: 

If there are subordinate application which not relate to any principle, which will have zero units, and we call [juju model wait_for_idle function](https://github.com/juju/python-libjuju/blob/50b42d013aee01536416e6334f99443f2b4f1e4c/juju/model.py#L2792) with no application filter, it will stuck forever because the check will never pass. 

Relate to: #549

samuelallan72:
Would it make sense to set wait_for_at_least_units=0 to wait_for_idle? This would be a quick fix and I reckon would solve it for us.

It would be nice to add a step to the initial verify_cloud step too, to check for applications that have zero units.

Issue #549, opened 2024-09-27
[Bug] Unnecessary step for supported app with zero units.


jneo8:
In the case that I have `cinder-ceph` application on the juju model, which is a subordinate charm but not relate to any principle so the unit number is zero. COU will still wait for it to become idle status but it's not possible. I suggest COU skip this unnecessary step.

samuelallan72:
Would it make sense to set wait_for_at_least_units=0 to wait_for_idle? This would be a quick fix and I reckon would solve it for us.

It would be nice to add a step to the initial verify_cloud step too, to check for applications that have zero units.

samuelallan72:
If we add `wait_for_at_least_units=0` to the `wait_for_idle` calls, this is not an issue - see https://github.com/canonical/charmed-openstack-upgrader/pull/564 .

PR #564, opened 2024-10-03
Fix waiting for idle when app has no units

samuelallan72:
Fixes: #559
Fixes: #549

### DIFF
--- a/cou/utils/juju_utils.py
+++ b/cou/utils/juju_utils.py
@@ -699,6 +699,7 @@ class Model:
                             raise_on_blocked=raise_on_blocked,
                             raise_on_error=raise_on_error,
                             status=status,
+                            wait_for_at_least_units=0,  # don't hang if an app has no units
                         )
                         for app in apps
                     )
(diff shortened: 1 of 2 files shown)


cou/utils/juju_utils.py
```diff
@@ -699,6 +699,7 @@ async def _wait_for_idle() -> None:
                             raise_on_blocked=raise_on_blocked,
                             raise_on_error=raise_on_error,
                             status=status,
+                            wait_for_at_least_units=0,  # don't hang if an app has no units
```
chanchiwai-ray, line 702: I wonder if we have been doing this erroneously: by default this is waiting for only 1 unit, so if we have more than one units for this `app`, then only one unit is in idle is enough

samuelallan72: So I think checking for number of units is independent of the status check. The logic seems to be:

1. check number of units
2. for each unit:
  3. check unit status

chanchiwai-ray (approved):

Pjack (approved):

PR #569, opened 2024-10-03
Remove extra $ from artefact name

samuelallan72:
This seems to have been a copy/paste error,
where our terraform templates use the double $$ to escape the $.


chanchiwai-ray (approved):

Pjack (approved):

jneo8 (approved):

samuelallan72:
Oops, I was too hasty on merging; func test workflow hadn't finished. :sweat_smile: Happily, the snap upload/download seems fine and is using the expected naming. :sweat_smile:

PR #570, opened 2024-10-03
Tidy check workflow naming and graph

samuelallan72:
Some usability improvements on the check workflow:
- shorter names so they aren't truncated by the github ui
- split unit and lint tests
- only build the snap if lint, unit, and mocked-plans succeed


chanchiwai-ray (approved):
Thanks 👍🏼

Pjack (approved):

jneo8 (approved):

Issue #460, opened 2024-06-27
Improve docs for `enable-version-pinning` config constraint


samuelallan72:
I observed this on a cloud with `enable-version-pinning` set to true for octavia-ovn-chassis and ovn-chassis.  COU plan/upgrade emits this error output:

```
Cannot generate plan for 'octavia-ovn-chassis'
        Cannot upgrade 'octavia-ovn-chassis'. 'enable-version-pinning' must be set to 'false'.
Cannot generate plan for 'ovn-chassis'
        Cannot upgrade 'ovn-chassis'. 'enable-version-pinning' must be set to 'false'.
```

I couldn't find the reason why it must be set to false, and I couldn't find any recommendation for what to do now (is it safe to set the config to false on a production cloud?).

We should improve the docs here.

Pjack:
An idea is to implement a `cou lint` command to check the environment and print out detail information.

PR #575, opened 2024-10-07
Improve error message for enable version pinning

chanchiwai-ray:
Add more context on why COU is requiring version pinning set to `False` for OVN charms.

Closes: #460

### DIFF
--- a/tests/unit/apps/test_auxiliary.py
+++ b/tests/unit/apps/test_auxiliary.py
@@ -1031,6 +1031,7 @@ def test_ovn_version_pinning_principal(model):
     target = OpenStackRelease("victoria")
     charm = "ovn-dedicated-chassis"
     machines = {"0": generate_cou_machine("0", "az-0")}
+    exp_regex_msg = ".*enable-version-pinning.*"
     app = OVNPrincipal(
         name=charm,
         can_upgrade_to="22.03/stable",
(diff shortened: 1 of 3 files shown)


jneo8 (approved):

tests/unit/apps/test_auxiliary.py
```diff
@@ -1151,7 +1160,16 @@ def test_ovn_check_version_pinning_version_pinning_config_True(model):
         },
         workload_version="22.03",
     )
-    exp_msg = f"Cannot upgrade '{app.name}'. 'enable-version-pinning' must be set to 'false'."
+    exp_msg = (
```
Pjack, line 1163: I’m fine with removing the message checks in the unit tests. 
It's unnecessary and annoyed to verify the entire message content.

jneo8: (non-blocking, free to resolve)

For me it's because the ApplicationError is too vague. How about create a new Error and put the messge there?

gabrielcocenza (dismissed):

tests/unit/apps/test_auxiliary.py
samuelallan72: My suggestion would be to retain the message check, but reduce it to a regex like `.*enable-version-pinning.*`, to at least assert that the error is related to the version pinning. Otherwise it could be crashing for another reason and the tests would not know.

chanchiwai-ray: - new error would be the best but highest cost
- regex would be good but still not solid
- removing it would be the easy but not good

It's fair to choose the middle one 👍🏼

samuelallan72 (approved):
Thanks, LGTM

jneo8 (approved):

tests/unit/apps/test_auxiliary.py
chanchiwai-ray: > If the message content is important, I’d prefer to match specific keywords, which should be sufficient.

Done

Issue #442, opened 2024-06-14
Improve testing for invalid args


samuelallan72:
Currently the tests for invalid args look like:

```
    with pytest.raises(SystemExit, match="2"):
        commands.parse_args(args)
```

But we aren't testing the error message (which is tricky currently without refactoring because the error message goes straight to stderr).  So this could be failing for any reason, not just the reason we think we're testing - eg. it could be failing for a reason like a typo in the test args, etc.  For example this test has invalid args not because the options are mutually exclusive, but because a subcommand is missing:

https://github.com/canonical/charmed-openstack-upgrader/blob/837696c946bd3d4ae6d504cac4f99ca13ee41e3d/tests/unit/test_commands.py#L58-L69

PR #580, opened 2024-10-08
Improve parse_args tests

samuelallan72:
Make the tests more precise, so we can be more confident
that the tests are verifying the expected cases.

Also fix a bug in datetime format parsing in purge_before_arg().

Fixes: #442

### DIFF
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -541,8 +541,6 @@ def parse_args(args: list[str]) -> CLIargs:  # pylint: disable=inconsistent-retu
     if len(args) == 0 or (len(args) == 1 and args[0] == "help"):
         parser.print_help()
         parser.exit()
-        # This line is reachable in unit tests when .exit() is mocked.
-        return  # type: ignore[unreachable]
 
     try:
         parsed_args = CLIargs(**vars(parser.parse_args(args)))
@@ -555,8 +553,6 @@ def parse_args(args: list[str]) -> CLIargs:  # pylint: disable=inconsistent-retu
                 case "upgrade":
                     subparsers.choices["upgrade"].print_help()
             parser.exit()
-            # This line is reachable in unit tests when .exit() is mocked.
-            return  # type: ignore[unreachable]
 
         # validate arguments
         validation_errors = []
@@ -564,8 +560,6 @@ def parse_args(args: list[str]) -> CLIargs:  # pylint: disable=inconsistent-retu
             validation_errors.append("--purge-before-date requires --purge")
         if validation_errors:
             parser.error("\n" + "\n".join(validation_errors))
-            # This line is reachable in unit tests when .error() is mocked.
-            return  # type: ignore[unreachable]
 
         return parsed_args
     except argparse.ArgumentError as exc:
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -118,7 +118,7 @@ def purge_before_arg(value: str) -> str:
     :rtype: str
     :raises argparse.ArgumentTypeError: if string format is invalid
     """
-    formats = ["%Y-%m-%d %H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%d"]
+    formats = ["%Y-%m-%d%H:%M:%S", "%Y-%m-%d%H:%M", "%Y-%m-%d"]
     for fmt in formats:
         try:
             datetime.strptime(value, fmt)
@@ -502,14 +502,17 @@ class CLIargs:
         return not self.auto_approve
 
 
-def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-statements
+def parse_args(args: list[str]) -> CLIargs:  # pylint: disable=inconsistent-return-statements
     """Parse cli arguments.
 
-    :param args: Arguments parser.
-    :type args: Any
+    Calls sys.exit via the argparse methods
+    if there are errors with the arguments,
+    or a help command is used.
+
+    :param args: List of arguments to parse
+    :type args: list[str]
     :return: CLIargs custom object.
     :rtype: CLIargs
-    :raises argparse.ArgumentError: Unexpected arguments input.
     """
     # Configure top level argparser and its options
     parser = argparse.ArgumentParser(
@@ -538,6 +541,8 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
     if len(args) == 0 or (len(args) == 1 and args[0] == "help"):
         parser.print_help()
         parser.exit()
+        # This line is reachable in unit tests when .exit() is mocked.
+        return  # type: ignore[unreachable]
 
     try:
         parsed_args = CLIargs(**vars(parser.parse_args(args)))
@@ -550,6 +555,8 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
                 case "upgrade":
                     subparsers.choices["upgrade"].print_help()
             parser.exit()
+            # This line is reachable in unit tests when .exit() is mocked.
+            return  # type: ignore[unreachable]
 
         # validate arguments
         validation_errors = []
@@ -557,6 +564,8 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
             validation_errors.append("--purge-before-date requires --purge")
         if validation_errors:
             parser.error("\n" + "\n".join(validation_errors))
+            # This line is reachable in unit tests when .error() is mocked.
+            return  # type: ignore[unreachable]
 
         return parsed_args
     except argparse.ArgumentError as exc:
--- a/cou/commands.py
+++ b/cou/commands.py
@@ -118,14 +118,14 @@ def purge_before_arg(value: str) -> str:
     :rtype: str
     :raises argparse.ArgumentTypeError: if string format is invalid
     """
-    formats = ["%Y-%m-%d%H:%M:%S", "%Y-%m-%d%H:%M", "%Y-%m-%d"]
+    formats = ["%Y-%m-%d %H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%d"]
     for fmt in formats:
         try:
             datetime.strptime(value, fmt)
             return value
         except ValueError:
             pass
-    raise argparse.ArgumentTypeError("purge before format must be YYYY-MM-DD[HH:mm][:ss]")
+    raise argparse.ArgumentTypeError("format must be YYYY-MM-DD[ HH:mm[:ss]]")
 
 
 def get_subcommand_common_opts_parser() -> argparse.ArgumentParser:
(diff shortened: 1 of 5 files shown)


cou/commands.py
```diff
@@ -538,6 +541,8 @@ def parse_args(args: Any) -> CLIargs:  # pylint: disable=inconsistent-return-sta
     if len(args) == 0 or (len(args) == 1 and args[0] == "help"):
         parser.print_help()
         parser.exit()
+        # This line is reachable in unit tests when .exit() is mocked.
+        return  # type: ignore[unreachable]
```
gabrielcocenza, line 545: I don't understand why are you adding a return here. I think this always will be unreachable with `parser.exit()`

samuelallan72: I added a comment above that line to explain:

> This line is reachable in unit tests when .exit() is mocked.

It was hard figuring out how best to test the function. I tried a few methods including various refactorings of the function. But the logic is all intertwined.

samuelallan72: If you have any suggestions for refactoring or improving testability, please let me know :)

cou/commands.py
```diff
@@ -118,7 +118,7 @@ def purge_before_arg(value: str) -> str:
     :rtype: str
     :raises argparse.ArgumentTypeError: if string format is invalid
     """
-    formats = ["%Y-%m-%d %H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%d"]
+    formats = ["%Y-%m-%d%H:%M:%S", "%Y-%m-%d%H:%M", "%Y-%m-%d"]
```
chanchiwai-ray, line 121: why remove the white space?

samuelallan72: Because without the whitespace is how it's documented:

> Date string format should be YYYY-MM-DD[HH:mm][:ss].

And it works better without whitespace for passing the value on the command line, otherwise you would need to quote it:

```
--purge-before-date "2024-01-01 01:01"
vs
--purge-before-date 2024-01-0101:01
```

It's more readable with the space, or even with the standard `T` separator (`2024-01-01T01:01`), but we have it documented without the space.

samuelallan72: CC @chanchiwai-ray @jneo8 @gabrielcocenza @rgildein it's probably not too late to change the format though - how would we prefer the datestamp here?

- "2024-01-01 01:01"
- 2024-01-0101:01
- 2024-01-01T01:01
- something else?

chanchiwai-ray: > Because without the whitespace is how it's documented:

Thanks, I assumed that was followed from the doc somewhere

 `"2024-01-01 01:01"`  is more natural to me

cou/commands.py
chanchiwai-ray: Adding `return` seems overkill. I also cannot find any error related to `.exit` being mocked

samuelallan72: I don't like it either, but if you remove the `return`, the unittests will fail. I'm happy for suggestions on how to improve the tests so we don't need this. Maybe a mock can raise systemexit, so it won't change the control flow but we can stil verify it was called? That function seems difficult to test. :/

samuelallan72: > Maybe a mock can raise systemexit

Aha, turns out this was key. I've reworked the tests, and the function no longer needs the return statements. :)

cou/commands.py
samuelallan72: I guess for now, to avoid expanding the scope of the PR too much, the question is: do we fix the code to match the docs, or fix the docs to match the code?

chanchiwai-ray: my guess it was just a typo in the doc, @jneo8 can confirm this

rgildein: Why do you want to do `--purge-before-date 2024-01-0101:01`? This is harder to understand.
I think that this will be mainly use with date, so I do not see any issue here.

samuelallan72: > Why do you want to do --purge-before-date 2024-01-0101:01? 

I don't, but all the COU docs say that's the format it uses.

tests/unit/test_commands.py
```diff
@@ -822,56 +829,72 @@ def test_parse_args_upgrade(args, expected_CLIargs):
...
+
+@patch("cou.commands.argparse.ArgumentParser.error")
+def test_parse_args_hypervisors_exclusive_options_reverse_order(mock_error):
+    """Test parsing mutually exclusive hypervisors specific options (reverse order)."""
+    mock_error.side_effect = SystemExit
+    with pytest.raises(SystemExit):
+        commands.parse_args(["upgrade", "hypervisors", "--az", "1", "-m", "2"])
+    mock_error.assert_called_once_with(
+        "argument --machine/-m: not allowed with argument --availability-zone/--az"
+    )
```
rgildein, line 851: Why we are mocking the `error` here and then catch what we set? The [error](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) function will raise SystemExit with return code 2 and we should check that instead of our mocked version.

tests/unit/test_commands.py
```diff
@@ -822,56 +829,72 @@ def test_parse_args_upgrade(args, expected_CLIargs):
...
 )
-def test_parse_invalid_args(args):
+@patch("cou.commands.argparse.ArgumentParser.error")
+def test_parse_invalid_args(mock_error, args):
     """Generic test for various invalid sets of args."""
-    with pytest.raises(SystemExit, match="2"):
+    mock_error.side_effect = SystemExit
+    with pytest.raises(SystemExit):
         commands.parse_args(args)
+    mock_error.assert_called()
```
rgildein, line 877: Why do you need to mock error also here? What value we will get from it like this?

samuelallan72: Here, I guess no real value. But at least it gives us opportunity to add value.

tests/unit/test_commands.py
samuelallan72: I had two main goals with this specific design:

1. have a more useful assert (previously it was asserting that SystemExit was raised, which could be caused by anything - case in point, some tests had typos, so the SystemExit was raised because of the function crashing on the typos, not because of the invalid condition that was supposed to be tested)
2. retain the original control flow of the function (if the mock doesn't raise SystemExit, then the function will continue instead of exiting). See https://github.com/canonical/charmed-openstack-upgrader/pull/580#discussion_r1791938631 for background.

cou/commands.py
samuelallan72: Ok, that seems like the consensus, thanks. I'll update the format to `YYYY-MM-DD[ HH:mm[:ss]]` to match the code. :+1:

tests/unit/test_commands.py
rgildein: For me we are simply losing value here, since we expect to raise something and just mocked it. If you want to check that it was called, I will rather use `wraps` so we can spy on object.

samuelallan72: > I will rather use wraps so we can spy on object.

This sounds interesting, could you explain more about what you mean or show links? I'm not aware of that.

tests/unit/test_commands.py
rgildein: > which could be caused by anything

We expect that the error call sys.exit(2), so we tested it. If there is any type the error will not be SystemExit, but something else.

> retain the original control flow of the function (if the mock doesn't raise SystemExit, then the function will continue instead of exiting).

You are mocking `error` function, which if we call it will raise an exception. So you do not need to mock it at all.

samuelallan72: > We expect that the error call sys.exit(2), so we tested it. If there is any type the error will not be SystemExit, but something else.

So for example, `test_parse_args_quiet_verbose_exclusive()` was an invalid test. It was designed to verify the quiet and verbose flags are mutually exclusive, however the args used to test were invalid for other reasons (missing the plan or upgrade subcommand). The test passed because the function raised SystemExit as expected, but that same test would still pass if we removed the mutually exclusive rules from quiet/verbose.

If we're going to test this case, we need a better assert.

tests/unit/test_commands.py
rgildein: You can check it [here](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock). The wraps can be used also with patch, so you will be able to check that it was called without mocking the output.

samuelallan72: Nice, thanks! I'll update the patching to use that. :+1: :+1:

samuelallan72: Hmm it doesn't look straight forward for using this for a class method, but I'll give it a shot.

samuelallan72: @rgildein have you used this kind of mock wrapping before? I can't get it to work.

rgildein:
```python
@patch("cou.commands.argparse.ArgumentParser.error", wraps=ArgumentParser.error, autospec=True)
def test_parse_args_quiet_verbose_exclusive(mock_error, args):
    """Test that quiet and verbose options are mutually exclusive."""
    with pytest.raises(SystemExit, match="2"):
        commands.parse_args(args)

    mock_error.assert_called_once_with(
        ANY, "argument --verbose/-v: not allowed with argument --quiet/-q"
    )
```

samuelallan72: ooh nice, `autospec` was what I was missing. That's kinda magic. Thanks for that :)

samuelallan72: Updated, ready for review again. :)

samuelallan72:
weird, unit tests pass for me locally, but fail in CI.

samuelallan72:
I can't reproduce the unit tests error from the CI. I tried in a jammy VM with python 3.10.12 and tox 4.21.2 - unit tests succeeded.

chanchiwai-ray:
> I can't reproduce the unit tests error from the CI. I tried in a jammy VM with python 3.10.12 and tox 4.21.2 - unit tests succeeded.

I can reproduce the error with
`python 3.10.12`
`tox == 4.18.0`

tests/unit/test_commands.py
samuelallan72: @chanchiwai-ray please read this thread: https://github.com/canonical/charmed-openstack-upgrader/pull/580#discussion_r1794772211

samuelallan72:
Oh my bad, I had the wrong branch checked out in the jammy VM. I can reproduce the unit test failure on jammy with python 3.10.12 and tox 4.21.2.

However the tests pass on noble with python 3.12.3 and tox 4.13.0.

tests/unit/test_commands.py
samuelallan72: @rgildein are you aware of any differences in the mocking lib between python 3.10 and 3.12 that could cause this to work on 3.12 but not 3.10?

samuelallan72: Ah, it's a bug in cpython: https://github.com/python/cpython/issues/75988

samuelallan72: Looks like it was backported to 3.11 and 3.12, but not 3.10. So we can't use this method here unfortunately unless we can find a workaround.

I propose dropping back to the earlier method: no `wraps`, and manually adding the SystemExit sideeffect.

samuelallan72: cc @rgildein @chanchiwai-ray ^

chanchiwai-ray (approved):

jneo8 (approved):

tests/unit/test_commands.py
samuelallan72: Resolving as I think the latest solution is the most reasonable option for now, given the comments and constraints we have in python 3.10. We can refactor later. :)

samuelallan72: @rgildein please let me know if you know of any better method? Otherwise we can wait until we switch to python >= 3.11 to use `wraps` with patch here.

Issue #561, opened 2024-10-02
[User Experience] If hypervisor upgrade failed and the user restarts COU, COU won't recognize that the user is in the middle of nova-compute upgrade.


jneo8:
In my case, the hypervisor upgrade failed due to an known issue #494 .
If I manually fix the issue on the machine and restart the COU, I will get error message like:

```
Analyzing cloud... ---- Logging error ---
Traceback (most recent call last):
  File "/home/jneo8/.rye/py/cpython@3.10.13/install/lib/python3.10/logging/__init__.py", line 1100, in emit
    msg = self.format(record)
  File "/home/jneo8/.rye/py/cpython@3.10.13/install/lib/python3.10/logging/__init__.py", line 943, in format
    return fmt.format(record)
  File "/home/jneo8/.rye/py/cpython@3.10.13/install/lib/python3.10/logging/__init__.py", line 678, in format
    record.message = record.getMessage()
  File "/home/jneo8/.rye/py/cpython@3.10.13/install/lib/python3.10/logging/__init__.py", line 368, in getMessage
    msg = msg % self.args
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/cou/apps/base.py", line 119, in __str__
    "units": {
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/cou/apps/base.py", line 124, in <dictcomp>
    "o7k_version": str(self.get_latest_o7k_version(unit)),
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/cou/apps/base.py", line 327, in get_latest_o7k_version
    compatible_o7k_versions = OpenStackCodenameLookup.find_compatible_versions(
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/cou/utils/openstack.py", line 431, in find_compatible_versions
    if version in version_range:
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/cou/utils/openstack.py", line 351, in __contains__
    service_version = Version(version)
  File "/home/jneo8/CanonicalProjects/charmed-openstack-upgrader/.venv/lib/python3.10/site-packages/packaging/version.py", line 202, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: 'zed'
Call stack:
[...]

PR #584, opened 2024-10-09
Check running ceph versions

samuelallan72:
Add a pre and post upgrade warning if running ceph versions are mismatched.

We can't error on these cases, because mismatched versions are expected if the upgrade is being resumed, but we can warn the user.

Fixes: #401


samuelallan72:
This is currently a work in progress, but I'm interested in reviews on the approach before continuing. This is untested code that implements a basic warning before and after an upgrade when ceph-mon reports running versions of ceph that are mismatched. See also https://github.com/canonical/charmed-openstack-upgrader/issues/401#issuecomment-2401401387 .

cou/steps/plan.py
```diff
@@ -160,6 +161,23 @@ async def post_upgrade_sanity_checks(analysis_result: Analysis) -> None:
     except ApplicationError:
         messages.append("Detected ceph 'noout' set, please unset noout for ceph cluster.")
 
+    for app in ceph.get_mon_apps(analysis_result.apps_control_plane):
```
chanchiwai-ray, line 164: (non-blocking) ideally, the same `verify_cloud` pattern can be used for this function, but that's more like a small refactoring:

```python
async def post_upgrade_sanity_checks():
    _post_verify_a()
    _post_verify_b()
```

cou/steps/ceph.py
```diff
@@ -266,3 +266,57 @@ async def set_require_osd_release_option(model: Model, apps: Sequence[Applicatio
             continue
 
         await set_require_osd_release_option_on_unit(model, ceph_mon_unit_name)
+
+
+def get_mon_apps(apps: Sequence[Application]) -> Sequence[Application]:
```
chanchiwai-ray, line 271: this can be either inline or used by other functions e.g. `set_require_osd_release_option` also gets ceph_mon apps

jneo8 (dismissed):

cou/steps/plan.py
samuelallan72: Agreed, let's leave such refactoring to a future PR

cou/steps/ceph.py
samuelallan72: Yes, I thought this function might be useful in other functions in the future, but that's also a refactoring that's unrelated to this change. :)

cou/steps/plan.py
```diff
@@ -278,6 +296,32 @@ async def _verify_vault_is_unsealed(analysis_result: Analysis) -> None:
         PlanStatus.add_message(str(e), MessageType.ERROR)
 
 
+async def _verify_ceph_running_versions_consistent(analysis_result: Analysis) -> None:
```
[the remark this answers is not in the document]
samuelallan72: For the design here, I wanted to pull the for-loop up to this method, to keep better separation of responsibilities between the two modules: so this function here is responsible for the verification (checking each ceph cloud, adding warning messages), while the functions in ceph.py are only responsible for the lower level ceph-related functions (getting ceph-mon apps, getting versions, etc.).

samuelallan72:
Tests added, some bugs fixed, and properly ready for review now. :)

jneo8 (approved):

gabrielcocenza (changes requested):
If I understood this implementation correctly, this will have an excess of warning that can confuse the user.

The critical point of having mismatched ceph versions are after upgrading ceph-osd or when you are going to start upgrading a cloud where all components are in the same OpenStack release. 
In the middle of an upgrade is expected that the cloud will have different versions. User can upgrade control-plane (ceph-mon, ceph-radosgw) and when the time comes for upgrading data-plane a warning message will show up which is not necessary at all.

I'm not in favor of always checking mismatched ceph versions every time that COU will upgrade without considering the context of the cloud. IMO this check should be a post upgrade check of ceph-osd

samuelallan72:
@gabrielcocenza thanks for your comments. There are a lot of factors to checking this, and I don't think there's a perfect solution unfortunately. I agree that false-positive warnings aren't ideal - it was a tradeoff to avoid too much extra complexity. We can definitely revisit it.

> IMO this check should be a post upgrade check of ceph-osd.

Some questions to consider around just doing a post upgrade check of ceph-osd:

- What about a pre-upgrade check to check everything is consistent before proceeding with the upgrade (eg. if cou was killed before all the ceph-osd upgrade steps completed)?
- What about the case where we have multiple ceph-osd applications related to ceph-mon?
- This isn't specific to ceph-osd - what about checking for ceph-radosgw versions for example?

Maybe we could have a meeting to discuss where we want to go with this? It's not clear what we should go with, because it looks like we can go as simple or as complex as we like here.

gabrielcocenza:
> What about a pre-upgrade check to check everything is consistent before proceeding with the upgrade (eg. if cou was killed before all the ceph-osd upgrade steps completed)?

Hum, good point. If we add a pre-check for ceph-mon to force the integrity, we catch this situation. Moreover, we can halt the upgrade to force ceph versions and not just give a warning message.

Ensuring that ceph-versions are equal **before** `ceph-mon`  can guarantee that the cloud is into consistent state. Checking **after** ceph-osd will also ensure consistency. Everything that it's between ceph-mon and ceph-osd upgrade shouldn't be checked.

> What about the case where we have multiple ceph-osd applications related to ceph-mon?

Yeah, that is a corner case, but we can add a post-upgrade check if the upgrade is targeting the whole cloud or data-plane. This way a check wouldn't be added if is targeted to upgrade control-plane or hypervisors

> This isn't specific to ceph-osd - what about checking for ceph-radosgw versions for example?

Right, that is why we should check before `ceph-mon` and after `ceph-osd`. radosgw shouldn't have different versions on those scenarios

jneo8:
> Hum, good point. If we add a pre-check for ceph-mon to force the integrity, we catch this situation. Moreover, we can halt the upgrade to force ceph versions and not just give a warning message.

> Ensuring that ceph-versions are equal before ceph-mon can guarantee that the cloud is into consistent state. Checking after ceph-osd will also ensure consistency. Everything that it's between ceph-mon and ceph-osd upgrade shouldn't be checked.

Agree that we should consider the context. The problem is kind of similar to this issue: https://github.com/canonical/charmed-openstack-upgrader/issues/561

IMO, COU don't have a mechanism to target the correct state during the middle of the upgrading now. Adding this mechanism now is a huge effort job. I agree too much warning may interfere user but it's better than nothing. I suggest we discuss how to detect the state first then move on to follow your proposal.

For this PR I suggest we can merge it first because I can see the cloud operator will choose to show the more information, including warning, instead of nothing.

gabrielcocenza:
What would be the worst case scenario? Some ceph-mon units upgrades and some not? Than I still believe the check is useful. Instead of raising mismatch, it will say that you have different ceph versions. Once all ceph-mon units are in the same version COU would go ahead to the next applications

jneo8:
> What would be the worst case scenario? Some ceph-mon units upgrades and some not? Than I still believe the check is useful. Instead of raising mismatch, it will say that you have different ceph versions. Once all ceph-mon units are in the same version COU would go ahead to the next applications

The check you mention here is the last step in ceph-osd upgrade or the verify check before any upgrade?

Pjack:
Do these two solutions conflict with each other, so we have to choose only one?

Solution A (Samuel's PR): Add a check in the sanity check to provide an earlier warning message to the user. These warnings may occasionally be false alarms, but from my perspective, this isn’t a major issue. The operators can choose to ignore it and contine the upgrade if they are confident it's not an issue.

Solution B (Samuel+Gabriel's proposal) : Add a check in the pre-check/post-check to enforce consistency between ceph_mon and ceph_osd units. We need further discussion to determine if this is a reasonable constraint. For example, do we want to enforce consistency every time, no matter it's caused by manual intervention or a bug somewhere? Personally, I vote 'yes,' but operators may be frustrated. 

If there is no strong dependency between these two solutions, why not implement Solution A first and then Solution B later? It’s better to have something than nothing.

gabrielcocenza:
> The check you mention here is the last step in ceph-osd upgrade or the verify check before any upgrade?

I see two potential places to implement:

1 - Pre-upgrade step on ceph-mon. Ceph-mon is the first application that should be upgraded and it doesn't matter the state of the cloud, before upgrading `ceph-mon` all ceph applications should be at the same ceph version. I don't see side effects in this implementation. If for some reason a ceph-mon unit doesn't upgrade it will raise mismatched and manual intervention is required as it happens with any other application. After ceph-mon is upgraded you'll have a cloud that is on half way to upgrade and similar checks starts having false positives.

2 - General post-upgrade check that happens when upgrading the whole cloud or data-plane. This require a little bit more business logic and should happen as last step of upgrading a cloud.

Implementing both would be double protection. E.g: user upgrade from `Ussuri` -> `Victoria`.  We would check before starting upgrading to Victoria on ceph-mon and then check again by the end when all data-plane apps are upgraded to `Victoria`.

Having both would be nice, but having just option 1 is enough IMO. I don't see the need of major refactors and the check is more precise without false positives.

samuelallan72:
Thanks for the discussion above, this has been helpful!

@gabrielcocenza 

> Pre-upgrade step on ceph-mon. Ceph-mon is the first application that should be upgraded...

> General post-upgrade check that happens when upgrading the whole cloud or data-plane. 

Agreed, these make sense and I think are doable without much refactoring.

I'll plan to make the following changes to this PR:

- remove the pre-plan verify check
- only run the post-upgrade sanity check if data-plane was included in the upgrades
- add a pre-upgrade check to ceph-mon

For later on, it would be nice to have a stand-alone method to check the cloud (eg. `cou verify|lint|check` - in the case where we want to check cloud consistency or suitability for upgrade, without actually planning an upgrade). Then we can add a check to warn on mismatched running ceph versions there. But here, I do see the downsides of a pre-plan warning because of the high chance of false positives.

jneo8:
Thanks for clarify. @gabrielcocenza  @samuelallan72 

The only concern I have is that we're missing a case:

### What happens if Ceph is already running on different versions? Should we allow the user to upgrade any applications in this scenario?

While the **Pre-upgrade step on ceph-mon** halts the Ceph upgrade, it doesn’t address other applications. We can't assume the user has the necessary information from the previous **General post-upgrade check**. If they’re unaware that Ceph is running on different versions, they might proceed with the upgrade without a warning.

#### Proposal:
We introduce a prompt in such cases, requiring the user to type `yes` to continue if the Ceph versions are inconsistent.

This wouldn't be a warning but would still provide the user with crucial information and seek explicit confirmation before proceeding.

samuelallan72:
@gabrielcocenza @jneo8 what do you think about going back to basics here - the original issue was that ceph-osd running version wasn't updating (sounds like ceph-osd daemon wasn't being restarted). So adding a check to ceph-osd after the upgrade may be the neatest option. Thoughts?

Of course this means that cou won't re-run the verification if cou crashes and is restarted, but then we're back to the discussion here: how to implement a pre-plan check without too many false positives.

gabrielcocenza:
> So adding a check to ceph-osd after the upgrade may be the neatest option. Thoughts?

It's not just ceph-osd that I found this issue. It also can happen on cepg-radosgw

I'm not following the concerns of halting the upgrade on ceph-mon if we have different ceph versions

gabrielcocenza (approved):

Issue #487, opened 2024-07-09
Unclear what "running mismatched openstack versions" means


samuelallan72:
For example:

```
2024-07-09 07:40:05 [WARNING] Cannot generate plan for 'aodh'
        Units of application aodh are running mismatched OpenStack versions: 'victoria': ['aodh/0'], 'wallaby': ['aodh/1', 'aodh/2']. This is not currently handled.
```

It would be good to output more information at a lower level to help with debugging:

- which packages/services were tested for version, and how they were tested
- what the expected version ranges were
- what the actual version ranges are

samuelallan72:
@gabrielcocenza @rgildein @jneo8 @chanchiwai-ray is there any disadvantage to checking directly for mismatched __workload__ versions instead? Then we can be clearer in the error message, by referencing the actual workload versions.

jneo8:
> Are there any cases where workload versions may be mismatched, but the relevant openstack release is still the same (eg. if the versions were 21.0.1 and 21.0.2, only differing by a patch release)? Is it ok if workload versions are mismatched at all?

This shouldn't happen and can be danger.

> Thoughts on this design?

Nice and clear, +1 for it.
And can we also output the expected version?

gabrielcocenza:
> Are there any cases where workload versions may be mismatched, but the relevant openstack release is still the same (eg. if the versions were 21.0.1 and 21.0.2, only differing by a patch release)? Is it ok if workload versions are mismatched at all?

This is not suppose to happen. We are using the [packaging](https://github.com/pypa/packaging) lib and in the openstack_lookup.csv we add the lower and upper version (not inclusive).
So for aodh, for example, it's considered as victoria: 11.0.0 <= pkg_version <  12.0.0

We consider mismatched regarding o7k releases map and not the fine grained version itself. So 11.0.0 and 11.9.9 is Victoria and don't have mismatched versions

samuelallan72:
Thanks for your comments. :)

PR #585, opened 2024-10-10
Improve checking for mismatched unit openstack versions

samuelallan72:
Improve checking for mismatched unit openstack versions.

If the units of an application are running mismatched workload versions,
this is normally ok
(a user may have manually ran an apt upgrade for example).
But if the versions are so mismatched
that they are detected as being for mismatched openstack releases,
then that's a problem,
potentially indicating a failed upgrade on a unit for example.
Eg. if one unit is running a workload version only for victoria,
while the other two units are running a version for wallaby,
this is a consistency problem,
and COU should block an automated upgrade.

COU was previously checking this case fine,
but the error message was not clear.

This commit improves the error message here,
to clearly show:
- the workload version per unit
- interpreted openstack release per unit
- that the information was retrieved from
  the unit workload information from juju

This should help the user to start debugging the issue.

Fixes: #487

### DIFF
--- a/cou/apps/base.py
+++ b/cou/apps/base.py
@@ -912,14 +912,18 @@ class OpenStackApplication(Application):
             )
 
     def _check_mismatched_versions(self) -> None:
-        """Check that there are no mismatched versions on app units.
+        """Check that there are no unexpected mismatched versions on app units.
+
+        For cases where mismatched versions may be expected
+        (eg. nova-compute which is upgraded unit-by-unit),
+        then it will not check.
 
         :raises MismatchedOpenStackVersions: When the units of the app are running
                                              different OpenStack versions.
         """
-        # NOTE (gabrielcocenza) nova-compute is upgraded using paused-single-unit,
-        # so it's possible to have mismatched version in applications units that are
-        # nova-compute or colocated with it.
+        # nova-compute is upgraded one unit at a time,
+        # so it's possible to have mismatched version in
+        # units of applications that are nova-compute or colocated with it.
         if any(
             "nova-compute" in app_charm
             for machine in self.machines.values()
@@ -927,14 +931,16 @@ class OpenStackApplication(Application):
         ):
             return
 
-        o7k_versions = self.o7k_release_units
-        if len(o7k_versions.keys()) > 1:
-            mismatched_repr = [
-                f"'{openstack_release.codename}': {units}"
-                for openstack_release, units in o7k_versions.items()
-            ]
-
+        if len({self.get_latest_o7k_version(unit) for unit in self.units.values()}) > 1:
+            formatted_results = "\n".join(
+                f"  {unit.name}: {self.get_latest_o7k_version(unit)} "
+                f"(workload: {unit.workload_version})"
+                for unit in sorted(self.units.values(), key=lambda unit: unit.name)
+            )
             raise MismatchedOpenStackVersions(
-                f"Units of application {self.name} are running mismatched OpenStack versions: "
-                f"{', '.join(mismatched_repr)}. This is not currently handled."
+                f"Units of application {self.name} are running mismatched OpenStack releases, "
+                "based on the workload versions as reported by juju status. "
+                "Observed OpenStack releases for each unit:\n"
+                f"{formatted_results}\n"
+                "This requires manually resolving the issue to continue."
             )
--- a/tests/unit/apps/test_base.py
+++ b/tests/unit/apps/test_base.py
@@ -407,15 +407,9 @@ def test_check_application_target_error(o7k_release, apt_source_codename):
         app._check_application_target(target)
 
 
-@patch("cou.apps.base.OpenStackApplication.o7k_release_units", new_callable=PropertyMock)
-def test_check_mismatched_versions_exception(mock_o7k_release_units, model):
+@patch("cou.apps.base.OpenStackApplication.get_latest_o7k_version", new_callable=MagicMock)
+def test_check_mismatched_versions_exception(mock_get_latest_o7k_version, model):
     """Raise exception if workload version is different on units of a control-plane application."""
-    exp_error_msg = (
-        "Units of application my-app are running mismatched OpenStack versions: "
-        r"'ussuri': \['my-app\/0', 'my-app\/1'\], 'victoria': \['my-app\/2'\]. "
-        "This is not currently handled."
-    )
-
     machines = {f"{i}": generate_cou_machine(f"{i}", f"az-{i}") for i in range(3)}
     units = {
         "my-app/0": Unit(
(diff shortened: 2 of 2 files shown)


samuelallan72:
Hmm this broke the mocked-plans tests. I see designate-bind has different workload versions in the mock model. Maybe it is possible for different workload versions to be seen on a valid cloud? Perhaps because one unit hasn't had `apt upgrade` run for a while? I guess COU will bring it all into line once it runs `apt upgrade` on all the units before upgrading.

Maybe we should check for openstack release match only? @jneo8 @gabrielcocenza

cou/apps/base.py
```diff
@@ -909,29 +909,34 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
...
-
-            raise MismatchedOpenStackVersions(
-                f"Units of application {self.name} are running mismatched OpenStack versions: "
-                f"{', '.join(mismatched_repr)}. This is not currently handled."
+        if len({unit.workload_version for unit in self.units.values()}) > 1:
+            formatted_results = "\n".join(
+                f"  {unit.name}: {unit.workload_version} ({self.get_latest_o7k_version(unit)})"
+                for unit in sorted(self.units.values(), key=lambda unit: unit.name)
+            )
+            raise MismatchedWorkloadVersions(
```
gabrielcocenza, line 936: You can't look at the version itself because it will break as it happens in designate-bind. The logic is right, you just need to refactor the message to show the workload version

samuelallan72: I'm confused - according to the conversation on https://github.com/canonical/charmed-openstack-upgrader/issues/487 : 
I asked:

> Are there any cases where workload versions may be mismatched, but the relevant openstack release is still the same (eg. if the versions were 21.0.1 and 21.0.2, only differing by a patch release)? Is it ok if workload versions are mismatched at all?

And both you and @jneo8 said that this case should not happen:

> This shouldn't happen and can be danger. @jneo8 

> This is not suppose to happen. @gabrielcocenza

chanchiwai-ray: I realized that this is not the same problem I thought it should be.

Should we check if the workload version is in the version range as defined by `openstack_lookup.csv` for a given o7k release?

E.g if COU detect the app is in yoga, and the version range is [A, A+10), the we only need to make sure that all workload version of the units in that app is in this range

gabrielcocenza: > I'm confused - according to the conversation

Sorry I've interpreted wrong the question :sweat_smile: 

I thought you were asking if for example 10.0.2 would be considered different from 10.0.3 when both are considered Ussuri.

They can have different versions in the same OpenStack release before upgrading. Users can potentially ssh to a unit so troubleshoot and upgrade the package and not do the same for the other units.

samuelallan72: Yep ok that makes sense. So we should go back to only verifying the openstack release matches. And the work here can be simply improving the error message.

tests/unit/apps/test_base.py
```diff
@@ -455,10 +450,11 @@ def test_check_mismatched_versions_exception(mock_o7k_release_units, model):
         workload_version="18.1.0",
     )
 
-    with pytest.raises(MismatchedOpenStackVersions, match=exp_error_msg):
+    with pytest.raises(MismatchedOpenStackVersions, match=".*mismatched OpenStack releases.*"):
         app._check_mismatched_versions()
 
 
+# TODO: fix this test - it's not accurate
```
samuelallan72, line 457: TODO

cou/apps/base.py
```diff
@@ -912,29 +912,35 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
...
+            formatted_results = "\n".join(
+                f"  {unit.name}: {self.get_latest_o7k_version(unit)} "
+                f"(workload: {unit.workload_version})"
+                for unit in sorted(self.units.values(), key=lambda unit: unit.name)
+            )
             raise MismatchedOpenStackVersions(
-                f"Units of application {self.name} are running mismatched OpenStack versions: "
-                f"{', '.join(mismatched_repr)}. This is not currently handled."
+                f"Units of application {self.name} are running mismatched OpenStack releases, "
+                "based on the workload versions as reported by juju status. "
```
samuelallan72, line 942: TODO: juju cli to get this info

cou/apps/base.py
samuelallan72: Updated :)

cou/apps/base.py
samuelallan72:
```
for unit in vault/{0,1,2}; do juju show-unit $unit --format json | jq '."'"${unit}"'"."workload-version"'; done
```

The best I could do is this - anyone know of a more concise way? :) Perhaps this kind of snippet could be put in COU docs for later reference and debugging help?

gabrielcocenza: To debug users need to ssh and solve. That could be a simple package update and upgrade or changing sources.

cou/apps/base.py
```diff
@@ -912,29 +912,35 @@ def _check_application_target(self, target: OpenStackRelease) -> None:
...
             return
 
-        o7k_versions = self.o7k_release_units
-        if len(o7k_versions.keys()) > 1:
-            mismatched_repr = [
-                f"'{openstack_release.codename}': {units}"
-                for openstack_release, units in o7k_versions.items()
-            ]
-
+        if len({self.get_latest_o7k_version(unit) for unit in self.units.values()}) > 1:
```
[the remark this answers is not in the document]
samuelallan72: I decided against using `self.o7k_release_units` in this method, because the data structure it returns is awkward in the context and it doesn't provide the actual workload versions.

I could use it just for this conditional, but I'd rather be consistent with the following code, than use two separate methods.

cou/apps/base.py
samuelallan72: Yep, I don't mean to debug though, I mean to manually get the individual units' workload versions. Just with `juju status` it shows a single way so we can't see the version per-unit.

cou/apps/base.py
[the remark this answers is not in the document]
gabrielcocenza: Solving as this is not a blocker

cou/apps/base.py
gabrielcocenza: COU will do that for you, so I don't see a need to point to a spinet

samuelallan72: True, I was more curious for my own manual debugging reference.

jneo8: Maybe you would like to check the details implementation of https://ops.readthedocs.io/en/latest/index.html#ops.Unit.set_workload_version ?

gabrielcocenza (approved):

jneo8 (approved):

Issue #492, opened 2024-07-12
Document where mysql backups are saved


samuelallan72:
It's currently hard to find the mysql backups.  We should add some documentation.  Suggestions:

- print the path to the backup file during the backup step
- add information to the user docs (maybe somewhere with an explanation of where files are stored, and/or from where the backup step is introduced)

samuelallan72:
Fixed by https://github.com/canonical/charmed-openstack-upgrader/pull/589

PR #589, opened 2024-10-18
Add document how to find mysql backup file and print path in output

jneo8:
- Print backup file path on output
- Refactor how-to backup document, add information where to find the file.

### DIFF
--- a/cou/steps/backup.py
+++ b/cou/steps/backup.py
@@ -47,8 +47,9 @@ async def backup(model: Model) -> Path:
     logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
 
-    # info() stop the current spinner and print the text with
-    # symbol ℹ.
+    # Note: a spinner is already running when this function is called,
+    # so to cleanly print text, we can use .info(), then restart the spinner.
+    # info() stop the current spinner and print the text with symbol ℹ.
     progress_indicator.info(f"Backed up MySQL to {local_file}")
     # Restart the spinner with previous text
     progress_indicator.start()
--- a/cou/steps/backup.py
+++ b/cou/steps/backup.py
@@ -44,8 +44,9 @@ async def backup(model: Model) -> Path:
     await model.run_on_unit(unit_name, f"chmod o+rx {basedir}")
 
     local_file = COU_DATA / os.path.basename(remote_file)
-    logger.info("SCP from  mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
+    logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
+    logger.info("Backup MySQL to %s", local_file)
 
     logger.info("Remove permissions to read mysql-innodb-cluster:%s ...", basedir)
     await model.run_on_unit(unit_name, f"chmod o-rx {basedir}")
--- a/cou/steps/backup.py
+++ b/cou/steps/backup.py
@@ -18,7 +18,7 @@ import os
 from pathlib import Path
 
 from cou.exceptions import UnitNotFound
-from cou.utils import COU_DATA
+from cou.utils import COU_DATA, progress_indicator
 from cou.utils.juju_utils import Model
 
 logger = logging.getLogger(__name__)
@@ -46,7 +46,9 @@ async def backup(model: Model) -> Path:
     local_file = COU_DATA / os.path.basename(remote_file)
     logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
-    logger.info("Backup MySQL to %s", local_file)
+
+    progress_indicator.info(f"Backed up MySQL to {local_file}")
+    progress_indicator.start()
 
     logger.info("Remove permissions to read mysql-innodb-cluster:%s ...", basedir)
     await model.run_on_unit(unit_name, f"chmod o-rx {basedir}")
--- a/cou/steps/backup.py
+++ b/cou/steps/backup.py
@@ -47,7 +47,10 @@ async def backup(model: Model) -> Path:
     logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
 
+    # info() stop the current spinner and print the text with
+    # symbol ℹ.
     progress_indicator.info(f"Backed up MySQL to {local_file}")
+    # Restart the spinner with previous text
     progress_indicator.start()
 
     logger.info("Remove permissions to read mysql-innodb-cluster:%s ...", basedir)
--- a/cou/steps/backup.py
+++ b/cou/steps/backup.py
@@ -49,6 +49,8 @@ async def backup(model: Model) -> Path:
 
     # Note: a spinner is already running when this function is called,
     # so to cleanly print text, we can use .info(), then restart the spinner.
+    # Not using `print` directly, because that will cause corrupted output
+    # while the spinner is running.
     # info() stop the current spinner and print the text with symbol ℹ.
     progress_indicator.info(f"Backed up MySQL to {local_file}")
     # Restart the spinner with previous text
--- a/docs/how-to/backup.rst
+++ b/docs/how-to/backup.rst
@@ -1,12 +1,24 @@
 ==========================================
-Plan/Upgrade without cloud database backup
+Backup database
 ==========================================
 
 By default, **COU** plans for and runs a backup step of the cloud database before
-proceeding to actual upgrade steps. This can be turned off with `--no-backup`  flag.
+proceeding to actual upgrade steps.
+
+Find mysql backup file
+~~~~~~~~~~~~~~~~~~~~~~
+
+The file is store under `$COU_DATA`, which is `/home/$USER/.local/share/cou` if `$USER` exists else current directory.
+
+
+Plan/Upgrade without cloud database backup
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To plan/upgrade without backup, this can be turned off with `--no-backup` flag.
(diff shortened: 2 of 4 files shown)


gabrielcocenza (dismissed):

docs/how-to/find-mysql-backup-file.rst
```diff
@@ -0,0 +1,5 @@
+======================
```
samuelallan72, line 1: I'm not sure this deserves its own docs page? How about in a doc where other paths and such are described?

cou/steps/backup.py
```diff
@@ -44,8 +44,9 @@ async def backup(model: Model) -> Path:
     await model.run_on_unit(unit_name, f"chmod o+rx {basedir}")
 
     local_file = COU_DATA / os.path.basename(remote_file)
-    logger.info("SCP from  mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
+    logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
+    logger.info("Backup MySQL to %s", local_file)
```
samuelallan72, line 49: Should this be past tense?

Also, will this be printed to stdout during normal operation? So the user doesn't need to dig through the log file.

samuelallan72 (changes requested):
@jneo8 is this addressing https://github.com/canonical/charmed-openstack-upgrader/issues/492 ?

docs/how-to/find-mysql-backup-file.rst
jneo8: The purpose is different. I think this document serve user who what to find the backup file.

cou/steps/backup.py
jneo8: Updated in last commit.

Now use halo's `spinner.info` to print the message.

cou/steps/backup.py
```diff
@@ -44,9 +44,12 @@ async def backup(model: Model) -> Path:
     await model.run_on_unit(unit_name, f"chmod o+rx {basedir}")
 
     local_file = COU_DATA / os.path.basename(remote_file)
-    logger.info("SCP from  mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
+    logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
 
+    progress_indicator.info(f"Backed up MySQL to {local_file}")
+    progress_indicator.start()
```
chanchiwai-ray, line 51: (question) I am not super familiar with `halo`, is this valid? should we `start` before `info`? do we need `start` and without `stop`

jneo8: `info` will stop the progress_indicator.

All the steps have `progress_indicator.start` at  https://github.com/canonical/charmed-openstack-upgrader/blob/0d3968d8998fb05df7eb3ab8369eb4ef44116a4d/cou/steps/execute.py#L120 .

chanchiwai-ray: thanks

chanchiwai-ray (approved):

docs/how-to/find-mysql-backup-file.rst
```diff
@@ -0,0 +1,5 @@
...
+Find mysql backup file
+======================
+
+The file is store under *$COU_DATA* , which is */home/$USER/.local/share/cou* if *$USER* exists else current directory.
```
samuelallan72, line 5: I think these should be code blocks, not italics

docs/how-to/find-mysql-backup-file.rst
samuelallan72: Yes, but it's only one sentence - that's why I'm thinking perhaps it can be put with an existing doc?

jneo8: The only document close to it in how-to is https://github.com/canonical/charmed-openstack-upgrader/blob/main/docs/how-to/no-backup.rst

I am thinking if it's a right approach to merge them and rename to `how-to-backup`.

@samuelallan72 How do you think?

jneo8: Updated.

cou/steps/backup.py
samuelallan72, line 51: do you need to pass anything to .start here? Elsewhere, it's `.start(step.description)`.

jneo8: I do the local testing and it will use the previous text. So it should be fine I guess.

samuelallan72 (dismissed):
Thanks!

docs/how-to/backup.rst
```diff
@@ -1,12 +1,24 @@
...
 ==========================================
 
 By default, **COU** plans for and runs a backup step of the cloud database before
-proceeding to actual upgrade steps. This can be turned off with `--no-backup`  flag.
+proceeding to actual upgrade steps.
+
+Find mysql backup file
+~~~~~~~~~~~~~~~~~~~~~~
+
+The file is store under `$COU_DATA`, which is `/home/$USER/.local/share/cou` if `$USER` exists else current directory.
```
chanchiwai-ray, line 11: Please fix the all code samples: use [backquotes](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup).

chanchiwai-ray (commented):
Please check the output document: https://canonical-charmed-openstack-upgrader--589.com.readthedocs.build/en/589/how-to/backup/

docs/how-to/backup.rst
jneo8: Updated

chanchiwai-ray (approved):

cou/steps/backup.py
rgildein: I think we should use `progress_indicator.start(f"Backed up MySQL to {local_file}")`, what is the benefit of using info and then start? Should we also do stop here `progress_indicator.succeed`?

rgildein (commented):
Title of this is "docs: Add document how to find mysql backup file" and you are changing also the code. Can you align it and provide output example? Thanks

cou/steps/backup.py
samuelallan72: I think the idea is to persist the message on the screen. It's not starting a process, it's logging a message. Probably using `print()` is fine like in other places in COU's codebase.

Or perhaps we could add the local path to the progress indicator for the backup step, so we need to touch the progres indicator from within the backup function at all? I'm thinking about it now, and that would actually make more sense, because the backup() function shouldn't be managing a progress indicator that is defined and managed outside the function - it makes it less portable.

chanchiwai-ray (dismissed):

cou/steps/backup.py
samuelallan72, line 51: actually can we not touch the progress indicator here? It doesn't seem right for a function to mess with the progress indicator, when it isn't fully managing the indicator (it prints a message and starts the spinner, but nothing else, so it's incomplete).

I'm happy if this simply uses `print()` (like elsewhere in the codebase), or the message be defined higher in the call chain so it's picked up by the existing spinner.

samuelallan72 (changes requested):

cou/steps/backup.py
jneo8: > actually can we not touch the progress indicator here? It doesn't seem right for a function to mess with the progress indicator, when it isn't fully managing the indicator (it prints a message and starts the spinner, but nothing else, so it's incomplete).

I guess that will be a disaster to use `print` when progress indicator is still running. Elsewhere in code base using `print` is when the progress indicator is not running.

And according to description of `spinner.info` [here](https://pypi.org/project/halo/):

```
Stops the spinner and changes symbol to ℹ. If text is provided, it is persisted else current text is persisted. Returns the instance.
```

So the behavior in this PR is the `.info()` will stop the spinner, print one line with symbol ℹ then `spinner.start()` will restart the spinner with previous text. And I think it's a much standard way to use *halo* library compare to using `print`.

samuelallan72: Ok, that's fair. It's not great separation of responsibilities, but I guess we can't improve the situation until we change how we manage output/logs/spinners. Could you add a comment above it to explain why the seemingly-random calls to progress_indicator instead of other methods there?

jneo8: Comment added

cou/steps/backup.py
```diff
@@ -44,9 +44,15 @@ async def backup(model: Model) -> Path:
     await model.run_on_unit(unit_name, f"chmod o+rx {basedir}")
 
     local_file = COU_DATA / os.path.basename(remote_file)
-    logger.info("SCP from  mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
+    logger.info("SCP from mysql-innodb-cluster:%s to %s ...", remote_file, local_file)
     await model.scp_from_unit(unit_name, remote_file, str(local_file))
 
+    # info() stop the current spinner and print the text with
```
samuelallan72, line 50: Thanks for the comment, but I was thinking more of a "why" comment like this:

```suggestion
    # NOTE: a spinner is already running when this function is called,
    # so to cleanly print text, we can use .info(), then restart the spinner.
    # Not using `print` directly,
    # because that will cause corrupted output while the spinner is running.
    # info() stop the current spinner and print the text with
```

samuelallan72 (changes requested):

cou/steps/backup.py
jneo8: I am ok with the change but do we need to explain why not using print?

samuelallan72: To me it's confusing using a global spinner object here - the code makes some assumptions (the global spinner will be running) and reasons for not using alternatives (text will be corrupted by the spinner), so we should clearly write out those assumptions and reasons. :)

jneo8: Updated

samuelallan72 (approved):
Thanks, for making the changes, I think that makes it clearer for future readers. :)

Deezzir (approved):

cou/steps/backup.py
rgildein: If you want to persist it, you can use `progress_indicator.start(f"Backed up MySQL to {local_file}")` and then `progress_indicator.succeed()` without message. Like that, the message will stay there, at least that's how I remember how it works.

PR #593, opened 2024-10-21
Require `enable-auto-restarts=False` for rabbitmq-server

chanchiwai-ray:
Closes: #545


jneo8 (commented):
Question:

Should we change the configuration back after upgrade finish?

chanchiwai-ray:
> Should we change the configuration back after upgrade finish?

We did not change the configuration during `cou upgrade`

jneo8 (dismissed):

cou/apps/auxiliary.py
```diff
@@ -324,12 +324,20 @@ def post_upgrade_steps(
...
+        Due to the charm [bug](https://bugs.launchpad.net/charm-rabbitmq-server/+bug/2046381),
+        if the `enable-auto-restarts` option is enabled, this check will raise an exception.
+
+        :raises ApplicationError: When enable-auto-restarts is enabled.
         """
+        if self.config["enable-auto-restarts"].get("value") is True:
+            raise ApplicationError(
+                "`enable-auto-restarts` must be `False` due to "
+                "https://bugs.launchpad.net/charm-rabbitmq-server/+bug/2046381 "
+                "Please set `enable-auto-restarts=True` when performing upgrades "
```
samuelallan72, line 338: This sentence contradicts the previous - did you mean this instead?:

```suggestion
                "Please set `enable-auto-restarts=false` when performing upgrades "
```

samuelallan72 (changes requested):
Thanks, this looks good! I just flagged a typo in the output message.

cou/apps/auxiliary.py
samuelallan72: And maybe the error message could include the full juju command to update the config for this rabbitmq-server application?

cou/apps/auxiliary.py
```diff
@@ -324,12 +324,21 @@ def post_upgrade_steps(
...
+        if the `enable-auto-restarts` option is enabled, this check will raise an exception.
+
+        :raises ApplicationError: When enable-auto-restarts is enabled.
         """
+        if self.config["enable-auto-restarts"].get("value") is True:
+            raise ApplicationError(
+                "`enable-auto-restarts` must be `False` due to "
+                "https://bugs.launchpad.net/charm-rabbitmq-server/+bug/2046381 "
+                f"Please run `juju config {self.name} enable-auto-restarts=False` "
+                "when performing upgrades and rollback to original value when "
```
samuelallan72, line 339:
```suggestion
                "before performing upgrades and rollback to original value after "
```

cou/apps/auxiliary.py
```diff
@@ -324,12 +324,21 @@ def post_upgrade_steps(
...
+
+        :raises ApplicationError: When enable-auto-restarts is enabled.
         """
+        if self.config["enable-auto-restarts"].get("value") is True:
+            raise ApplicationError(
+                "`enable-auto-restarts` must be `False` due to "
+                "https://bugs.launchpad.net/charm-rabbitmq-server/+bug/2046381 "
+                f"Please run `juju config {self.name} enable-auto-restarts=False` "
+                "before performing upgrades and rollback to original value after "
+                "upgrade is completed"
```
gabrielcocenza, line 340: I think it would be a better user experience to show a message showing that is a know bug and recommend to change the configuration and prompt if the user wants to proceed instead of raising ApplicationError.

Not sure if that would require a lot of refactor or increase the complexity...

samuelallan72: @gabrielcocenza there aren't many things we prompt for like this - I think only vault upgrade? Could you open a github issue about it to consider more prompts in the design of COU (and how this should work with no-prompt mode)?

Something related I realised that would improve the ux here: this would be a good candidate for a pre-plan check in verify_cloud. We could keep both: a hard error here, but also a verify-cloud check to alert the user that they need to change the config before they even start upgrading (along with the rest of the warnings about the cloud). What do you think @chanchiwai-ray ?

chanchiwai-ray: I vaguely remember that pre_plan_sanity_check will also run before upgrade, that's why I can still put it here. 

>Something related I realised that would improve the ux here: this would be a good candidate for a pre-plan check in verify_cloud....

pre_plan_sanity_check and verify_cloud has similar functionality, and we should consider migrating pre_plan_sanity_check to verify_cloud

chanchiwai-ray: > I think it would be a better user experience to show a message showing that is a know bug and recommend to change the configuration and prompt if the user wants to proceed instead of raising ApplicationError.

adding a prompt may make the `--auto-approve` and other logic more complex: do we auto approve in this situation? even though it's not working when it's `True`? do the user know you will change the `enable-auto-restarts=False` on behalf of them? How do you guarantee COU can rollback to the original setting in case of failure during the upgrade, etc... I think those are the complexities...

samuelallan72: Ah right my bad, I missed that this added check is part of the upgrade_plan_sanity_checks. So this is fine by me, and we can refactor later to merge the sanity checks with verify cloud. :+1: :+1:

samuelallan72 (approved):

gabrielcocenza (approved):

Issue #562, opened 2024-10-02
[Feature] Check valid nova-cloud-controller configuration scheduler-default-filters


jneo8:
Based on the charm source code [here](https://opendev.org/openstack/charm-nova-cloud-controller/src/commit/b915f8e6f81df6a776fc1ad6725207a199440a3b/hooks/nova_cc_context.py#L421), the charm uses the `scheduler-default-filters` configuration value to set the `enabled_filters` in `/etc/nova/nova.conf`.

However, some of these filters have been deprecated, as shown [here](https://opendev.org/openstack/charm-nova-cloud-controller/src/commit/b915f8e6f81df6a776fc1ad6725207a199440a3b/hooks/nova_cc_context.py#L367-L396). If any deprecated filters are included in the Nova configuration, the `nova-scheduler` service will fail to start.

COU will hit this issue when upgrade from antelope to bobcat. 

I recommend that the COU (Charmed OpenStack Users) check if the configuration is valid or we should contribute to the charm by implementing validation for this configuration.

samuelallan72:
Ah I saw this issue a little too late... I opened a duplicate, and will close it. But I'll save what I wrote here for reference:

---

The filter option `AvailabilityZoneFilter` was [removed](https://opendev.org/openstack/nova/src/branch/master/releasenotes/notes/az-filter-removal-587f7337040b6216.yaml) at some point, in Bobcat [according to charm-nova-cloud-controller](https://opendev.org/openstack/charm-nova-cloud-controller/src/commit/b915f8e6f81df6a776fc1ad6725207a199440a3b/hooks/nova_cc_context.py#L388).

This means that if we have a scenario where `schedule-default-filters` has been set on `nova-cloud-controller` to a value that includes  `AvailabilityZoneFilter`, and we upgrade from antelope to bobcat, then nova-scheduler will crash with the following message in `/var/log/nova/nova-scheduler.log`:

```txt
2024-10-01 10:42:10.920 724488 CRITICAL nova [None req-4a4d2a3f-bd36-434b-a228-11e3e236d034 - - - - - -] Unhandled error: nova.exception.SchedulerHostFilterNotFound: Scheduler Host Filter AvailabilityZoneFilter could not be found.
2024-10-01 10:42:10.920 724488 ERROR nova Traceback (most recent call last):
2024-10-01 10:42:10.920 724488 ERROR nova   File "/usr/bin/nova-scheduler", line 10, in <module>
[...]

PR #594, opened 2024-10-22
Verify nova_cloud_controller_scheduler_default_filters

chanchiwai-ray:
- Fix some unit tests
- Add verify step for `scheduler_default_filters` for `nova-cloud-controller` during antelope to bobcat upgrade

Closes: #562


cou/steps/plan.py
```diff
@@ -237,6 +243,37 @@ def _verify_highest_release_achieved(analysis_result: Analysis) -> None:
...
+) -> None:
+    """Verify scheduler_default_filters option is set correctly for each OpenStack release.
+
+    :param args: CLI arguments
+    :type args: CLIargs
+    :param analysis_result: Analysis result.
+    :type analysis_result: Analysis
+    """
+    if args.upgrade_group != CONTROL_PLANE:
+        return
```
samuelallan72, line 257: I think this verify function should always run, so that the user gets a picture of the whole cloud before an upgrade.

tests/unit/steps/test_plan.py
```diff
@@ -588,7 +651,7 @@ def test_verify_highest_release_achieved(o7k_release, series):
         (
             OpenStackRelease("ussuri"),
             OpenStackRelease("ussuri"),
-            "Please, upgrade control-plane before data-plane",
+            "Please upgrade control-plane before data-plane",
```
samuelallan72, line 654: I'm curious why this test needed updating - was it passing before with the comma?

samuelallan72 (commented):
Looks good, just a couple of questions/suggestions, thanks. :)

cou/steps/plan.py
chanchiwai-ray: That also make sense :+1:, My thought was that if they upgrade DATA_PLANE, they will not care about nova-cloud-controller/

tests/unit/steps/test_plan.py
chanchiwai-ray: we missed `assert` in line 672

samuelallan72: Ah right, thanks :+1: Nice catch for the tests :+1:

cou/steps/plan.py
samuelallan72: Yeah that is true - and if they're upgrading data-plane only, then that will only succeed if control-plane is already upgraded. Fine to leave as is. :+1:

samuelallan72 (dismissed):

cou/steps/plan.py
chanchiwai-ray: ah, I reverted

chanchiwai-ray: I think both are fine, do you have a preference?

chanchiwai-ray: revert my revert. on a second thought, if they only upgrade data plane, it does not make sense to ask them to update `scheduler_default_filters`

samuelallan72: Yep that makes sense. Should be fine considering the check. To be more accurate, we would need something like `if nova-cloud-controller would be upgraded and nova-cloud-controller.current_release == antelope`, but I'm not sure if COU would support that.

samuelallan72 (dismissed):

cou/steps/plan.py
```diff
@@ -237,6 +243,37 @@ def _verify_highest_release_achieved(analysis_result: Analysis) -> None:
...
+
+    curr_release = analysis_result.current_cloud_o7k_release
+    nova_cloud_controllers = get_applications_by_charm_name(
+        analysis_result.apps_control_plane, "nova-cloud-controller"
+    )
+    for nova_cloud_controller in nova_cloud_controllers:
+        config = nova_cloud_controller.config.get("scheduler-default-filters", "")
+        if curr_release == "antelope" and "AvailabilityZoneFilter" in config:
```
jneo8, line 265: **issue:** 

`AvailabilityZoneFilter` may not be the only case.

I suggest to check the details of release notes of nova and understand all the available filter for different versions.

You can take a reference from the implementation of charm nova-cloud-controller: https://opendev.org/openstack/charm-nova-cloud-controller/src/commit/b915f8e6f81df6a776fc1ad6725207a199440a3b/hooks/nova_cc_context.py#L350-L396

jneo8 (changes requested):
Need to consider more cases.

cou/steps/plan.py
```diff
@@ -237,6 +243,37 @@ def _verify_highest_release_achieved(analysis_result: Analysis) -> None:
...
+    args: CLIargs, analysis_result: Analysis
+) -> None:
+    """Verify scheduler_default_filters option is set correctly for each OpenStack release.
+
+    :param args: CLI arguments
+    :type args: CLIargs
+    :param analysis_result: Analysis result.
+    :type analysis_result: Analysis
+    """
+    if args.upgrade_group != CONTROL_PLANE:
```
gabrielcocenza, line 256: Upgrading the whole cloud (not passing an upgrade group) should also have this verify IMO.
A comment here is also good for future travelers explaining that data_plane already upgraded nova cntroller
```suggestion
    if args.upgrade_group not in {CONTROL_PLANE, None}:
```

gabrielcocenza (changes requested):

cou/steps/plan.py
samuelallan72: Isn't your suggested code change contradictory to what you wrote? :thinking:

cou/steps/plan.py
samuelallan72: Can we leave other cases as out of scope for this PR? We can open a new issue to hunt for other cases.

chanchiwai-ray: I agree with Samuel, it's too early to check all cases from changelogs. This is the one we encountered, and mentioned explicitly in the [changelog](https://docs.openstack.org/charm-guide/latest/release-notes/2023.2-bobcat.html#nova-availabilityzonefilter-removal-in-2023-2).

samuelallan72 (approved):

Pjack (approved):

cou/steps/plan.py
chanchiwai-ray: Thanks! that's been fixed. (it's very easy to forget that `None` means upgrade the whole cloud)

cou/steps/plan.py
chanchiwai-ray: @jneo8 please consider creating new issue to track other potential issues that have not been observed (e.g. from ussuri -> victorial, it's possible becuase `RetryFilter` is not in victorial)

gabrielcocenza (approved):

cou/steps/plan.py
samuelallan72: I created https://github.com/canonical/charmed-openstack-upgrader/issues/600 to track this. :)

PR #607, opened 2025-01-30
Fix unit and func tests

gabrielcocenza:
- as mentioned in the [Python Docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop):
> consider using the higher-level [asyncio.run()](https://docs.python.org/3/library/asyncio-runner.html#asyncio.run) function, instead of using these lower level functions to manually create and close an event loop.

Using `asyncio.run()` fixed the unit tests

- Func tests were weirdly failing because COU was not able to [connect to the model](https://github.com/canonical/charmed-openstack-upgrader/actions/runs/13120383434/job/36604991356) and I couldn't reproduce locally. After upgrading the lib juju version, it seems to fix.

- `pkg_resources.require` is deprecated, so I've changed to use `metadata`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,3 +106,5 @@ ignore-words-list = "assertIn"
 filterwarnings = [
     "ignore::RuntimeWarning",
 ]
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
 import pytest
@@ -67,12 +66,3 @@ def plan_status() -> None:
     """Get an empty PlanStatus for every test case."""
     PlanStatus.error_messages = []
     PlanStatus.warning_messages = []
-
-
-@pytest.fixture
-def event_loop():
-    """Create a fresh event loop for each test."""
-    loop = asyncio.new_event_loop()
-    asyncio.set_event_loop(loop)  # Ensure it's the active loop
-    yield loop
-    loop.close()  # Properly close the loop after the test
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
 import pytest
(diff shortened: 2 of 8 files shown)


tests/unit/conftest.py
```diff
@@ -66,3 +67,12 @@ def plan_status() -> None:
     """Get an empty PlanStatus for every test case."""
     PlanStatus.error_messages = []
     PlanStatus.warning_messages = []
+
+
+@pytest.fixture
+def event_loop():
```
samuelallan72, line 73: Would it be useful to add a comment here that this is automatically used by pytest asyncio (or that it overrides the internal fixture provided already by pytest-asyncio - I'm not sure exactly how it works)?

samuelallan72 (approved):
LGTM - just a question about docs. :)

Deezzir (commented):
I would suggest adding some documentation or a link to the issue for context

gabrielcocenza:
@Deezzir @samuelallan72 functional tests were also failing with timeout because it was not connecting to the model. I've updated the PR description and now both unit and functional tests are working

pyproject.toml
```diff
@@ -106,3 +106,5 @@ ignore-words-list = "assertIn"
 filterwarnings = [
     "ignore::RuntimeWarning",
 ]
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
```
samuelallan72, line 110: After removing the custom fixture and switching to asyncio.run, are these config changes still required?

samuelallan72 (approved):
Looks good to me, just one query. Thanks for digging into this! :)

pyproject.toml
gabrielcocenza: If I remove it shows this deprecation warning:

```
PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))
  ```
  I think we can be one step ahead and be ready for this deprecation. Moreover, the `asyncio_mode = "auto"` makes easier for create unit tests with async and it's not necessary to add the `@pytest.mark.asyncio` decorator.
  
  In summary, it's not strictly necessary to have now, but I think are good changes for the future

samuelallan72: :+1: thanks for confirming :)

PR #609, opened 2025-02-07
Fix lint with the newer mypy version

gabrielcocenza:
Since the new release of the new mypy version (1.15.0), our lint started failing. The bug discription can be found here:
- https://github.com/python/mypy/issues/18634

I propose we pin the version to lower than the new release until the upstream code fix it.


samuelallan72 (changes requested):
Given the conversation on https://github.com/python/mypy/issues/18634 and that it's closed as not planned, could we investigate an alternate fix please? It seems that the code needs some kind of type narrowing before checking `.cr_code`?

samuelallan72 (approved):
:+1: LGTM, thanks

Deezzir (approved):

PR #613, opened 2025-02-25
Refactor tics workflow

gabrielcocenza:
- After [asking](https://warthogs.atlassian.net/browse/TICSISSUES-39) to change all soleng projects expected path for coverage.xml, the current workflow will work, but it can be simplified.
- remove environment in /tmp for the TICS action to use
- rely on pyproject config for the path of the coverage.xml
- tics analysis uses artifacts generated from unit or functional tests.
- Artifacts files are SQLite 3.x databases generated by using the `COVERAGE_FILE` env var from coverage.
- merge of coverage files needs to be the databases and **not** the xml files


Deezzir:
```
Error: [ERROR 5073] The following error(s) occurred while 'pylint' was running: '/usr/bin/python: No module named pylint
Error: [ERROR 5073] The following error(s) occurred while 'flake8' was running: '/usr/bin/python: No module named flake8
```
The last time I worked with the Tics Action, those errors caused the action to fail silently, so that is why I had to include the following line:

```
 # TICS action requires all the project's dependencies to be present to run the analysis.
# Add the the tox bin directory to the PATH so that the TICS action can find the tests dependencies.
# Other paths are added to the PATH to make sure that the TICS action can find the required binaries.
PATH: "/tmp/tox/tics/bin:/snap/bin:/home/runner/.local/bin:/home/runner/.cargo/bin:/bin:/usr/bin:/usr/local/bin:"
```
The results were uploaded successfully, which is nice; maybe they've changed something in the action itself. The errors are still there, though. Can you confirm they are not causing any changes in the final Tiobe report? Have you tested the workflow using the line above?

Deezzir:
I see that they've definitely changed something, as now the bot reports the status of the report on the PR (nice :+1: )

samuelallan72:
> remove environment in /tmp for the TICS action to use

why is this no longer required?

> I 've created a draft PR forking from the repo (otherwise the credentials won't work) 

You could close this one; you can simply open a PR from a branch in this repo; you don't need to use a personal fork. ;)

chanchiwai-ray (approved):

samuelallan72 (changes requested):
Nice! I like the look of this approach, and I think it will allow applying as-is to other projects that may have coverage for both functional and unit tests, or one but not the other.

I noted a couple of things inline though.

.github/workflows/tics.yaml
```diff
@@ -14,32 +14,46 @@ jobs:
...
+        with:
+          name: coverage-unit
+          path: ./artifacts
+        continue-on-error: true  # Don’t fail if artifact is missing
+
+      - name: Download functional test coverage artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: coverage-functional
+          path: ./artifacts
```
chanchiwai-ray, line 32: have you try `pattern: coverage-*` + `merge-multiple: true`, this might combine the result without having manually download two artifact. https://github.com/actions/download-artifact?tab=readme-ov-file#inputs

.github/workflows/check.yaml
```diff
@@ -52,6 +52,23 @@ jobs:
...
+        else
+          echo "exists=false" >> $GITHUB_OUTPUT
+        fi
+
+    - name: Upload Unit Test Coverage Artifact
+      if: ${{ steps.check_unit_coverage.outputs.exists == 'true' }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-unit
+        path: tests/report/coverage-unit.xml
```
chanchiwai-ray, line 69: have you try `if-no-files-found: ignore` to replace the check? https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs

.github/workflows/tics.yaml
samuelallan72: Something important that Ray raised before: we'll get coverage files from every run - eg. when we have tests on different architectures.

We can ensure we don't upload conflicting files by adding some unique identifier to each coverage file, and using this kind of wildcard download.

samuelallan72:
Something we discussed: can we consider adding this tics workflow as a job on the end of the check workflow (with a condition for it to only run if it's on the main branch)?

- the workflow will be run automatically on every new commit to main
- no duplication of work is done for gathering the coverage
- we can run it manually, it and will be automatically be run once a week as part of our weekly tests workflow

tox.ini
```diff
@@ -35,7 +35,7 @@ commands =
 [testenv:unit]
 deps = .[unittests]
 commands = pytest {toxinidir}/tests/unit \
-    {posargs:-v --cov --cov-report=term-missing --cov-report=html --cov-report=xml}
```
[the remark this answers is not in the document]
gabrielcocenza: In the end the file necessary to merge one or more reports are the databases and not the xml files. I left the coverage as it is

.github/workflows/tics.yaml
```diff
@@ -14,23 +14,43 @@ jobs:
         with:
```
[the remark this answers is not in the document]
gabrielcocenza: I tried to tweak a little bit. TICS will work just on manual trigger or when a merge to main happens, but it's located at the check itself.

.github/workflows/tics.yaml
gabrielcocenza: Good point. I've added the arch in the name of the artifact.
I've also applied Ray's suggestion, so now all artifact are downloaded in a single step

.github/workflows/check.yaml
```diff
@@ -140,3 +152,66 @@ jobs:
...
+            fi
+          done
+
+          if [ -n "$coverage_files" ]; then
+            echo "Merging coverage files: $coverage_files"
+            coverage combine $coverage_files
+            coverage xml -o tests/report/coverage.xml
+          else
+            echo "No coverage files found, skipping merge"
+            # Create an empty file to avoid downstream failure
```
Deezzir, line 205: Can you explain what it means to have a downstream failure?

.github/workflows/check.yaml
```diff
@@ -1,6 +1,8 @@
 name: Tests
 
 on:
+  push:
+    branches: [ main ]
```
samuelallan72, line 5: This isn't needed - the check workflow is called by the release workflow on push to main already: https://github.com/canonical/charmed-openstack-upgrader/blob/db6c4cdc51d0b12ccc40a65fed681752b9a8cf0d/.github/workflows/release.yaml#L9-L10

.github/workflows/check.yaml
```diff
@@ -140,3 +152,66 @@ jobs:
...
+      - name: Upload Functional Test Coverage Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          if-no-files-found: ignore
+          name: coverage-functional-${{ env.SYSTEM_ARCH }}
+          path: coverage-functional
+
+  tics-analysis:
+    runs-on: ubuntu-22.04
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }} || ${{ github.event_name == 'workflow_dispatch' }}
```
samuelallan72, line 165: For ` ${{ github.event_name == 'workflow_dispatch' }}`, it could be worth restricting to the main branch still? Otherwise this will trigger if someone manually runs the check workflow on a PR branch.

.github/workflows/check.yaml
Deezzir: Ah, I see, so the tics action won't fail if we don't have any coverage files, but is this even possible?

.github/workflows/check.yaml
```diff
@@ -140,3 +152,66 @@ jobs:
...
+          merge-multiple: true
+          path: artifacts/
+        continue-on-error: true  # Don’t fail if artifact is missing
+
+      - name: Merge coverage reports
+        run: |
+          # Create the path that is expected to have a coverage.xml for tics
+          mkdir -p tests/report/
+
+          coverage_files=""
```
Deezzir, line 192: Maybe something simpler, like coverage_files=$(echo ./artifacts/coverage-*)

.github/workflows/check.yaml
Deezzir: Why does the empty file have a different name, won't this cause any problems with the action?

Deezzir (commented):
LGTM, just a couple of nits

tox.ini
```diff
@@ -32,15 +32,11 @@ commands =
...
-description = Run all tests for snap and produce coverage report
-commands =
-    pytest {toxinidir}/tests \
-    -v --cov --cov-report=xml:{toxinidir}/tests/report/coverage.xml --junit-xml={toxinidir}/tests/report/tests.xml
-
 [testenv:unit]
 deps = .[unittests]
+setenv =
+    {[testenv]setenv}
+    COVERAGE_FILE = coverage-unit
```
samuelallan72, line 39: I would much prefer to see the coverage db file location left as the default (`.coverage`), to avoid needing these extra envvars, gitignore updates, etc. We can use renaming and such as needed when using the github upload/download actions to avoid file name conflicts where needed (we'll need something anyway, because even with the naming split between coverage-unit and func, there will be multiple coverage-unit or coverage-func files - one for each architecture tested).

.github/workflows/check.yaml
```diff
@@ -52,6 +54,16 @@ jobs:
...
       run: tox -e unit
 
+    - name: Determine system architecture
+      run: echo "SYSTEM_ARCH=$(uname -m)" >> $GITHUB_ENV
+    - name: Upload Unit Test Coverage File
+      uses: actions/upload-artifact@v4
+      with:
+        if-no-files-found: ignore
+        name: coverage-unit-${{ env.SYSTEM_ARCH }}
+        path: coverage-unit
```
chanchiwai-ray, line 64: I think this may not be working as expected: since the coverage db file name is the same for all arches. When downloading and merging the artifact files, only one artifact file will remains.

e.g.

uploaded artifacts:
- coverage-unit-aarch64/coverage-unit
- coverage-unit-amd64/coverage-unit
- coverage-func-amd64/coverage-unit
- coverage-func-amd64/coverage-unit

download and merging artifacts (even before `coverage combine`):
- coverage-unit

 Same below

chanchiwai-ray (commented):
The ideal is sound, but need some proof for multi-arch multi-coverage cases

.github/workflows/check.yaml
gabrielcocenza: I didn't understand your point.
Yeah, they will have the same name for several arch, but when we download it's just for the arch of the runner.

```
      - name: Download Coverage Files
        uses: actions/download-artifact@v4
        with:
          pattern: coverage-*-${{ env.SYSTEM_ARCH }}
```
 
 So if the runner is `aarch64`, it will download just the artifacts of this arch and merge it.

.github/workflows/check.yaml
gabrielcocenza: Different name it was my mistake. Fixed it, thanks.

tox.ini
gabrielcocenza: IMO this is personal taste. The changes are minimal:
- one extra env var for each test
- one extra gitignore

The benefit of the current approach is that it's easier to replicate locally.  If for some reason starts failing you can run locally, separate files will be generated to analyse. Otherwise you can easily overwrite and it's not very intuitive

Deezzir (approved):
Thanks Gabriel!

.github/workflows/check.yaml
```diff
@@ -189,12 +189,7 @@ jobs:
           # Create the path that is expected to have a coverage.xml for tics
           mkdir -p tests/report/
 
-          coverage_files=""
-          for file in ./artifacts/coverage-*; do
-            if [ -f "$file" ]; then
-              coverage_files="$coverage_files $file"
-            fi
-          done
+          coverage_files=$(echo ./artifacts/coverage-*)
```
samuelallan72, line 192: Does it work with a simple glob, removing the echo middleman?

```suggestion
          coverage_files=./artifacts/coverage-*
```

.github/workflows/check.yaml
samuelallan72: Ah I see, but don't we want to download all the coverage files, across all arches, to merge them all? This would get coverage for any arch-specific code.

tox.ini
samuelallan72: Hmm I see your point. I'm not sure I'm convinced though - the html/xml files output will still overwrite the previous one.

If we go this way, please keep them as hidden files, following the pattern of the original file (`.coverage` -> `.coverage-unit`, `.coverage-func`). That should help avoid confusion about these new files, and you can edit the existing gitignore line (`.coverage` -> `.coverage*`).

samuelallan72: Hmm I see your point. I'm not sure I'm convinced though - the html/xml files output will still overwrite the previous one.

If we go this way, please keep them as hidden files, following the pattern of the original file (`.coverage` -> `.coverage-unit`, `.coverage-func`). That should help avoid confusion about these new files, and you can edit the existing gitignore line (`.coverage` -> `.coverage*`).

.github/workflows/check.yaml
chanchiwai-ray: > Ah I see, but don't we want to download all the coverage files, across all arches, to merge them all? This would get coverage for any arch-specific code.

Yes, that's my question. I am not sure how tics workflow work. Do we need to upload multiple times to tiobeweb for different arches?

chanchiwai-ray: Maybe the arch specific part is not relevant here since it only has amd64 here. I will ask this in https://github.com/canonical/hardware-observer-operator/pull/395  when this is done. Feel free to take a look at the comments above and resolve it

tox.ini
gabrielcocenza: Ok. Changed to be hidden files

.github/workflows/check.yaml
gabrielcocenza: Now that the files are hidden, using echo is necessary. Before it was possible to remove the echo

.github/workflows/check.yaml
gabrielcocenza: I don't see in the SSDLC spec about tics integration what the projects should do for multi arch and the GUI seems to not be prepared for that. Moreover, in our projects most of the time we don't use low-level system operation.

I think we can let to be out of scope by now. If there is a clear guidance in the future, merging multiple coverage from different archs or sending into different files we can easily adapt it.

Deezzir (approved):
LGTM

.github/workflows/check.yaml
samuelallan72: why is it necessary for hidden files? Doesn't this work: `echo ./artifacts/.coverage-*`?

.gitignore
```diff
@@ -11,7 +11,7 @@ __pycache__/
 
 # Test files and directories
 .pytest_cache/
-.coverage
+.coverage-*
```
samuelallan72, line 14: I would suggest this so that existing `.coverage` files (or `.coverage` files in projects that don't have the new workflow yet) will continue to be ignored.

```suggestion
.coverage*
```

samuelallan72 (dismissed):
A couple of nonblocking nits, but otherwise this looks fine to me :+1:

.github/workflows/check.yaml
gabrielcocenza: Yes, it works. I don't know exactly what I did before, but it was not working, Leaving the parenthesis and removing the echo works.

.github/workflows/check.yaml
```diff
@@ -191,7 +191,7 @@ jobs:
           # Create the path that is expected to have a coverage.xml for tics
           mkdir -p tests/report/
 
-          coverage_files=$(echo ./artifacts/.coverage-*)
+          coverage_files=(./artifacts/.coverage-*)
```
samuelallan72, line 194: The parens must be removed now here, otherwise it's treated as a bash array, and the logic further on won't work the same any more.

```suggestion
          coverage_files=./artifacts/.coverage-*
```

samuelallan72 (changes requested):
see https://github.com/canonical/charmed-openstack-upgrader/pull/613/commits/401a26e759a1615d0412bced7807e684e01376a5#r1988095978

.github/workflows/check.yaml
samuelallan72: It doesn't work for me keeping the parentheses. It should work without the parens - see https://github.com/canonical/charmed-openstack-upgrader/pull/613/commits/401a26e759a1615d0412bced7807e684e01376a5#r1988095978

.github/workflows/check.yaml
gabrielcocenza: I use zsh and apparently bash and zsh interpret this differently. I've checked on bash and your suggestion works, but breaks on zsh.

Too much confusion for a simple thing.  Leaving echo as it was initially wasn't a big deal, but in any case it's as you suggested.

Deezzir (approved):

PR #629, opened 2025-05-21
Fix tics

samuelallan72:
Without this, you'll see the following error:

```
INFO: pip is looking at multiple versions of charmed-openstack-upgrader[functests] to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r ./tests/functional/requirements.txt (line 1), charmed-openstack-upgrader and charmed-openstack-upgrader[functests]==2.2.0.post5 because these package versions have conflicting dependencies.

The conflict is caused by:
    charmed-openstack-upgrader[functests] 2.2.0.post5 depends on juju<3.5 and >=3.0
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
    charmed-openstack-upgrader 2.2.0.post5 depends on juju<3.5 and >=3.0
    zaza 0.0.2.dev1 depends on juju<3.0.0
```

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -191,10 +191,6 @@ jobs:
 
   tics-analysis:
     runs-on: [self-hosted, linux, amd64, tiobe, jammy]
-    if: >
-      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
-      (github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main')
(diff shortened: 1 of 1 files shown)


samuelallan72:
@gabrielcocenza we'll need to add this to automation, to fix this for other projects too.

Pjack (approved):

Issue #640, opened 2025-06-16
Upgrade order when using OVN leads to unexpected data plane outage


facundofc:
The upgrade order COU uses is (highlighting the relevant parts only):

* ovn-central
* nova-compute

Upgrading nova-compute will in turn upgrade OVN in the compute.

This order is not correct according to [OVN's upstream docs](https://docs.ovn.org/en/latest/intro/install/ovn-upgrades.html#rolling-upgrade). The `ovn-controller` running in the computes needs to be upgraded first, and only then the control plane components of OVN may be upgraded (`ovn-central`).

The order used by COU is known to cause a complete data plane outage right after `ovn-central` is upgraded. `ovn-controller` running in the computes will start logging lines like these:

```
2025-06-13T11:04:17.334Z|266778|lflow|WARN|error parsing actions "reg0[15] = check_out_port_sec(); next;": Syntax error at `check_out_port_sec' expecting field name.
```

It's worth noting that this was not observed consistently in all upgrades, so we have reasons to believe it depends on the state of the cloud at the time of the upgrade.

Only workaround so far is to upgrade the computes.

PR #641, opened 2025-06-16
Fix order of COU upgrade

gabrielcocenza:
COU is not respecting the upgrade order of OVN upstream [documentation](https://docs.ovn.org/en/latest/intro/install/ovn-upgrades.html#upgrade-procedures) which can cause data-plane outage. The order should be:
1. Upgrade ovn-controller (ovn-chassis)
2. Upgrade OVN Databases and ovn-northd (ovn-central)
3. Upgrade OVN Integration (neutron-api)

Because ovn-chassis is subordinate, the payload get upgraded just when nova-compute unit upgrades. That is why the ovn-central and neutron-api should upgrade after the nova-compute units.

Fix: #640

### DIFF
--- a/cou/apps/auxiliary.py
+++ b/cou/apps/auxiliary.py
@@ -38,7 +38,6 @@ from cou.utils.juju_utils import Unit
 from cou.utils.openstack import (
     OPENSTACK_TO_TRACK_MAPPING,
     TRACK_TO_OPENSTACK_MAPPING,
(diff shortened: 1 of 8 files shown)


chanchiwai-ray (approved):

gabrielcocenza:
Turning to draft until we have a clear position from OpenStack if this can cause issue

jneo8 (approved):
Thanks, LGTM.

